PR #3, opened 2017-07-08
progress bar (WIP)

brigand:
Hey, cool project!

I'm working on adding a progress bar component, but getting a weird error:

```
$ ./examples/run progress
/Users/fb/github/brigand/ink/lib/instance.js:10
const isClassComponent = component => component.prototype instanceof Component;
                                               ^

TypeError: Cannot read property 'prototype' of undefined
    at isClassComponent (/Users/fb/github/brigand/ink/lib/instance.js:10:39)
    at new Instance (/Users/fb/github/brigand/ink/lib/instance.js:40:27)
    at getOrCreateInstance (/Users/fb/github/brigand/ink/lib/instance.js:20:14)
    at children.forEach.child (/Users/fb/github/brigand/ink/lib/instance.js:84:27)
    at Array.forEach (native)
    at Instance.mount (/Users/fb/github/brigand/ink/lib/instance.js:82:17)
    at children.forEach.child (/Users/fb/github/brigand/ink/lib/instance.js:87:19)
    at Array.forEach (native)
    at Instance.mount (/Users/fb/github/brigand/ink/lib/instance.js:82:17)
    at diff (/Users/fb/github/brigand/ink/lib/diff.js:40:16)
```

[Example file](https://github.com/brigand/ink/blob/progress-bar/examples/progress.js) | [Component file](https://github.com/brigand/ink/blob/progress-bar/lib/components/bar.js)

Any ideas?

PR #4, opened 2017-07-08
Throw when component is missing

brigand:
React has a similar warning. Ran into it when working on #3

### DIFF
--- a/lib/h.js
+++ b/lib/h.js
@@ -5,7 +5,7 @@ const VNode = require('./vnode');
 
 module.exports = (component, props, ...children) => {
 	if (typeof component !== 'function' && typeof component !== 'string') {
-		throw new TypeError(`Expected component to be a function, but received "${typeof component}". You may have forgotten to export a component.`);
+		throw new TypeError(`Expected component to be a function, but received ${typeof component}. You may have forgotten to export a component.`);
 	}
 
 	props = props || {};
--- a/lib/h.js
+++ b/lib/h.js
@@ -5,7 +5,7 @@ const VNode = require('./vnode');
 
 module.exports = (component, props, ...children) => {
 	if (typeof component !== 'function' && typeof component !== 'string') {
-		throw new TypeError(`ink.h: expected component to be a function, but received "${component}". You may have forgotten to export a component.`);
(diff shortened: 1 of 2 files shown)


vadimdemedes:
Shouldn't this be an error, rather than a warning? Otherwise the previous `Cannot read property 'prototype' of undefined` error is still going to appear after warning.

brigand:
@vadimdemedes fixed.

lib/h.js
```diff
@@ -4,6 +4,10 @@ const flatten = require('lodash.flattendeep');
 const VNode = require('./vnode');
 
 module.exports = (component, props, ...children) => {
+	if (typeof component !== 'function' && typeof component !== 'string') {
+		throw new TypeError(`ink.h: expected component to be a function, but received "${component}". You may have forgotten to export a component.`);
```
vadimdemedes: Could you remove `ink.h` and start with `Expected`?

Also, replace `"${component}"` with `typeof component`.

test/h.js
```diff
@@ -163,3 +163,10 @@ test.failing('don\'t merge children of components', t => {
 
 	t.deepEqual(node.props.children, ['x', 'y']);
 });
+
+test('warns on invalid type', t => {
+	t.throws(() => {
```
vadimdemedes: => `t.throws(() => h(), 'Expected component to be a function, got undefined. You may have forgotten to export a component.');`

test/h.js
```diff
@@ -163,3 +163,10 @@ test.failing('don\'t merge children of components', t => {
 
 	t.deepEqual(node.props.children, ['x', 'y']);
 });
+
+test('warns on invalid type', t => {
```
vadimdemedes: `warns` => `throws`
`invalid type` => `missing component`

lib/h.js
brigand: Fixed

test/h.js
brigand: Fixed

test/h.js
brigand: Fixed

vadimdemedes:
Thanks!

PR #6, opened 2017-07-09
Fix tabs in README.md 

damassi:
This PR fixes some of the display issues that come with tabs and GitHub so that the README is more readable:

Before: 

screen shot 2017-07-09 at 4 28 55 pm

After: 

screen shot 2017-07-09 at 4 26 59 pm

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -24,33 +24,33 @@ $ npm install --save ink
 const {h, mount, Component, Text} = require('ink');
 
 class Counter extends Component {
-	constructor() {
-		super();
+  constructor() {
+    super();
 
-		this.state = {
-			i: 0
-		};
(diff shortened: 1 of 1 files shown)


sindresorhus:
The tabs are intentional. Use something like https://github.com/sindresorhus/tab-size-on-github if the default indentation size bothers you.

damassi:
To be clear, it's not the "tabs" that are the issue but the 8 characters Github converts each one into.

sindresorhus:
That's something GitHub should fix, not every project using tabs. You'll encounter this on every project using tabs, which is at least 20% of all projects on GitHub. The code uses tabs so we use it in the readme too for consistency.

PR #12, opened 2017-07-11
Add more example to readme.md

tpai:
Add example which could work without JSX transpiler.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -94,6 +94,35 @@ const {h} = require('ink');
 const Demo = () => <div/>;
 ```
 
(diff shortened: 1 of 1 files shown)


vadimdemedes:
Not sure I want to encourage that. You can use [import-jsx](https://github.com/vadimdemedes/import-jsx) if you are like me and don't like transpiling stuff.

P.S. Sorry for not getting back to you that long.

sindresorhus:
I think it could be mention, but it shouldn't take up valuable readme space. Could be in a separate recipe/tip document or something.

tpai:
Ok, I'll close this PR.

PR #14, opened 2017-07-11
Add `ink-checkbox-list` to the readme

MaxMEllon:
Thank you very nice library.

I tried to make a list of check box list.
Please merge this PR, if you'd like.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -66,6 +66,7 @@ render(<Counter/>);
 - [ink-password-input](https://github.com/vadimdemedes/ink-password-input) - Password input.
 - [ink-progress-bar](https://github.com/brigand/ink-progress-bar) - Configurable component for rendering progress bars.
 - [ink-spinner](https://github.com/vadimdemedes/ink-spinner) - Spinner.
+- [ink-checkbox-list](https://github.com/MaxMEllon/ink-checkbox-list) - Checkbox.
 
 
 ## Guide


vadimdemedes:
Thanks! Good thing that you decided to make this component, it was on my todo list!

At this moment, there are quite of a few improvements that I'd make to the component before including it in the readme:

- Correct "ink" name to "Ink" in readme
- Add tests
- Enable Travis CI and add badge to readme
- Leave only one demo gif in the readme and record it without custom terminal prompt. Also reduce the width in half to make it look good on retina
- In "Install", you don't need to install `readline`, it's a core module
- You can also skip "ink" in "npm install", since it's pretty obvious already that it should be used with Ink
- "KeyBind" could be clearer, "Key bindings" or something
- "Deside" -> "Decide" -> "Submit"
- Prioritize JSX example or remove non-JSX example altogether
- `component.getCheckedItem()` could be removed and selected values could be passed as an argument of `onSubmit()`
- Document props like [here](https://github.com/vadimdemedes/ink-text-input#props)
- Don't enable raw mode on stdin automatically, let the user do it. See the bottom of https://github.com/vadimdemedes/ink-text-input#usage
- I was under the impression that if I'm passing `<ListItem>` components, I can customize the output, but `<List>` just parses the props, essentially ignoring the `<ListItem>` component
- Don't modify `this.props` (https://github.com/MaxMEllon/ink-checkbox-list/blob/master/src/components/list.js#L8), instead use default values
- Would be nice to make the cursor circular (when pressing up, while on the first item, jump to the end) and make it configurable
- Typo https://github.com/MaxMEllon/ink-checkbox-list/blob/master/src/components/list.js#L40
- Use `keypress` events instead of `data`
- I'd use JSX in the code, it would make it much more readable
- `renderCheckbox` could be replaced with a component
- Don't put `ink` as a dependency (https://github.com/MaxMEllon/ink-checkbox-list/blob/master/package.json#L43), put it as a dev dependency instead
- package.json needs to be formatted

I realize this is too much, so let me know if you are willing to finish it. No pressure ;)

sindresorhus:
In addition:

- The pointer could be nicer. Use `pointer` in https://github.com/sindresorhus/figures and you get automatic Windows fallback. You can use that module for the other symbols too.
- One of the `checkedChar` in the docs should be `uncheckedChar`. I also don't see the point in shortening it to `char`, just do `checkCharacter` or `checkedSymbol`.
- `Type: String`  => `Type: string`
- I think it would make sense to have a `onChange` event for each uncheck/check.

MaxMEllon:
Maybe I fixed all.

vadimdemedes:
Few questions/suggestions:

- Does `onSubmit()` receive checked indexes or values? I think values would be more convenient.
- Why does `left` and `right` keys trigger the check? Does Inquirer have the same behavior?
- Could you improve the gif by removing the displaying of keystrokes? It makes the demo bit messy.
- Props in readme are different than in code.
- There's extra newline between `Type` and `Default` in props section of the readme.
- Instead of putting <kbd>Enter</kbd> and <kbd>Space</kbd> in parenthesis, you could insert a few words like "This is triggered by pressing [key]".
- Could you improve the usage example in readme by putting a meaningful list items and removing `setTimeout` to call `unmount()`? `unmount()` part is irrelevant to demonstrating this component.
- "Ink-checkbox-list" -> "ink-checkbox-list".
- "Key bindings" would be more useful after the "Usage" or even "Props".
- `nocheckedCharacter` => `uncheckedCharacter`.
- https://github.com/MaxMEllon/ink-checkbox-list/blob/b1161203cc29c4f2a2807b82c9e9f17e725c25d7/package.json#L6 could just be `"MaxMEllon/ink-checkbox-list"`.

vadimdemedes:
Also, could you please prettify the demo? See https://github.com/yeoman/yo/blob/master/screenshot.png.

MaxMEllon:
> Does onSubmit() receive checked indexes or values? I think values would be more convenient.

Will return a VNode, if Returning not index will return.
I think user need is String value.
how implements it?

vadimdemedes:
Good point. I'd add a `value` prop to `<ListItem>` similarly to `<option>` in HTML.

```js
<List>
  <ListItem value="key">Label</ListItem>
</List>
```

MaxMEllon:
- [x] Does onSubmit() receive checked indexes or values? I think values would be more convenient.
> return value of ListItem.
- Why does left and right keys trigger the check? Does Inquirer have the same behavior?
> Removed.
- [x] Could you improve the gif by removing the displaying of keystrokes? It makes the demo bit messy.
- [x] Props in readme are different than in code.
- [x] There's extra newline between Type and Default in props section of the readme.
- [x] Instead of putting Enter and Space in parenthesis, you could insert a few words like "This is triggered by pressing [key]".
- [x] Could you improve the usage example in readme by putting a meaningful list items and removing setTimeout to call unmount()? unmount() part is irrelevant to demonstrating this component.
- [x] "Ink-checkbox-list" -> "ink-checkbox-list".
- [x] "Key bindings" would be more useful after the "Usage" or even "Props".
- [x] https://github.com/MaxMEllon/ink-checkbox-list/blob/b1161203cc29c4f2a2807b82c9e9f17e725c25d7/package.json#L6 could just be "MaxMEllon/ink-checkbox-list".
- [x] Update ink@0.2.X
  - [x] Use DefaultProps
  - [x] Use PropTypes

jimthedev:
Excited for this. Is anyone working on a select dropdown?

vadimdemedes:
@jimthedev It's on my todo list :)

vadimdemedes:
There are some cosmetic things left, but overall looks good. Thanks @MaxMEllon!

jimthedev:
@vadimdemedes 

cheers

vadimdemedes:
@jimthedev my favorite gif! how did you know 😆

jimthedev:
Oh I have my ways.

psych

vadimdemedes:
@jimthedev Done - https://github.com/vadimdemedes/ink-select-input!

PR #23, opened 2017-07-22
Pass props in setState() callback

ForbesLindesay:
React passes props as the second argument to functional setState calls.


vadimdemedes:
Thanks!

Could you modify this test https://github.com/vadimdemedes/ink/blob/master/test/components.js#L464 to verify that props are being passed to `setState()` callback?

ForbesLindesay:
I've added a test for functional setState

vadimdemedes:
Thank you!

PR #38, opened 2017-08-13
Add ink-autocomplete and ink-table to Useful Components 

maticzav:
Hi! Thank you so much for the library! 

I created an autocomplete-like Text input component to search and select from list of objects.
Please, merge this PR if you like it!

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -71,7 +71,7 @@ render(<Counter/>);
 - [ink-confirm-input](https://github.com/kevva/ink-confirm-input) - Yes/No confirmation input.
 - [ink-checkbox-list](https://github.com/MaxMEllon/ink-checkbox-list) - Checkbox.
 - [ink-select-input](https://github.com/vadimdemedes/ink-select-input) - Select (dropdown) input.
-- [ink-select](https://github.com/maticzav/ink.git) - Select (autocomplete) input.
+- [ink-select](https://github.com/maticzav/ink-select.git) - Select (autocomplete) input.
 
 
 ## Guide
--- a/readme.md
+++ b/readme.md
@@ -72,6 +72,7 @@ render(<Counter/>);
 - [ink-checkbox-list](https://github.com/MaxMEllon/ink-checkbox-list) - Checkbox.
 - [ink-select-input](https://github.com/vadimdemedes/ink-select-input) - Select (dropdown) input.
 - [ink-select](https://github.com/maticzav/ink-select.git) - Select (autocomplete) input.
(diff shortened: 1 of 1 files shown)


maticzav:
First, sorry for same PR.
I am creating some more components for Ink, because I want to use Ink in my admin dashboard. So, I created Table component which I find very useful, also very open to suggestions for improvement!

Hope you like it! 😄

vadimdemedes:
> Hi! Thank you so much for the library!

Hi, you're welcome!

Regarding `ink-select`, isn't it a text autocomplete component? `ink-select` makes me think of `<select>` dropdown in HTML.

Re `ink-table`:

1. Is there an actual need to customize table borders (`crosses`, `lines` and `corners`)? Looks like it shouldn't be exposed to configuration at all. 
2. You should pass props directly, instead of passing an object of them to `config`. E.g. `<Table config={{padding: 1}}/>` => `<Table padding={1}/>`.
3. Could you describe the props the same way you did in `ink-select`?
4. The demo image doesn't represent an example code.

Aside from that, `ink-table` looks good! Thanks for making these components and keep them coming!

readme.md
```diff
@@ -71,6 +71,8 @@ render(<Counter/>);
 - [ink-confirm-input](https://github.com/kevva/ink-confirm-input) - Yes/No confirmation input.
 - [ink-checkbox-list](https://github.com/MaxMEllon/ink-checkbox-list) - Checkbox.
 - [ink-select-input](https://github.com/vadimdemedes/ink-select-input) - Select (dropdown) input.
+- [ink-select](https://github.com/maticzav/ink-select.git) - Select (autocomplete) input.
```
vadimdemedes: `.git` suffix is unnecessary.

maticzav:
ah, yes I will fix the demo image in `ink-table`, so that it matches the demo! I will also change the config as recommended, thanks!

+ Re `ink-select`, actually I am a bit unsure about where to go here, because I thought the module would be something like `react-select` in which way I should add recommendations below (something like `ink-select-input`) and accompany it with simple search. On the other hand, though, it could be more like autocomplete only, but I suppose there should be a better autocomplete (fuzzy maybe...) in that case. Looking for opinions... 😄 

Really thankful for the response!

maticzav:
I think everything should be done now. Changed `ink-select` to `ink-autocomplete` and changed functionality a bit, hope you like it!

maticzav:
Hey! Is there anything else I should add before merging?

vadimdemedes:
All good, thank you! Great modules!

PR #41, opened 2017-09-19
Add link to ink-tab

jdeniau:
I created a small `tab` component. It can be useful for others.


vadimdemedes:
I can't come up with a real use case where this component could be used. Do you have any examples?

jdeniau:
It is useful if your want to build a monitoring system or something like that.

Personally I use it in [changelog-view](https://github.com/jdeniau/changelog-view).

Your can try it by doing `npx --ignore-existing changelog-view vadimdemedes/ink@0.3.0 jdeniau/changelog-view@1.2.0 facebook/react@16.1.0` for example

Issue #39, opened 2017-08-21
Add Yeoman generators to readme


vadimdemedes:
There are 2 Yeoman generators to easily get started with Ink:

1. https://github.com/vadimdemedes/generator-ink-component - Create a reusable Ink component
2. https://github.com/vadimdemedes/generator-ink-cli - Create a CLI that uses Ink

Having them in the readme would ease the intro to Ink.

shakyink:
@vadimdemedes - I've made a [PR](#44) that adds information on these two generators to the readme.

vadimdemedes:
Fixed via #44.

PR #44, opened 2017-11-16
Adding Documentation on Generators for Ink

shakyink:
This addresses issue #39, I've added some text to the `README.md` to give a quick explanation on what each generator does and how to install them.


vadimdemedes:
I think it should be in the "Getting Started" section. And installation commands can be removed too. Just a simple sentence with two links to these repositories, they have their own docs anyway.

shakyink:
@vadimdemedes - Made the change in commit `0a65985`.

vadimdemedes:
Made a slight change. Thanks for doing this!

PR #49, opened 2017-12-29
Use svg demo

marionebl:
* Better quality
* Smaller file: 170kb => 3.5kb


marionebl:
Linting errors tripping up Travis are unrelated

vadimdemedes:
Could you add svg to this repo instead? Great work on `svg-term-cli` by the way.

marionebl:
Done, glad you like it!

vadimdemedes:
Just noticed something weird, demo doesn't loop, but bounces back. If you take a look at it, tests number increases than decreases. Like a boomerang in Instagram. Is there a way to make it loop, starting at the beginning each time like a regular gif?

marionebl:
I deliberately did the demo like this to have a "true" loop with the first and last frame showing **0**.
Can change this, no problem.

marionebl:
@vadimdemedes Should I change it to a "linear" loop?

vadimdemedes:
Yep, that could be cool.

marionebl:
Updated the demo. I also added the used demo script and command to generate a new version.

vadimdemedes:
Thank you, looking great!

PR #53, opened 2018-02-10
Add ink-router and ink-broadcast to Useful Components in README

jimmed:
I'm a huge fan of Ink, and was excited at the possibility of using familiar patterns to build command-line apps. To that end, I've built my own implementations of react-router (and accordingly react-broadcast) to make developing Ink applications even sweeter.

I'm open to pull requests, suggestions and issue reports, and I hope my contribution to this ecosystem is welcome and useful to some!

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -73,6 +73,8 @@ render(<Counter/>);
 - [ink-select-input](https://github.com/vadimdemedes/ink-select-input) - Select (dropdown) input.
 - [ink-autocomplete](https://github.com/maticzav/ink-autocomplete) - Autocomplete.
 - [ink-table](https://github.com/maticzav/ink-table) - Table component.
+- [ink-broadcast](https://github.com/jimmed/ink-broadcast) - Implementation of react-broadcast for Ink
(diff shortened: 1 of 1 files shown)


jimmed:
@vadimdemedes Any thoughts? If you don't see these libraries as being useful, please close this PR. Otherwise, I would love to share my creations with other developers using ink.

vadimdemedes:
Hey Jim, sorry for late reply, I'm too busy with my job lately. I love these, keep them coming! One thing that looks a bit weird to me is that `ink-router` still uses URLs for routing. I'd switch from `/parent/child/route` to something like `parent:child:route` to make it feel more native to CLIs. Anyway, happy to merge these!

jimmed:
Thanks for the feedback! I agree it's a little jarring; for now it's just because I depend on the [`history`](http://npm.im/history) module, and was initially more focused on getting the core functionality working.

For the moment I'm going to play around and try to get some decent examples built using it, just to really kick the tyres, and then I'll be looking at numerous ways to improve the developer experience :)

PR #55, opened 2018-03-15
replace all tabs to 2 spaces

MarcoWorms:
In `.editorconfig` there are options set to "2 spaces", so IDK if you guys agreed to go with tabs and ignores that file or if it's just a confusion, so I replaced all tabs with 2 spaces because 2 tabs makes the readability insane in github as it converts 1 tab to 4 spaces

Github readme before:
screenshot from 2018-03-15 16-06-36

After:
screenshot from 2018-03-15 16-08-03

just like the readme, all other files in github have this same behaviour when looking at the code. Again, IDK if this is intended or not, if it's not intended than here is the PR to fix it :)


ewnd9:
There is a Chrome extension to modify tabs size on GitHub https://github.com/lukechilds/github-custom-tab-size (if you are using Chrome of course)

MarcoWorms:
awesome :D, does this problem exist in other browsers?

guylyons:
Is the deletion of the Ink logo intentional as part if this commit?

MarcoWorms:
Oh, just saw it

https://github.com/vadimdemedes/ink/pull/55/files#diff-6ed4612926cab877551ba3f6da61e3ea

I probably broke the image because I used a recursive mass-replace :P

sindresorhus:
No thanks :)

PR #62, opened 2018-03-30
Add 'ink-boxen' to the README

yardnsm:
First of all, thanks for the library! Awesome work 🚀
 
I created a small component to draw boxes which uses [boxen](https://github.com/sindresorhus/boxen/) under the hood. This PR adds a link to the project in the README.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -80,6 +80,7 @@ render(<Counter/>);
 - [ink-link](https://github.com/sindresorhus/ink-link) - Link component.
 - [ink-select](https://github.com/karaggeorge/ink-select) - Select component.
 - [ink-scrollbar](https://github.com/karaggeorge/ink-scrollbar) - Scrollbar component.
+- [ink-boxen](https://github.com/yardnsm/ink-boxen) - Box component.
 
 
 ## Guide


armand1m:
also, @sindresorhus published [`ink-box`](https://github.com/sindresorhus/ink-box) today which also uses [`boxen`](https://github.com/sindresorhus/boxen)

vadimdemedes:
> First of all, thanks for the library! Awesome work 🚀

@yardnsm Thank you, I really appreciate that!

As for `ink-boxen`, sorry, I'm going to include `ink-box` instead, because it has a more comprehensive documentation and it uses `renderToString()` method to support any type of children, which makes it universal.

With that said, I have no intention of letting you down, so I've made a little research and I found another opportunity for you to make an Ink component: animated text gradients. For example, you could make a simple port of https://github.com/bokub/chalk-animation. I imagine it would look like this:

```jsx
<Rainbow>
	Text I want to be animated
</Rainbow>
```

If you don't want to implement all the effects of `chalk-animation`, you could limit it to just rainbow, call it `ink-rainbow` and it would be awesome! What do you think, would you be able to implement it?

yardnsm:
@vadimdemedes No worries, I understood.

Regarding the component you suggested, the idea sounds cool. Maybe I'll try working on it when I have some spare time. I just hope that @sindresorhus won't make something better :)

vadimdemedes:
@sindresorhus Don't you take it, @yardnsm is the one who reserved `<Rainbow>`! 🤣 Looking forward to seeing it live!

PR #64, opened 2018-04-02
Create Renderer class

### LINKED ISSUE #61: Add a renderToStream method
I'm working on a package and I need to render a component temporarily (like renderToString does) but I need the lifecycle events to occur and emit the updates to a stream. However, I don't need the stdin hooks for escape, Ctrl+c etc. Plus, when I call unmount, I don't want the program to exit. 

Right now, as a workaround, I call `render(<Component />, {stdout, stdin})` with my stdout being the stream I implement, and `stdin` being a Noop version of a Readable stream made to look like stdin. To unmount, I have to import `call-tree` directly and manually unmount, since calling the returned method of `render` ends my process.

I think implementing a method `renderToStream` that does the basic functionality without being bound to stdin/stdout is fairly easy, and then render could use that, and add the stdin interactions. 

If that sounds like a good idea, I can write a PR.
If there is an easier way to do this that I'm not seeing, let me know.

Also, I think it would make sense to be passing the updated component by itself (without the eraseLine characters that the `log-update` uses to clear the terminal)

karaggeorge:
Fixes #61

the `Renderer` class takes a tree as an argument when constructed (`new Renderer(tree)`). It implements `EventEmitter` and emits an `update` event every time the tree updates. 

This makes creates an easy way to render a component with its lifecycle and get updates without involving stdin/stdout. 

The classic render method is re-written using this class, and calls the log-update's `log` with the updated component on `update`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -9,6 +9,7 @@ const diff = require('./lib/diff');
 const h = require('./lib/h');
 const Indent = require('./lib/components/indent');
 const Text = require('./lib/components/text');
+const Renderer = require('./lib/renderer');
 
 exports.StringComponent = StringComponent;
 exports.Component = Component;
@@ -41,32 +42,15 @@ exports.render = (tree, options) => {
 
 	const log = logUpdate.create(stdout);
 
-	const context = {};
-	let isUnmounted = false;
-	let currentTree;
-
 	readline.emitKeypressEvents(stdin);
 
 	if (stdin.isTTY) {
 		stdin.setRawMode(true);
 	}
 
-	const update = () => {
-		const nextTree = build(tree, currentTree, onUpdate, context); // eslint-disable-line no-use-before-define
-		log(renderToString(nextTree));
-
-		currentTree = nextTree;
-	};
-
-	const onUpdate = () => {
-		if (isUnmounted) {
-			return;
-		}
-
-		update();
-	};
-
-	update();
+	const currentTree = new Renderer(tree);
+	currentTree.on('update', log);
+	currentTree.update();
 
 	const onKeyPress = (ch, key) => {
 		if (key.name === 'escape' || (key.ctrl && key.name === 'c')) {
@@ -76,7 +60,7 @@ exports.render = (tree, options) => {
 
 	if (stdin.isTTY) {
 		stdin.on('keypress', onKeyPress);
-		stdout.on('resize', update);
+		stdout.on('resize', currentTree.update);
 	}
 
 	const consoleMethods = ['dir', 'log', 'info', 'warn', 'error'];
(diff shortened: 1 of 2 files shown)


vadimdemedes:
Your use case is definitely valid and I want to fix it, but I think it could be done in a simpler way. So basically what you need is to disable stdin handling and exiting the program after unmount. What do you think about this:

```jsx
render(<App/>, {
	handleStdin: false,
	exitOnUnmount: false
});
```

karaggeorge:
@vadimdemedes 

That was my initial thought, but then you also have to deal with creating a stream object and the mess of the `log-update` which will prepend the ansi characters to clean the previous output from the terminal. So you have to clean those, which is not as simple as removing all ansi characters, since you want the colors to stay.

That's why I figured implementing a simple `.on('update', updatedComponent => doSomething())` would make it much simpler to work with.

Thoughts?

karaggeorge:
@vadimdemedes 

To explain what I was saying above, in the component I am creating, in order to get the same results as this class would get me by overwritting the streams, I have to do this:
- Extend a WrittableStream:
```
class AlignedStream extends stream.Writable {
  constructor(cb) {
    super();
    this.cb = cb;
  }

  _write(chunk, enc, next) {
    this.cb(chunk.toString());
    next();
  }
}
```
- Render the component using that as stdout:
```
const stream = new AlignedStream(updated => this.updateComponent(index, stripAnsi(updated)));
    const component = _.cloneDeep(c);
    render(component, {stdout: stream, stdin: new NoopStream()});
```
- Strip ansi characters when the update is called (since it's wrapped from log-update):
`updated => this.updateComponent(index, stripAnsi(updated))`
using a specific regex patern to only target the erase type of characters:
`const stripAnsi = s => s.replace(new RegExp('\\u001B[[\\]()#;?\\d]*[A-Z]', 'g'), '');`
(modified from https://github.com/chalk/ansi-regex/blob/master/index.js)

That's why I think it would be much easier to just do this instead:
```
const component = new Renderer(_.cloneDeep(c));
component.on('update', this.updateComponent)
```

vadimdemedes:
I see your point, makes sense. Out of curiosity, what kind of CLI you're building that doesn't need to replace previous output (the default now, `log-update`), but keep appending output to the stream? Really interested to know the context here 🙂

karaggeorge:
I'm making a package called `ink-aligned` which renders its direct children next to each other (side to side) instead of one above the other as it would normally. To do that, I need to temporarily render all the components, but display them line by line manually. I tried using `renderToString` at first, but then there is no lifecycle or state/props.

Example of how it works with all the workarounds:
kapture 2018-03-30 at 0 26 00

Using the left-right arrows to change which `ink-select-input` component has `focus: true`

vadimdemedes:
I see! Thanks for working on this and explaining your issue in great detail.

PR #68, opened 2018-04-05
Add 'ink-text-animation' to the README

yardnsm:
I created a small component that can animate text, which uses [chalk-animation](https://github.com/bokub/chalk-animation) under the hood. This PR adds a link to the project in the README.


vadimdemedes:
Awesome, thanks @yardnsm! By the way, to avoid trickery with working around `chalk-animation` using `log-update` to display animation (https://github.com/yardnsm/ink-text-animation/blob/master/src/components/TextAnimation.jsx#L57-L63), you might want to copy the code from `chalk-animation` directly to handle the animation in your component: https://github.com/bokub/chalk-animation/blob/master/index.js#L28. Or, even better, you could submit a PR to `chalk-animation` and fix this 🙂

Issue #65, opened 2018-04-03
Rename `<Text>` component to `<Color>`


sindresorhus:
It makes it sound like it can only color text, but it works on any Ink component, so you could use it to color [`ink-box`](https://github.com/sindresorhus/ink-box), for example.

We can add separate components for the existing `<Text bold>` and `<Text underline>` styling: `<Bold>` and `<Underline>`.

karaggeorge:
I can work on a PR for this if you want

sindresorhus:
@karaggeorge Sure :)

karaggeorge:
I think this can be closed now 😄 https://github.com/vadimdemedes/ink/pull/69

PR #69, opened 2018-04-05
Rename Text to Color

karaggeorge:
Implements https://github.com/vadimdemedes/ink/issues/65

Renames the Text component to Color since it can be used to wrap any type of component. Also adds Bold and Underline components for those styles that can be achieved with chalk.

Also updates the tests and adds tests for those new components.


vadimdemedes:
Thanks @karaggeorge!

PR #70, opened 2018-04-05
Support JSX fragments

### LINKED ISSUE #46: Support JSX fragments
https://reactjs.org/blog/2017/11/28/react-v16.2.0-fragment-support.html

karaggeorge:
Fixes #46

This adds support for using the new React Fragments `<></>`. 
However: `plugin-transform-react-jsx` only supports them from babel v7 and on, which has no moved to `@babel/plugin-transform-react-jsx`, which requires `@babel/core`, which would mean `@babel/register` should be used instead of `babel-register`. Again, in order to use babel 7, ava had to be bumped to use the new `1.0.0-beta` versions. 

Coulnd't get xo to play nice with the fragments. I kept getting this error:
```
test/components.js:672:6
  ✖  672:6  Parsing error: Unexpected token >  
```
Maybe @sindresorhus can help me figure out how to get xo to recognize them.

This would also mean that every package that uses ink, should also upgrade to the new babel versions. 

A less invasive way of introducing fragments would be to ommit the shorthand `<></>` syntax and jsut add a `<Fragment></Fragment>` component which would basically be a `span`. That wouldn't require any new package versions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,13 +48,14 @@
 		"@babel/plugin-transform-react-jsx": "^7.0.0-beta.44",
 		"@babel/register": "^7.0.0-beta.44",
 		"ansi-styles": "^3.1.0",
+		"ava": "^1.0.0-beta.3",
+		"babel-eslint": "^8.2.2",
 		"eslint-config-xo-react": "^0.13.0",
 		"eslint-plugin-react": "^7.1.0",
 		"sinon": "^2.3.4",
 		"strip-ansi": "^4.0.0",
 		"svg-term-cli": "^2.1.1",
-		"xo": "^0.18.2",
-		"ava": "^1.0.0-beta.3"
+		"xo": "^0.18.2"
 	},
 	"ava": {
 		"require": [
@@ -87,12 +88,13 @@
 				"pragmaFrag": "h.Fragment"
 			}
 		},
-		"ignores": ["test/components.js"],
+		"parser": "babel-eslint",
 		"overrides": [
 			{
 				"files": "test/*.js",
 				"rules": {
-					"react/prop-types": 0
+					"react/prop-types": 0,
+					"quotes": 0
 				}
 			}
 		]
--- a/package.json
+++ b/package.json
@@ -44,33 +44,37 @@
 		"prop-types": "^15.5.10"
 	},
 	"devDependencies": {
+		"@babel/core": "^7.0.0-beta.44",
+		"@babel/plugin-transform-react-jsx": "^7.0.0-beta.44",
+		"@babel/register": "^7.0.0-beta.44",
 		"ansi-styles": "^3.1.0",
-		"ava": "^0.21.0",
-		"babel-plugin-transform-react-jsx": "^6.24.1",
-		"babel-register": "^6.24.1",
 		"eslint-config-xo-react": "^0.13.0",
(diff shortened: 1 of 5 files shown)


vadimdemedes:
Could we go with only `<Fragment>` for now to keep it clean and until `<></>` is widely used?

Thanks for all of your efforts by the way, hugely appreciated and keep them coming! 🙂

sindresorhus:
> Could we go with only <Fragment> for now to keep it clean and until <></> is widely used?

Why? `<>` is the future. I don't see the point in adding `<Fragment>`.

vadimdemedes:
Didn't know it's already widely used, I'm still using `<Fragment>` at work. Let's go with it then!

package.json
```diff
@@ -44,33 +44,37 @@
...
-		"babel-plugin-transform-react-jsx": "^6.24.1",
-		"babel-register": "^6.24.1",
 		"eslint-config-xo-react": "^0.13.0",
 		"eslint-plugin-react": "^7.1.0",
 		"sinon": "^2.3.4",
 		"strip-ansi": "^4.0.0",
 		"svg-term-cli": "^2.1.1",
-		"xo": "^0.18.2"
+		"xo": "^0.18.2",
+		"ava": "^1.0.0-beta.3"
```
vadimdemedes: `ava` should come after `ansi-styles`, could you reinstall it so that npm puts it in the right place?

package.json
```diff
@@ -79,9 +83,11 @@
 		],
 		"settings": {
 			"react": {
-				"pragma": "h"
+				"pragma": "h",
+				"pragmaFrag": "h.Fragment"
 			}
 		},
+		"ignores": ["test/components.js"],
```
vadimdemedes: Why do we need to ignore it?

karaggeorge: I couldn't get xo to pass like I mentioned, but I wanted to make sure the tests were passing. When I get xo working I'll put it back 😄

karaggeorge:
@vadimdemedes sounds good, I'll make the updates and clean it up 👍 

@sindresorhus 
> use the XO parser option and use babel-eslint instead.
I'll try that
> needs some docs updates.
Of course. I just wanted to see what direction we'll head in before updating everything

package.json
```diff
@@ -79,14 +84,17 @@
...
 			}
 		},
+		"parser": "babel-eslint",
 		"overrides": [
 			{
 				"files": "test/*.js",
 				"rules": {
-					"react/prop-types": 0
+					"react/prop-types": 0,
+					"quotes": 0
```
karaggeorge: https://github.com/babel/babel-eslint/issues/554

karaggeorge:
@vadimdemedes @sindresorhus 

Give this another look when you get the chance. It should be good

readme.md
```diff
@@ -537,6 +525,50 @@ const Demo = (
 );
 ```
 
+Ink also supports Fragments for returning multiple children from a component's render method.
+
+```jsx
+render(
+	<h.Fragment>
```
vadimdemedes: Could you export `Fragment` on the root export instead of `h`, so that it becomes:

```jsx
import { render, Fragment } from 'ink';

<Fragment>
	<A/>
	<B/>
	<C/>
</Fragment>
```

vadimdemedes:
This is awesome, thank you @karaggeorge!

PR #81, opened 2018-05-08
context passed in contructor

zeakd:
This is bug fix PR.

``` js
class Outer extends Component {
  getChildContext () {
    return { name: 'Joey' };
  }

  render () {
    return <Inner />
  }
}

class Inner extends Component {
  constructor (...args) {
    super(...args);
    
    console.log(this.context.name) // FIX: undefined => 'Joey'
  }

  render () {
    render this.context.name;
  } 
}
```

### DIFF
--- a/lib/diff.js
+++ b/lib/diff.js
@@ -29,7 +29,7 @@ const mount = (vnode, context, onUpdate) => {
 	checkPropTypes(vnode.component, props);
 
 	if (isClassComponent(vnode.component)) {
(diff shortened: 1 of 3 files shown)


vadimdemedes:
React doesn't pass context to constructor anymore: https://reactjs.org/docs/react-component.html#constructor, that's why I don't think it should be in Ink either, since eventually I want Ink to have API parity with React, like Preact does.

Issue #88, opened 2018-05-22
State updates overwrite each other


mgrip:
Seems like state updates that happen before the component has a chance to render will overwrite each other

PR #89, opened 2018-05-22
fix: queued state updates overwriting each other

mgrip:
fixes https://github.com/vadimdemedes/ink/issues/88


vadimdemedes:
Thanks!

Issue #100, opened 2018-10-19
memory leak 😨 / how can I assist with v2?


rozzzly:
Hey @vadimdemedes! I really like this project, and have already started to use the `@next` branch in a project of mine but have run into several issues. I would like to help with the v2 rewrite.

The major issue is a memory leak. You use yoga which (for node) is compiled from C++ into a native add on. As with anything native, memory management is a vital concern and it appears the current implementation does not properly account for that. For every new react/dom node, a new [yoga node is created](https://github.com/vadimdemedes/ink/blob/next/src/create-renderer.js#L7) but when the react/dom node is removed [here](https://github.com/vadimdemedes/ink/blob/next/src/create-reconciler.js#L50) and [here](https://github.com/vadimdemedes/ink/blob/next/src/create-reconciler.js#L76) nothing is done to free up the allocated object. I believe this particular fix would be as simple as calling `child.yogaNode.free()` after `parent.removeChild(child)`. 

However, every time you [render](https://github.com/vadimdemedes/ink/blob/next/src/create-renderer.js#L101) the tree, a new yoga node is created for _every item_ in the tree and any old yoga nodes are left out in limbo.

I've found several other smaller issues with the implementation and would like to PR them.
 
I just wanted to A) make sure you were aware of these issues B) get your blessing C) discuss how I could help with the rewrite in general. For example: I think the dependency on `logUpdate` should be dropped and a lower-level implementation written that updates just the rows (or even columns) affected by a rerender; right now, it's redrawing the entire screen, even the unchanged parts—this causes me flickers.

vadimdemedes:
Thanks, glad you like the project!

Regarding memory leak, yes I'm aware of this issue. There's a simpler way to free up all nodes - `yogaNode.freeRecursive()`. I didn't add it, because `<Box>` has this method https://github.com/vadimdemedes/ink/blob/v2.0.0-2/src/components/Box.js#L52, which returns actual width of component. It's useful for things like progress bar, where you need to calculate the percentage of component width. Not sure how to work around it yet, what do you think?

> logUpdate should be dropped and a lower-level implementation written that updates just the rows (or even columns) affected by a rerender
I've tried that actually, there's a module exactly for this - https://github.com/mafintosh/ansi-diff-stream. Unfortunately, while upgrading [emoj](https://github.com/sindresorhus/emoj) to Ink 2, I noticed that this module sometimes renders spaces incorrectly. If you know of an alternative, happy to try it too!

[...]

PR #104, opened 2018-11-14
Next: Use reconciler api to create text nodes

littldr:
Hi,

first thanks for your great work! :)

While playing around with the `next` branch I've found that text nodes are not rendered correctly when they are the only child.
E.g. the spinner (copied from `ink-spinner` and made compatible with `next`, I will create a separate PR for this)
```javascript
"use strict";

const { Color, render } = require("ink");
const React = require("react");
const spinners = require("cli-spinners");
const omit = require("object.omit");

class Spinner extends React.Component {
	constructor() {
		super();
		this.state = { frame: 0 };
		this.switchFrame = this.switchFrame.bind(this);
	}

	getSpinner() {
		return spinners[this.props.type] || spinners.dots;
	}

	render() {
		const colorProps = omit(this.props, "type");
		const spinner = this.getSpinner();
		const frame = spinner.frames[this.state.frame];
		return <Color {...colorProps}>{frame}</Color>;
	}

	componentDidMount() {
		const spinner = this.getSpinner();
		this.timer = setInterval(this.switchFrame, spinner.interval);
	}

	componentWillUnmount() {
		clearInterval(this.timer);
	}

	switchFrame() {
		const { frame } = this.state;
		const spinner = this.getSpinner();
		const isLastFrame = frame === spinner.frames.length - 1;
		const nextFrame = isLastFrame ? 0 : frame + 1;

		this.setState({ frame: nextFrame });
	}
}

render(<Spinner>);
```

It prints:
```
⠋⠹⠙⠹⠸⠼⠴⠦⠧⠇⠏⠋⠙⠹⠸⠼⠴⠦⠧⠇⠏⠋⠙⠹⠸⠼⠴⠦⠧⠇⠏⠋⠙⠹...
# (the output is growing)
```

This only happens, when there is only one child with type string. If we change the render method to `return <Color {...colorProps}>{frame}{frame}</Color>;` (two childs with type string), everything works as expected.

I found out that the implemented `reconciler` logic creates and adds always a new childNode, if the `children` property is of type string (or number).

After digging some time in the react repo I also found out that if you really want to wrap all text nodes in another node (as the comment is saying in the `create-reconciler.js#15` file), you should disable the `shouldSetTextContent` hook and use `createTextInstance` to create these text nodes in the host config.

I'm not sure if this will break something or not. So please review these changes carefully. :)

### DIFF
--- a/src/create-reconciler.js
+++ b/src/create-reconciler.js
@@ -10,14 +10,7 @@ import {
 
 const applyProps = (node, props) => {
 	for (const [key, value] of Object.entries(props)) {
-		if (key === 'children') {
-			if (typeof value === 'string' || typeof value === 'number') {
-				// Text node must be wrapped in another node, so that text can be aligned within container
-				const textElement = createNode('span');
-				textElement.textContent = value;
-				appendChildNode(node, textElement);
-			}
-		} else if (key === 'style') {
+		if (key === 'style') {
 			Object.assign(node.style, value);
 		} else if (key === 'unstable__transformChildren') {
 			node.unstable__transformChildren = value; // eslint-disable-line camelcase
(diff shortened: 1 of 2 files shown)


rozzzly:
Thanks for PRing what I documented in https://github.com/vadimdemedes/ink/issues/100#issuecomment-432436650

vadimdemedes:
Released `2.0.0-9` which fixes this regression.

PR #105, opened 2018-11-16
#Next #WIP Yoga Node Border

skorfmann:
I was playing around with printing borders and was wondering if that would match conceptually how you would implement it? 

According to the [Yoga docs](https://yogalayout.com/docs/margins-paddings-borders/), the actual implementation of borders is up to the higher level framework.

An implementation would look like this:

```jsx
      <Box padding={2} border={1}>
        <Color green>
          <Text underline>Yeah Foo</Text>
        </Color>
      </Box>
```

It's really a rough sketch and is not meant to be merged at this stage. Would love to get some feedback and start a discussion how printing borders could be implemented :)


vadimdemedes:
I would actually recommend using something like https://github.com/sindresorhus/ink-box. I wanted to postpone borders a bit and see if there are any Yoga layout issues first :)

skorfmann:
> I would actually recommend using something like https://github.com/sindresorhus/ink-box.

That's what we tried (or to be more precise, use the underlying Boxen as a custom component) Unfortunately it did handle only text properly. Components like a spinner weren't working. And one of the recent commits did break it entirely:

screenshot 2018-11-19 at 20 49 33

Anyway, looking at the implementation of [Boxen](https://github.com/sindresorhus/boxen), it has many redundant features which ink#next provides already and hence seems to be superfluous.

### Our Use-Case

Perhaps let's start with our use-case. We're after something like this:

```jsx
  <Box flexDirection="column">
    <Box justifyConten="flex-start" padding={2}>
      <Box padding={2} border={1}>
        <Color green>
          <Spinner /> <Text underline>Spinning Text</Text>
        </Color>
      </Box>
    </Box>
  </Box>
```

spinner

We would probably have multiple of these boxes aligned across multiple rows.

```jsx
<Box flexDirection="column">
    <Box justifyConten="flex-start" padding={2} borderBottom={1}>
      <Box padding={2} border={1}>
        <Color green>
          <Spinner />
          <Text underline>Spinning Text</Text>
        </Color>
      </Box>
      <Box padding={2} border={1}>
        <Color green>
          <Spinner />
          <Text underline>Spinning Text</Text>
        </Color>
      </Box>
    </Box>
    <Box justifyConten="flex-start" padding={2} borderBottom={1}>
      <Box padding={2} border={1}>
        <Color green>
          <Spinner />
          <Text underline>Spinning Text</Text>
        </Color>
      </Box>
      <Box padding={2} border={1}>
        <Color green>
          <Spinner />
          <Text underline>Spinning Text</Text>
        </Color>
      </Box>
      <Box padding={2} border={1}>
        <Color green>
          <Spinner />
          <Text underline>Spinning Text</Text>
        </Color>
      </Box>
    </Box>
  </Box>
```

2018-11-19 at 21 12

It seems rather natural to me, to just use the "native" box layout components rather than falling back to an additional component (and of course, use [cli-boxes](https://www.npmjs.com/package/cli-boxes) for proper border styles)

I'd be more than happy to help to make this feature happen :)

skorfmann:
Just as a heads-up: I added some commits to make the borders look nice - Still rather rough implementation wise.

Plus, `padding` was set as `margin` on the yogaNode - Not sure if that was intentionally or not. Either way, I fixed it since it makes a difference with borders.

screenshot 2018-11-22 at 22 18 50

I'm still keen on feedback :)

icopp:
Any particular reason this was closed? I've run into the same problems with `ink-box`, and I also think it's kind of silly to stick to the idea of a separate component with different behavior just to get borders.

vadimdemedes:
I thought I commented on it, my bad. Before release I did a lot of changes, which caused large merge conflicts in this PR. The plan is to integrate `ink-box`'s borders into Ink soon.

PR #111, opened 2018-12-07
Update readme.md

MartinPham:
I'd like to list also my project, which brings crossplatform rendering for React. Thanks.


msikma:
A small request (or nitpick?) but, do you think it might be an idea to use this PR to also convert the tabs in the examples to spaces? Not that it matters what anyone's code uses of course, but it's just that on Github the tabs expand to a length of 8 characters and that's a bit much.

vadimdemedes:
@MartinPham Do you still maintain this project? Looking at the history last activity was 2 months ago.

PR #117, opened 2019-01-29
fixed component

0xflotus:


vadimdemedes:
I have to close this PR, because it's targeting Ink 1, which isn't maintained anymore and will be replaced by Ink 2 asap. Thanks though!

PR #118, opened 2019-01-29
Add live demo to README.md

jankeromnes:
Hi, thanks for making Ink! It's a really cool project.

I work on [gitpod.io](https://gitpod.io), and I thought you might like to have a live demo in your README.md, so I created one for you. Gitpod provides free online workspaces, and I've configured it to initialize this project with `npm install`, open `media/demo.js` in the editor, and run the demo with `node media/demo.js`.

Here is what it looks like:

[Open in Gitpod](https://gitpod.io#snapshot/0e3fcd17-f401-42b8-aee4-ff64a957ac63)
screenshot 2019-01-29 at 13 42 11

Please let me know what you think. 🙂

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,0 +1,3 @@
+tasks:
+- init: npm install
+  command: node media/demo.js
--- a/readme.md
+++ b/readme.md
@@ -60,6 +60,12 @@ render(<Counter/>);
 	<img src="media/demo.svg" width="600">
 </p>
 
+## Live Demo
+
+Play with the example above in a free pre-configured online workspace:
+
+[![Open in Gitpod](https://gitpod.io/button/open-in-gitpod.svg)](https://gitpod.io/#https://github.com/vadimdemedes/ink/blob/master/media/demo.js)
+
 ## Useful Components
 
 - [ink-redux](https://github.com/vadimdemedes/ink-redux) - Redux bindings.


vadimdemedes:
Interesting project, thank you for setting it up! Is it possible to make this pod public? I imagine people would be unwilling to connect via GitHub in an app they don't know only to see a demo of Ink.

jankeromnes:
Hi @vadimdemedes, thanks for the feedback! Unfortunately, there is no way currently to make a pod completely public. The reason Gitpod needs a verified email address (that's all it asks for by the way, repository rights are not required) is that the demo is a full Ubuntu machine where users can do all sorts of things (including bad ones).

vadimdemedes:
I see, that makes sense then. Unfortunately, I still think public access is an important feature to let people quickly "check it out" and not everyone would want to oauth. Sorry, but I think I will go with https://repl.it, I tried it out and it works perfectly for my use case. I wish good luck with your project though, don't let this put you down!

jankeromnes:
Hi @vadimdemedes, thanks a lot for your kind reply. Public access does make a lot of sense for live demos, and I think it's really cool that Ink works in Repl.it -- it's gotten much better since last time I tried it!

Maybe Gitpod makes more sense as a way to automate dev environments rather than as a live demo. If you think it could make contributions to Ink itself easier, I'd be happy to set it up, but otherwise good luck with your project too!

Issue #115, opened 2019-01-27
ink@next: TypeError: Cannot read property 'length' of undefined


sdgluck:
I am receiving this error after updating my app to use `ink@next`, and moving from `ink-redux` to `redux` with `react-redux`:

```sh
  TypeError: Cannot read property 'length' of undefined
      at buildLayout (C:\_dev\htdocs\wpk-cli\wpk\packages\command-build\node_modules\ink\build\create-renderer.js:92:23)
      at buildLayout (C:\_dev\htdocs\wpk-cli\wpk\packages\command-build\node_modules\ink\build\create-renderer.js:100:11)
      at buildLayout (C:\_dev\htdocs\wpk-cli\wpk\packages\command-build\node_modules\ink\build\create-renderer.js:51:31)
      at node (C:\_dev\htdocs\wpk-cli\wpk\packages\command-build\node_modules\ink\build\create-renderer.js:221:9)
      at onRender (C:\_dev\htdocs\wpk-cli\wpk\packages\command-build\node_modules\ink\build\render.js:65:9)
      at invokeFunc (C:\_dev\htdocs\wpk-cli\wpk\packages\command-build\node_modules\lodash.throttle\index.js:160:19)
      at trailingEdge (C:\_dev\htdocs\wpk-cli\wpk\packages\command-build\node_modules\lodash.throttle\index.js:207:14)
      at Timeout.timerExpired [as _onTimeout] (C:\_dev\htdocs\wpk-cli\wpk\packages\command-build\node_modules\lodash.throttle\index.js:195:14)
      at ontimeout (timers.js:498:11)
      at tryOnTimeout (timers.js:323:5)
      at Timer.listOnTimeout (timers.js:290:5)
```

Not sure what other information to provide, so please let me know if there is anything.

Code for the refactor of `ink` -> `ink@next` is here in case that is useful: https://github.com/wpk-cli/wpk/blob/ink-next/packages/command-build/src/terminal/index.js

sdgluck:
Looks like there is an assumption that the property `node.childNodes` always exists, but in this instance it doesn't. Seems like a bug. I can put in a PR with a fix if you agree.

zkat:
@vadimdemedes I can confirm I'm getting this error, too.

Repro steps:

1. `git clone https://github.com/npm/libpickupdates && cd libpickupdates`
2. `npm i && npm run build`
3. `node index.js` and continue without selecting any packages to update
4. crash is reproduced, every time

[...]

PR #121, opened 2019-02-08
Renderer respects that node doesn't have to have child nodes.

tomasbonco:
I was having an issue with Text nodes not having `childNodes`.

### DIFF
--- a/src/create-renderer.js
+++ b/src/create-renderer.js
@@ -67,7 +67,7 @@ const buildLayout = (node, options) => {
 		yogaNode.setHeight(style.height);
 	}
 
-	if (node.childNodes.length > 0) {
+	if (node.childNodes && node.childNodes.length > 0) {
 		const childNodes = node.childNodes.filter(childNode => {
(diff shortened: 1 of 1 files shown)


vadimdemedes:
Interesting, would you be able to share the code that triggers this bug?

zkat:
@vadimdemedes https://github.com/vadimdemedes/ink/issues/115#issuecomment-464829096

zkat:
The error seems to happen if you try to render an empty string when using `{''}` -- so `<Text>{''}</Text>` will totes repro this.

vadimdemedes:
Hey @tomasbonco, thanks for PR! I've merged your changes manually, because I've reorganized files a bit and merge conflict was quite big.

tomasbonco:
Guys, thanks for resolving this without me. I was really busy and didn't had time to try and reproduce the bug.

vadimdemedes:
No worries at all, thanks for submitting original PR!

Issue #138, opened 2019-03-07
Add support for keypress events


sindresorhus:
To be able to have the component rerendered on keypresses and get the keys.

Ink would need to do `readline.emitKeypressEvents(stdin);` in https://github.com/vadimdemedes/ink/blob/master/src/components/App.js#L61 And it requires raw-mode.

After enabling the above, Ink could do something like this internally:

```js
process.stdin.on('keypress', (str, key) => {
    if (key.ctrl && key.name === 'f') {
        console.log('foo')
    }
})
```

My initial idea was to expose a `useKeypress()` React hook. @vadimdemedes suggested we could just expose the keypresses by default if raw-mode is activated, as the output is being diffed anyway, so the cost is negligible.

jedahan:
what im doing currently as an example:

```js
const useKeyHandler = keyHandler => {
  const { stdin, setRawMode } = useContext(StdinContext)

  useEffect(() => {
    setRawMode(true)
    stdin.on('data', keyHandler)
    return () => {
      stdin.off('data', keyHandler)
      setRawMode(false)
    }
  }, [stdin, setRawMode])
}
...snip
 // handle keypresses
  useKeyHandler(data => {
    if (data === 's') setSyncing(syncing => !syncing)
    if (data === 'p') setCreature(ascii())
    if (data === 'q') process.exit(0)
  })
```

from https://github.com/jedahan/creature/blob/ink/src/App.jsx

jedahan:
The suggestion to just expose the handler seems nicer for those coming from nodeland - but i wonder if you have different keys in different contexts if the reactish way helps with multi-screen apps...

SimenB:
Not sure if needed, but I have a use case for this. I've been looking at rewriting https://github.com/jest-community/jest-watch-typeahead to use `ink`.

The thing I want to achieve is this:
![demo](https://user-images.githubusercontent.com/1404810/53986008-56585f80-411d-11e9-89bc-1fc9d15be2f5.gif)

So an text input, and below it a list over hits. _If_ the user hits the down arrow, then I want to select from the list.

It might be enough for https://github.com/vadimdemedes/ink-select-input to just have a mode for "none selected" instead of starting with the first in the list as "selected"? But I _think_ I can also listen for keypresses myself and only make the select I render `focus={true}` if arrow down is pressed. I also need to know when `enter` is used for the text input

vadimdemedes:
[...]

PR #147, opened 2019-03-08
initial useKeypress

jedahan:
first draft of #138 

**todo**
- [x] fix typescript definition (never written typescript before)
- [x] run npm test
~- [ ] write documentation~
- [x] write example
- [x] write some tests

### DIFF
--- a/examples/usekeypress/usekeypress.js
+++ b/examples/usekeypress/usekeypress.js
@@ -1,0 +1,42 @@
+'use strict';
+const {useState} = require('react');
+const React = require('react');
+const {render, useKeypress, Box} = require('../..');
+
+const {exit} = process;
+
+const Robot = () => {
+	const [x, setX] = useState(1);
+	const [y, setY] = useState(1);
+
+	useKeypress(key => {
+		switch (key) {
+			case 'h':
+				setX(Math.max(1, x - 1));
+				break;
+			case 'l':
+				setX(Math.min(80, x + 1));
+				break;
+			case 'j':
+				setY(Math.min(40, y + 1));
+				break;
+			case 'k':
+				setY(Math.max(1, y - 1));
+				break;
+			case 'q':
+				exit();
+				break;
+			default:
+				break;
+		}
+	});
+
+	return (
+		<Box flexDirection="column">
+			<Box>Use h, j, k, and l to move the face. q to exit</Box>
+			<Box marginTop={y} marginLeft={x}>^_^</Box>
+		</Box>
+	);
+};
+
+render(<Robot/>);
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,4 +1,5 @@
 import * as React from "react";
+import { Key } from 'readline';
 
 export interface RenderOptions {
 	/**
@@ -43,6 +44,14 @@ export type Instance = {
 };
 
 export type Unmount = () => void;
+
+/**
+ * Hook that calls keyPressHandler callback with whatever key has been pressed.
+ * See key to handle modifiers correctly.
+ */
+export function useKeypress(
+	keyPressHandler: (str?: string, key?: Key) => void
+): void;
 
 /**
  * Mount a component and render the output.
--- a/src/hooks/useKeypress.js
+++ b/src/hooks/useKeypress.js
@@ -1,0 +1,15 @@
+import {useEffect, useContext} from 'react';
+import {StdinContext} from '..';
+
+export default keypressHandler => {
+	const {stdin, setRawMode} = useContext(StdinContext);
+
+	useEffect(() => {
+		setRawMode(true);
+		stdin.on('keypress', keypressHandler);
+		return () => {
+			stdin.off('keypress', keypressHandler);
+			setRawMode(false);
+		};
+	}, [stdin, setRawMode, keypressHandler]);
+};
--- a/test/fixtures/handles-keypresses.js
+++ b/test/fixtures/handles-keypresses.js
@@ -1,0 +1,22 @@
+'use strict';
+const React = require('react');
+const {render, useKeypress} = require('../..');
+
+const input = new Set('abcdefghijklmnopqrstuvwxyz'.split(''));
+
+const KeypressTest = () => {
+	useKeypress(str => {
+		input.delete(str);
+		if (input.size === 0) {
+			app.unmount();
+		}
+	});
+	return null;
+};
+
+const app = render(<KeypressTest/>);
+
+(async () => {
+	await app.waitUntilExit();
+	console.log('exited');
+})();
--- a/examples/usekeypress/index.js
+++ b/examples/usekeypress/index.js
@@ -1,0 +1,2 @@
+'use strict';
+require('import-jsx')('./usekeypress');
--- a/package.json
+++ b/package.json
@@ -118,7 +118,10 @@
 		],
 		"overrides": [
 			{
-				"files": "src/components/*.js",
+				"files": [
+					"src/components/*.js",
+					"src/hooks/*.js"
+				],
 				"rules": {
 					"unicorn/filename-case": "off",
 					"react/require-default-props": "warning"
--- a/src/index.js
+++ b/src/index.js
@@ -6,3 +6,4 @@ export {default as AppContext} from './components/AppContext';
 export {default as StdinContext} from './components/StdinContext';
 export {default as StdoutContext} from './components/StdoutContext';
 export {default as Static} from './components/Static';
+export {default as useKeypress} from './hooks/useKeypress';
--- a/test/exit.js
+++ b/test/exit.js
@@ -47,6 +47,48 @@ test('exit with thrown error', async t => {
 	t.true(output.includes('errored'));
 });
 
+test.cb('handles keypresses', t => {
+	const term = spawn('node', ['./fixtures/run', './handles-keypresses'], {
+		name: 'xterm',
+		cols: 100,
+		cwd: __dirname,
+		env: process.env
+	});
+
+	let output = '';
+
+	term.on('data', data => {
+		output += data;
+	});
+
+	let isExited = false;
+
+	term.on('exit', code => {
+		isExited = true;
+
+		if (code === 0) {
+			t.true(output.includes('exited'));
+			t.pass();
+			t.end();
+			return;
+		}
+
+		t.fail();
+		t.end();
+	});
+
+	setTimeout(() => {
+		t.false(isExited);
+		'abcdefghijklmnopqrstuvwxyz'.split('').forEach(key => term.write(key));
+	}, 100);
+
+	setTimeout(() => {
+		term.kill();
+		t.fail();
+		t.end();
+	}, 5000);
+});
+
 test.cb('don\'t exit while raw mode is active', t => {
 	const term = spawn('node', ['./fixtures/run', './exit-double-raw-mode'], {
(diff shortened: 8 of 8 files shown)


sindresorhus:
This needs to handle the case if raw mode is already enabled when this hook is used, it should not be disabled when this hook ends.

vadimdemedes:
That's actually going to be handled by `setRawMode()` itself, so hook doesn't need to re-implement that logic :)

src/hooks/useKeypress.js
```diff
@@ -0,0 +1,17 @@
...
+  const { stdin, setRawMode } = useContext(StdinContext)
+
+  useEffect(() => {
+    setRawMode(true)
+    stdin.on('keypress', keypressHandler)
+    return () => {
+      stdin.off('keypress', keypressHandler)
+      setRawMode(false)
+    }
+  }, [stdin, setRawMode])
```
SimenB: need to add `keypressHandler` as well to the dependency array

index.d.ts
```diff
@@ -44,6 +44,8 @@ export type Instance = {
 
 export type Unmount = () => void;
 
+export function useKeypress<T extends (str?: string, key?: any) => null>(keyPresshandler: T): null;
```
SimenB: type for `key`: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/de6730f4463cba69904698035fafd906a72b9664/types/node/readline.d.ts#L5-L11

You can import it by `import {Key} from 'readline';`

Should also add a dependency on `@types/node` - I forgot in #1

src/hooks/useKeypress.js
```diff
@@ -0,0 +1,17 @@
+import { useEffect, useContext } from 'react'
+import StdinContext from '../components/StdinContext'
+
+// TODO: memo? useRef? Will this re-init on render?
+// Should this re-initialize only if the keypressHandler changes?
```
SimenB: shouldn't be necessary for you to think about, just putt everything you pass into the closure in the dep array

index.d.ts
SimenB: Maybe not the dependency and just hope consumers add it if they need it? IIRC the error from `tsc` is ok (it specifically calls out installing `@types/node`)

The global nature of TS's types seems to cause issues https://github.com/facebook/jest/issues/8092

test/keypress.js
```diff
@@ -0,0 +1,29 @@
...
+				t.deepEqual(hits, input);
+				t.end();
+			}
+		});
+		return null;
+	};
+
+	render(<KeypressTest/>);
+	for (const letter of input.split('')) {
+		process.stdin.push(Buffer.from(letter));
```
sindresorhus: It's more correct to use `process.stdin.write()`. Then you don't need to wrap it in a Buffer either.

index.d.ts
```diff
@@ -141,11 +144,11 @@ export interface BoxProps {
...
-		| "flex-start"
-		| "center"
-		| "flex-end"
-		| "space-between"
-		| "space-around";
+	| "flex-start"
+	| "center"
+	| "flex-end"
+	| "space-between"
+	| "space-around";
```
sindresorhus: Don't do unrelated changes.

index.d.ts
```diff
@@ -44,6 +45,8 @@ export type Instance = {
 
 export type Unmount = () => void;
 
+export function useKeypress<T extends (str?: string, key?: Key) => null>(keyPresshandler: T): null;
```
sindresorhus: No need to use generics here.

sindresorhus: It returns `void`, not `null`.

jedahan:
So I got stuck on my test - it would be cool if ava could allow you to get a handle to stdin/stdout or the subprocess when running tests so we can simulate keypresses etc.

index.d.ts
jedahan: Is the current version okay regarding types? Is there a particular class/book/tutorial you recommend on learning the ins and outs of practical typescript?

vadimdemedes:
See https://yarnpkg.com/en/package/node-pty on how you can write to stdin of the process created by `node-pty` (which is used in `test/exit.js`) ;)

vadimdemedes:
Hey @jedahan, do you have time to finish this PR? Totally ok if not ;)

jedahan:
Unfortunately I cannot get `master` to npm install on my laptop now. Which is weird because a fresh checkout of node-pty 0.8.1 does npm install correctly, but thats where its erroring

vadimdemedes:
Try using Node.js v10.x.

sindresorhus:
@jedahan Did you manage to get it to npm install?

jedahan:
`npx -p node@10 npm install && npx -p node@10 npm test` passes

jedahan:
Working on fixing the test now

examples/usekeypress/usekeypress.js
```diff
@@ -0,0 +1,48 @@
...
+			setY(Math.max(1, y - 1));
+		}
+
+		if (str === 'q') {
+			exit();
+		}
+	});
+
+	// eslint-disable-next-line react/no-array-index-key
+	const ypad = new Array(y).fill().map((_, i) => <span key={i}> <br/></span>);
```
sindresorhus:
```suggestion
	const yPad = new Array(y).fill().map((_, i) => <span key={i}> <br/></span>);
```

examples/usekeypress/usekeypress.js
```diff
@@ -0,0 +1,48 @@
...
+
+	const paddedface = '^_^'.padStart(x+3);
+
+	return (
+		<div>
```
sindresorhus: This should use `<Box>`

examples/usekeypress/usekeypress.js
```diff
@@ -0,0 +1,48 @@
...
+const React = require('react');
+const {render, useKeypress} = require('../..');
+
+const {exit} = process;
+
+const Robot = () => {
+	const [x, setX] = useState(1);
+	const [y, setY] = useState(1);
+
+	useKeypress(str => {
```
sindresorhus:
```suggestion
	useKeypress(key => {
```

index.d.ts
```diff
@@ -1,4 +1,5 @@
 import * as React from "react";
+import { Key } from 'readline';
```
sindresorhus:
```suggestion
import {Key} from 'readline';
```

index.d.ts
```diff
@@ -44,6 +45,14 @@ export type Instance = {
 
 export type Unmount = () => void;
 
+/**
+ * Hook that calls keyPressHandler callback with whatever key has been pressed.
+ * See key to handle modifiers correctly.
+ */
+export function useKeypress(
+	keyPressHandler: (str?: string, key?: Key) => void
```
sindresorhus:
```suggestion
	keyPressHandler: (string?: string, key?: Key) => void
```

index.d.ts
```diff
@@ -44,6 +45,14 @@ export type Instance = {
 
 export type Unmount = () => void;
 
+/**
+ * Hook that calls keyPressHandler callback with whatever key has been pressed.
+ * See key to handle modifiers correctly.
```
sindresorhus: See what? Do you mean `Key`?

test/fixtures/handles-keypresses.js
```diff
@@ -0,0 +1,19 @@
...
+		if (input.size === 0) {
+			app.unmount();
+		}
+	});
+	return null;
+};
+
+const app = render(<KeypressTest/>);
+
+app.waitUntilExit().then(() => console.log('exited'));
```
sindresorhus:
```suggestion
(async () => {
	await app.waitUntilExit();
	console.log('Exited');
})();
```

test/fixtures/handles-keypresses.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+const React = require('react');
+const {render, useKeypress} = require('../..');
+
+const input = new Set('abcdefghijklmnopqrstuvwxyz'.split(''));
+
+const KeypressTest = () => {
+	useKeypress(str => {
```
sindresorhus:
```suggestion
	useKeypress(string => {
```

examples/usekeypress/usekeypress.js
jedahan: so useKeypress calls the callback with the string representation of the key, plus the Key object `(str, key)`. In this case, we just care about the string representation. Its might be a bit confusing and we should just choose the full Key object representation, allowing people to destructure with `({ str })` perhaps. What are your thoughts

vadimdemedes: What are the use cases for wanting the full object, not just the string representation?

SimenB: if `ctrl` is held down etc I'd guess?

vadimdemedes: It could be passed as a second argument to the callback in that case:

```js
useKeypress((key, { ctrl }) => ...)
```

SimenB: That's what happens with the current implementation

SimenB: key object is probably more useful, btw.

```js
const readline = require('readline');
readline.emitKeypressEvents(process.stdin);
process.stdin.setRawMode(true);
process.stdin.on('keypress', console.log);
```

```sh-session
$ node script.js
# hit s
s { sequence: 's',
  name: 's',
  ctrl: false,
  meta: false,
  shift: false }
# hit shift + s
S { sequence: 'S', name: 's', ctrl: false, meta: false, shift: true }
# hit ctrl + s
 { sequence: '\u0013',
  name: 's',
  ctrl: true,
  meta: false,
  shift: false }
```


So I think just a single `Key` object might be better?

vadimdemedes: I'd keep first argument as a key string for convenience to avoid destructuring when you only care about what key was pressed and keep `Key` object as a 2nd argument with `ctrl`, `shift`, etc props.

jedahan: We could experiment with alternative, simpler apis and how they feel with higher level hooks maybe like ‘useKey = cb => useKeypress((_, key) => cb(key))’

Other ways to interpret the key object into a string might be a good middle ground as well.

‘useKeystring = cb => useKeypress((_, key) => cb(keyToString(key)))’

Where keyToString converts the key object to strings like ‘meta+shift+s’ or ‘meta+S’.

jedahan:
I'm rebasing

jedahan:
I think I've made all the requested changes

vadimdemedes:
Hey @jedahan, I've tried running `node examples/usekeypress` and process exits right away, after rendering the output. It doesn't wait for user input.

vadimdemedes:
I digged deeper and was able to fix it by replacing `useEffect` with `useLayoutEffect`. The problem with `useEffect` is that it fires the handler after rendering is finished, so `setRawMode` isn't called fast enough. Process sees that there are no more tasks left and exits.

I also decided to switch from using `keypress` to `data` event, because when user pastes text into terminal, multiple `keypress` events are emitted for each character. This is not as convenient as using `data` event, which contains the whole string in one event.

I'll open another PR, which includes your changes and fixes these things. I will make sure to mention you, of course.

vadimdemedes:
Opened https://github.com/vadimdemedes/ink/pull/227 to take this PR further. I would've asked @jedahan, but I want to release it asap.

PR #148, opened 2019-03-09
Add related package - ink-color-pipe

LitoMore:
Add [ink-color-pipe](https://github.com/LitoMore/ink-color-pipe) to `Related` section.

`ink-color-pipe` is a better solution to render some mutable color text, we could put all styles to one prop.

Screenshot:


LitoMore:
@vadimdemedes ping

vadimdemedes:
Sorry for late response. Nice, let's add it!

PR #161, opened 2019-03-15
Fix waitUntilExit TypeScript definition

karaggeorge:
`waitUntilExit` is a function that returns a promise, rather than a promise itself


vadimdemedes:
Thanks!

PR #165, opened 2019-03-18
Allow passing an error on exit

### LINKED ISSUE #163: Use exit to pass exit code
Would it be possible to pass an exit code to the call to `AppContext`'s `exit` function? Then you'd get the value with `const code = await app.waitUntilExit()` and verify it's correct during testing.

I'd be willing to PR this.

mAAdhaTTah:
Passing a result to `exit` will cause `waitUntilExit` to reject
with the value as the error.

Fixes #163.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -707,6 +707,8 @@ Usage:
 </AppContext.Consumer>
 ```
 
+If `exit` is called with an Error, `waitUntilExit` will reject with the error.
+
 #### `<StdinContext>`
 
 `<StdinContext>` is a [React context](https://reactjs.org/docs/context.html#reactcreatecontext), which exposes input stream.
--- a/readme.md
+++ b/readme.md
@@ -707,7 +707,7 @@ Usage:
 </AppContext.Consumer>
 ```
 
-If `exit` is called with an Error, `waitUntilExit` will reject with the error.
+If `exit` is called with an Error, `waitUntilExit` will reject with that error.
 
 #### `<StdinContext>`
 
--- a/src/instance.js
+++ b/src/instance.js
@@ -38,8 +38,9 @@ export default class Instance {
 
 		this.container = reconciler.createContainer(this.rootNode, false, false);
 
-		this.exitPromise = new Promise(resolve => {
+		this.exitPromise = new Promise((resolve, reject) => {
 			this.resolveExitPromise = resolve;
+			this.rejectExitPromise = reject;
 		});
 	}
 
@@ -94,12 +95,16 @@ export default class Instance {
 		reconciler.updateContainer(tree, this.container);
 	}
 
-	unmount() {
+	unmount(error) {
 		this.onRender();
 		this.log.done();
 		this.ignoreRender = true;
 		reconciler.updateContainer(null, this.container);
-		this.resolveExitPromise();
+		if (error instanceof Error) {
+			this.rejectExitPromise(error);
+		} else {
+			this.resolveExitPromise();
+		}
 	}
 
 	waitUntilExit() {
--- a/index.d.ts
+++ b/index.d.ts
@@ -182,7 +182,7 @@ export const AppContext: React.Context<{
 	/**
 	 * Exit (unmount) the whole Ink app.
 	 */
-	readonly exit: () => void;
+	readonly exit: (error?: Error) => void;
 }>;
 
 /**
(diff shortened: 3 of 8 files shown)


sindresorhus:
Why can you not use `process.exitCode` as commented in https://github.com/vadimdemedes/ink/issues/163#issuecomment-473788517?

readme.md
```diff
@@ -638,6 +638,8 @@ Usage:
 </AppContext.Consumer>
 ```
 
+If `exit` is called with a value, `waitUntilExit` will reject with the value.
```
sindresorhus: This should be enforced to be an error.

src/instance.js
```diff
@@ -94,12 +95,16 @@ export default class Instance {
 		this.reconciler.updateContainer(tree, this.container);
 	}
 
-	unmount() {
+	unmount(error) {
```
sindresorhus: This needs to be documented.

sindresorhus:
Why do you need to pass the `exit` to have `waitUntilExit()` reject. I would have expected throwing an error anywhere in the Ink app would make `waitUntilExit()` reject.

sindresorhus:
You need to update https://github.com/vadimdemedes/ink/blob/master/index.d.ts

readme.md
mAAdhaTTah: Should we check that it's an Error instance in the code or just document it as such?

sindresorhus: I would check it in the code too, as it could catch hard-to-debug mistakes.

mAAdhaTTah: Updated.

src/instance.js
mAAdhaTTah: Done.

mAAdhaTTah:
> Why do you need to pass the error to exit() to have waitUntilExit() reject? I would have expected throwing an error anywhere in the Ink app would make waitUntilExit() reject.

That seems like a larger issue to solve than this. I'm not sure how to add the plumbing to handle that unless React does a bunch of that work for us.

vadimdemedes:
> Why do you need to pass the error to exit() to have waitUntilExit() reject? I would have expected throwing an error anywhere in the Ink app would make waitUntilExit() reject.

That's actually a good idea. We already have an `<App>` component which wraps the whole Ink app. We could add `componentDidCatch()` there, which would reject `waitUntilExit()`. Feel free to leave it out of this PR, we can tackle it separately.

src/instance.js
vadimdemedes: Actually I think unmounting with a custom error doesn't really make sense, it's only related to `exit()`. It's only useful for internal purposes. I think it should be removed from docs and convert this https://github.com/vadimdemedes/ink/blob/07ff24bd0a15adacad75cf04d7beb33bb733a2d2/src/render.js#L34 to `unmount: () => instance.unmount()`?

mAAdhaTTah:
> We already have an `<App>` component which wraps the whole Ink app. We could add `componentDidCatch()` there, which would reject `waitUntilExit()`.

I dig it. I'd be willing to address that in a separate PR.

mAAdhaTTah:
Updated.

sindresorhus:
@mAAdhaTTah => https://github.com/vadimdemedes/ink/pull/165/files#r267193683

mAAdhaTTah:
@sindresorhus I made that change. The `#unmount` on the App needs to take the error, but I updated the returned instance from `render` so it doesn't get passed along.

mAAdhaTTah:
Minor adjustment made: I updated the test case to use the error object for `console.log`, so we verify that the error object is actually passed through the rejected promise.

vadimdemedes:
Thanks @mAAdhaTTah, great work!

vadimdemedes:
Could you fix the merge conflict, so that we can get it to master? ;)

mAAdhaTTah:
Resolved.

vadimdemedes:
Thanks @mAAdhaTTah, I will take another look tomorrow and merge it!

mAAdhaTTah:
@vadimdemedes Not to be a pain, but a release with this fix + #170 would be really helpful for the CLI I'm working on. Thanks!

vadimdemedes:
@mAAdhaTTah Sorry for the delay, I pushed some big changes in the core and wanted to give people a few a days to test them out. Going to release in a few minutes ;)

vadimdemedes:
New release is out - https://github.com/vadimdemedes/ink/releases/tag/v2.1.0 🎆

mAAdhaTTah:
No worries, thank you!

Issue #139, opened 2019-03-07
Warnings are printed when testing an Ink component


sindresorhus:
```
Warning: Detected multiple renderers concurrently rendering the same context provider. This is currently unsupported.
Warning: Detected multiple renderers concurrently rendering the same context provider. This is currently unsupported.
Warning: Detected multiple renderers concurrently rendering the same context provider. This is currently unsupported.
Warning: Detected multiple renderers concurrently rendering the same context provider. This is currently unsupported.
Warning: Detected multiple renderers concurrently rendering the same context provider. This is currently unsupported.
Warning: Detected multiple renderers concurrently rendering the same context provider. This is currently unsupported.
```

According to @vadimdemedes:

> It’s from React actually. I don’t know how to make it forget about previous Ink reconciler instances. Because in tests you create a new Ink instance in every test, which creates a new React reconciler instance. And React apparently keeps track of context providers, but this conflict happens. Even if tests are run serially, it still complains. I tried unmounting between tests, but didn’t help. I don’t know how to clean up its cache or whatever to make React forget that this context was used before.

mAAdhaTTah:
Has there been any outreach to the React team for insight?

vadimdemedes:
There are issues already opened (and fixed) in React's repo: https://github.com/facebook/react/issues/13150 and https://github.com/facebook/react/issues/13177. Just need to figure out how to bring the fix to Ink. If you would be willing to work on that, any help is appreciated!

vadimdemedes:
I digged into React's source code and found that React only checks for multiple renderers in development mode, so executing Ink app with `NODE_ENV=production` removes that warning. That's just a workaround though, not a solution. I'm yet to figure out what it is.

vadimdemedes:
I think I found a React bug 😮React tracks if the same context is rendered more than once by assigning `_currentRenderer = {}` to it. Problem is, it doesn't clear that property when context is no longer used (on unmount, for example). So when you `render()` again and use the same context, even with a fresh reconciler, containers, etc, it still thinks that context is being used. For example (pseudo-code):

```js
const Context = React.createContext();
Context._currentRenderer; //=> undefined

const {unmount} = render(<Context.Provider>...</Context.Provider>);
Context._currentRenderer; //=> {}

unmount();
Context._currentRenderer; //=> {}, but should be undefined
```

I will open an issue on React's repo tomorrow.

vadimdemedes:
[...]

PR #170, opened 2019-03-22
Initialize reconciler once

vadimdemedes:
Fixes https://github.com/vadimdemedes/ink/issues/139, see that issue for explanation behind this PR. I would appreciate if you tried it out in your Ink-based apps to see if there are any regressions!


vadimdemedes:
Tested it on [Emma](https://github.com/maticzav/emma-cli), Ink's examples and various tests. Seems fine!

Issue #166, opened 2019-03-19
setRawMode fails when running with non-TTY stdin


eweilow:
Before the rest of the issue, I must say that Ink is amazing 👍 

An issue I've encounted is that when Ink is running in a process where `process.stdin.setRawMode` is undefined (for instance in some CI environments), these lines will throw:
https://github.com/vadimdemedes/ink/blob/master/src/components/App.js#L73
https://github.com/vadimdemedes/ink/blob/master/src/components/App.js#L83

This behaviour is detailed by the Node docs on https://nodejs.org/api/tty.html#tty_readstream_setrawmode_mode 

> When Node.js detects that it is being run with a text terminal ("TTY") attached, process.stdin will, by default, be initialized as an instance of tty.ReadStream and both process.stdout and process.stderr will, by default be instances of tty.WriteStream. The preferred method of determining whether Node.js is being run within a TTY context is to check that the value of the process.stdout.isTTY property is true

After a search for `setRawMode`, it seems that this issue has resurfaced: https://github.com/vadimdemedes/ink/issues/40 

While the easy fix would be to wrap all those calls with a check for `stdin.isTTY`, that might lead to unexpected behaviour when an app actually expects raw mode to have been set!

vadimdemedes:
> While the easy fix would be to wrap all those calls with a check for stdin.isTTY, that might lead to unexpected behaviour when an app actually expects raw mode to have been set!
Not sure there's other solution, since it's impossible to listen for user input in non-tty env, isn't it? I think Ink should probably do nothing when `setRawMode` is called in a non-tty env. What do you think?

> Before the rest of the issue, I must say that Ink is amazing 👍

Thank you!

eweilow:
> Not sure there's other solution, since it's impossible to listen for user input in non-tty env, isn't it? I think Ink should probably do nothing when `setRawMode` is called in a non-tty env. What do you think?

What might be good, since it's always nice to have a proper fallback, is to have some way for the app rendered by Ink to determine if `setRawMode` is available! I believe this might be useful for features like https://github.com/vadimdemedes/ink/pull/147

In most cases, Ink would just ignore that it can't enable raw mode, but for some cases an app might want to entirely disable an input component. So just doing nothing when `setRawMode` is called in a non-tty environment works, but there could be additional functionality added to handle it nicely as well!

mAAdhaTTah:
[...]

PR #172, opened 2019-03-22
Fix setRawMode with non-TTY stdin and expose isRawModeSupported on StdinContext

eweilow:
This pull request tries to implement the fixes discussed in https://github.com/vadimdemedes/ink/issues/166

There are not yet any tests, as I'd like to get feedback on whether this is the route to go 👍 

A thing that is concerning is that it would be nice to have useful errors for the situation in https://github.com/vadimdemedes/ink-text-input/issues/25, while an application that supports running in CI might want to just ignore setRawMode.

## Todo
- [x] Should a helpful error be thrown if setRawMode isn't supported, with a configuration option to fall back to doing nothing?
- [x] Write tests

### DIFF
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -37,7 +37,7 @@ export default class App extends PureComponent {
 					value={{
 						stdin: this.props.stdin,
 						setRawMode: this.handleSetRawMode,
-						isRawModeSupported: this.isRawModeSupported(this.props.stdin.isTTY)
+						isRawModeSupported: this.isRawModeSupported(this.props.stdin)
 					}}
 				>
 					<StdoutContext.Provider
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -36,7 +36,8 @@ export default class App extends PureComponent {
 				<StdinContext.Provider
 					value={{
 						stdin: this.props.stdin,
-						setRawMode: this.handleSetRawMode
+						setRawMode: this.handleSetRawMode,
+						isRawModeSupported: this.isRawModeSupported(this.props.stdin.isTTY)
 					}}
 				>
 					<StdoutContext.Provider
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -18,6 +18,11 @@ export default class App extends PureComponent {
 		onExit: PropTypes.func.isRequired
 	};
 
+	// Determines if TTY is supported on the provided stdin
+	static isRawModeSupported(stdin) {
+		return stdin.isTTY;
+	}
+
 	constructor() {
 		super();
 
@@ -37,7 +42,7 @@ export default class App extends PureComponent {
 					value={{
 						stdin: this.props.stdin,
 						setRawMode: this.handleSetRawMode,
-						isRawModeSupported: this.isRawModeSupported(this.props.stdin)
+						isRawModeSupported: App.isRawModeSupported(this.props.stdin)
 					}}
 				>
 					<StdoutContext.Provider
@@ -60,7 +65,7 @@ export default class App extends PureComponent {
 		cliCursor.show(this.props.stdout);
 
 		// ignore calling setRawMode on an handle stdin it cannot be called
-		if (this.isRawModeSupported(this.props.stdin)) {
+		if (App.isRawModeSupported(this.props.stdin)) {
 			this.handleSetRawMode(false);
 		}
 	}
@@ -69,15 +74,10 @@ export default class App extends PureComponent {
 		this.handleExit(error);
 	}
 
-	// Pass stdin directly here, for theoretical better reusability
-	isRawModeSupported(stdin) {
-		return stdin.isTTY;
-	}
-
 	// Handle setRawMode being called. Throws if setRawMode cannot be called on 'this.props.stdin'.
 	handleSetRawMode = isEnabled => {
 		const {stdin} = this.props;
-		if (!this.isRawModeSupported(stdin)) {
+		if (!App.isRawModeSupported(stdin)) {
 			if (stdin === process.stdin) {
 				throw new Error(
 					'setRawMode is not supported on the current process.stdin, which Ink uses as input stream by default.\nRead about how to prevent this error on https://github.com/vadimdemedes/ink/#israwmodesupported'
@@ -121,7 +121,7 @@ export default class App extends PureComponent {
 
 	handleExit = error => {
 		// ignore calling setRawMode on an handle stdin it cannot be called
-		if (this.isRawModeSupported(this.props.stdin)) {
+		if (App.isRawModeSupported(this.props.stdin)) {
 			this.handleSetRawMode(false);
 		}
 
--- a/src/components/App.js
+++ b/src/components/App.js
@@ -58,7 +58,11 @@ export default class App extends PureComponent {
 
 	componentWillUnmount() {
 		cliCursor.show(this.props.stdout);
-		this.handleSetRawMode(false);
+
+		// ignore calling setRawMode on an handle stdin it cannot be called
+		if (this.isRawModeSupported(this.props.stdin)) {
+			this.handleSetRawMode(false);
+		}
 	}
 
 	componentDidCatch(error) {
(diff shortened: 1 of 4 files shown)


vadimdemedes:
Everything looks good to me!

Considering that you've added `isRawModeSupported`, I think it does make sense to throw an error if `setRawMode` is called, but raw mode isn't supported.

eweilow:
Yeah, I agree. 

`handleSetRawMode` now throws if `isRawModeSupported(stdin)` returns false. I've added a check for if `stdin === process.stdin`, to more nicely give feedback for apps not using a custom `stdin`. Both errors provide a link to https://github.com/vadimdemedes/ink/#israwmodesupported (which will be valid if and once this PR is merged)!

The behaviour now is also that `handleSetRawMode` is ignored in `componentWillUnmount` and `handleExit` if `isRawModeSupported(stdin)` returns false. This is because nothing is run in `handleSetRawMode` if it throws, so it's not necessary to clean anything up 👍

src/components/App.js
```diff
@@ -36,7 +36,8 @@ export default class App extends PureComponent {
 				<StdinContext.Provider
 					value={{
 						stdin: this.props.stdin,
-						setRawMode: this.handleSetRawMode
+						setRawMode: this.handleSetRawMode,
+						isRawModeSupported: this.isRawModeSupported(this.props.stdin.isTTY)
```
vadimdemedes: I think there's a mistake here, it should pass `this.props.stdin`, not `this.props.stdin.isTTY`.

src/components/App.js
```diff
@@ -57,11 +58,30 @@ export default class App extends PureComponent {
...
-		this.handleSetRawMode(false);
+
+		// ignore calling setRawMode on an handle stdin it cannot be called
+		if (this.isRawModeSupported(this.props.stdin)) {
+			this.handleSetRawMode(false);
+		}
+	}
+
+	isRawModeSupported(stdin) {
+		return stdin.isTTY;
```
vadimdemedes: Is there a reason this method accepts `stdin`, instead of reading it directly from `this.props.stdin`?

eweilow: My thinking was that it's more natural to not read from this.props directly, which makes it more clear what exactly that function does. It might not even have to be an instance method on App

vadimdemedes:
Don't forget to unmark this PR as draft when you think it's ready ;)

eweilow:
The code changes are ready, but I'm not comfortable with the tests yet (I've not been able to properly run the fixtures on Windows). Going to try and wrap it up with a Mac tomorrow afternoon :)

vadimdemedes:
Hey @eweilow, how's it going? If you don't have the time, that's ok, just let me know ;)

eweilow:
I think it's ready, but the tests aren't perfect (I would have liked to have done more thorough testing of App, but I can't figure out how to do it 😢 - I'm too stuck in how Jest works). The ones I've written appear to pass on my machine, so that's something!

vadimdemedes:
It uses Ava, not Jest :) What were you struggling with in tests?

eweilow:
Ah yes, what I meant to say is that I'm too used to Jest 👍

What I also wanted to do was something like this, but it ended up troublesome because App required to be exported (which affects the public API):
```javascript
test('isRawModeSupported should return correctly', t => {
	t.true(App.prototype.isRawModeSupported({ isTTY: true }));
	t.false(App.prototype.isRawModeSupported({ isTTY: false }));
});
```
It also feels wrong to use `App.prototype` to run tests like this, which should be covered by the other tests I've written.

vadimdemedes:
You can test that by overriding `stdin` that Ink app uses like you did it here - https://github.com/vadimdemedes/ink/pull/172/files#diff-809f79e6e080c3848062e9756f30a67cR261. I would also prefer if you could address https://github.com/vadimdemedes/ink/pull/172#discussion_r269405292, that way the test you've mentioned wouldn't even be needed.

test/components.js
```diff
@@ -316,6 +317,131 @@ test('disable raw mode when all input components are unmounted', t => {
...
+	unmount();
+
+	t.is(didCatchInMount.callCount, 1);
+	t.is(didCatchInUnmount.callCount, 1);
+	t.false(stdin.setRawMode.called);
+	t.false(stdin.resume.called);
+	t.false(stdin.pause.called);
+});
+
+test('isRawModeSupported should be used if setRawMode is not supported', t => {
```
vadimdemedes: I don't think I understand the test title here, what is test about?

eweilow: It makes sure that components using the isRawModeSupported properly doesn't result in errors thrown. Thanks for the note 👍

eweilow:
Sorry for the delay. I've had to focus my attention writing on my degree project report.

Regarding https://github.com/vadimdemedes/ink/pull/172#discussion_r269405292 - it is tested by the two component tests implemented right now. 
Another, perhaps nicer way to implement the actual function, is to simply make it a static method on App. I've done this in 0f3a31ca5b59562ae9f588258e51d7528f7b4e45 if you'd want to take a look

But this essentially boils down to preference, where you'll have the final say.

vadimdemedes:
> Sorry for the delay. I've had to focus my attention writing on my degree project report.

No worries, this is open source :)

> Another, perhaps nicer way to implement the actual function, is to simply make it a static method on App. I've done this in 0f3a31c if you'd want to take a look

I don't think I understand why it has to be a `static` method. Could you revert back `isRawModeSupported` to this?

```js
isRawModeSupported() {
  return this.props.stdin.isTTY;
}
```

eweilow:
Sure, it's now reverted to that.

vadimdemedes:
Looking great, sorry for late reply! I will make some tiny adjustments (mostly naming) and merge it. Thank you!

vadimdemedes:
Great work, @eweilow, this is very useful!

vadimdemedes:
Out in `2.2.0`!

eweilow:
Awesome 👍 Now I can remove the TTY workaround I had to implement 💃

mAAdhaTTah:
This should probably have been considered a breaking change, as throwing here breaks the expected semantics of the function. You've introduced new, unexpected behavior where it worked previously. I updated `ink-testing-library`, which has issues with this change, as well as `ink-text-input`, but there are likely other components that will now throw in places where they wouldn't previously. Doing nothing would be "nicer"–not sure what the justification for throwing > silently ignoring is.

eweilow:
Calling `setRawMode` in `StdinContext.Consumer` only throws if the underlying `stdin` doesn't have `stdin.isTTY` truthy (which implies that `stdin.setRawMode` is defined). If `stdin.isTTY` is falsy, then calling `setRawMode` in the context would throw regardless since `stdin.setRawMode` would be undefined. It's much nicer to throw a useful error in this case, rather than just let a non-descriptive 'undefined is not a function' slip through.

My reasoning for not silently ignoring is that a component might want to not provide input at all if running in something like a CI environment (while still being able to use Ink). Ignoring a call to `setRawMode` might lead to a weird place where components _think_ it's enabled, while it's not.

In your PR https://github.com/vadimdemedes/ink-testing-library/pull/7 I see that you updated the  incomplete mock of https://nodejs.org/api/tty.html#tty_class_tty_readstream (which before your PR excludes the isTTY property). For a normal application (that passes a real tty.ReadStream to `stdin`), the changes won't throw 🙂

mAAdhaTTah:
Do I understand correctly: besides `ink-testing-library`, every time this function _would_ throw with our custom check, it already _does_ because `setRawMode === undefined`?

eweilow:
That's the idea. If `stream.isTTY` is true, then `stream.setRawMode` is a function (at least that's what the Node docs say). Otherwise `setRawMode === undefined` and the function on the context throws just like it did before.

The function after this PR just does slightly stricter checking for _when_ to throw and throws a nicer error.

vadimdemedes:
I agree with @eweilow, as this was technically a bug in Ink. Ink shouldn't have allowed enabling raw mode where it's unsupported, because it will crash the process.

PR #178, opened 2019-03-30
Add componentDidCatch and exit if render throws

mAAdhaTTah:
This ensures that any errors in child components can be
handled gracefully by the app with a proper Error.


vadimdemedes:
Good idea, thanks!

PR #180, opened 2019-04-01
Revert "Rewrite text squashing and text wrapping"

isaacs:
This reverts commit 1a9ba27397c92a4301c0c80e451a39be4aa7a650.

Fixes issue where `undefined` is being passed to slice-ansi, resulting in hella crashes.

There may be a more surgical way to fix the problem, but everything else that I tried (detecting undefined and passing `''` instead for example) changed the output or had some other side effects.


isaacs:
Looks like a better fix for this is coming in 2.1.1.

PR #181, opened 2019-04-03
Add ink-multi-select to useful components

karaggeorge:
Wanted to update `ink-select` to use Ink 2.0, but I can't think of anything that would be different from `ink-select-input` so I'm not sure what to do with it. Instead I made a `ink-multi-select` component, based on the `ink-select-input` which pretty much does the same as the outdated `ink-checkbox-list`.


vadimdemedes:
Nice!

PR #193, opened 2019-05-13
Improve the example Babel config in the readme

grantglidewell:
The example babel config is incomplete. This works, could save some other people time.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -95,19 +95,21 @@ The key difference you have to remember is that the rendering result isn't a DOM
 To ensure all examples work and you can begin your adventure with Ink, make sure to set up Babel with a React preset. After [installing Babel](https://babeljs.io/docs/en/usage), configure it in `package.json`:
 
 ```json
-  "babel": {
-    "presets": [
-      "@babel/preset-react",
-      [
-        "@babel/preset-env",
-        {
-          "targets": {
-            "node": "current"
-          }
-        }
-      ]
-    ]
-  }
+{
+	"babel": {
+		"presets": [
+			"@babel/preset-react",
+			[
+				"@babel/preset-env",
+				{
+					"targets": {
+						"node": true
+					}
+				}
+			]
+		]
+	}
+}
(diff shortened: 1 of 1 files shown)


sindresorhus:
How exactly is it incorrect? The point of it is to transpile JSX, not to transpile unsupported JS features in Node.js.

grantglidewell:
Sorry, incomplete would be more accurate. Using the information in the
readme, I was unable to get anything running on the command line. Node is
the target and this configurations works.

If there is something else I’m missing you might consider adding it to the
readme so other users are able to get this running. I’m just trying to be
helpful and get people who may be less familiar with Babel up and going
faster.

grantglidewell:
My initial description was unclear, apologies. to describe the issue more precisely, if I follow the readme directly it requires me to research how I want to use babel in order to get up and going with Ink. With this change to the documentation, users can immediately start using Ink without having to do that extra step.

vadimdemedes:
What syntax exactly caused an error for you?

grantglidewell:
import syntax is technically whats causing the error (the babel config in the readme doesnt handle it).

readme.md
```diff
@@ -95,13 +95,19 @@ The key difference you have to remember is that the rendering result isn't a DOM
...
-		]
-	}
-}
+  "babel": {
+    "presets": [
+      "@babel/preset-react",
+      [
+        "@babel/preset-env",
+        {
+          "targets": {
```
vadimdemedes: Is it possible to skip `targets` to make the config shorter? Also, could you use tabs here, not spaces?

grantglidewell: if you want something that works, in node, this is the most reliable way to do it.
https://babeljs.io/docs/en/babel-preset-env#targetsnode
happy to add this
https://prettier.io/docs/en/options.html#tabs

vadimdemedes: Nope, no need to add prettier, you can just convert spaces to tabs by hand.

PR #196, opened 2019-05-21
Add example with using forms

erikras:
As discussed in https://github.com/vadimdemedes/pastel/issues/9.


vadimdemedes:
Nice, thanks!

PR #197, opened 2019-05-23
Add warning to check support before calling setRawMode

eweilow:
Followup to #172 - this PR adds a warning to the readme that indicates that `setRawMode` might throw depending on the value of `isRawModeSupported`.

Felt like this might be useful 👍


vadimdemedes:
Nice, thanks!

PR #198, opened 2019-05-23
Readme: Format the properties of options clearer

eweilow:
The formatting of the subtypes of options was only indicated by a gray color - this makes it clearer, by prefixing the properties with `options.`.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -148,21 +148,21 @@ Type: `ReactElement`
 
 Type: `Object`
 
-###### stdout
+###### options.stdout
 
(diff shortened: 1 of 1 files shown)


vadimdemedes:
Thanks for the PR! However, not sure I'm a fan of this `options.` prefix, looks a bit weird to me.

sindresorhus:
> The formatting of the subtypes of options was only indicated by a gray color

It's using the heading hierarchy to indicate it. This is a very common way to do it.

Issue #143, opened 2019-03-07
Suggestion: create-ink-app


gc:
a "create-ink-app" CLI application, made by ink, seems pretty fitting - and would be very useful I think in instantly having a project setup without having to setup babel/etc. For example, this is definitely a cool project that a lot of people are having a look at, if they could instantly have an example running on their machine it would be nice.

It could also serve as an example of a fairly complex app made with ink, showing people how to structure things, etc

SimenB:
There is https://github.com/vadimdemedes/generator-ink-cli

Having one that works with `npm init` and `yarn create` would be cool, though

vadimdemedes:
@SimenB I'm actually planning to deprecate it in favor of having `create-ink-app`! Was meant to create an issue about it, but @gc beat me to it!

yestoall:
i need that to make babel work in cli

any working environment sample test?

zimme:
Name suggestion here, `create-react-cli`, it would compliment `create-react-app` quite well i think.

LitoMore:
We need a **new CLI library** or a **sub-command** for create an ink component.

For the ink component, we don't need add `ink` and `react` to the `dependencies`, put them to the `devDependencies` just be OK.

Like this: https://github.com/LitoMore/ink-color-pipe/blob/master/package.json

vadimdemedes:
It's out now! https://github.com/vadimdemedes/create-ink-app

PR #202, opened 2019-06-20
chore(docs): adds ink-starter to the README.md

pranshuchittora:
Added ink-starter to the README.md

**Why boilerplate?**
Because running ink program requires a lot of transformations to run. Like transpiring with babel,etc.

So the idea is that a user can focus on the business logic, not the infra.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -65,7 +65,7 @@ Feel free to play around with the code and fork this repl at [https://repl.it/@v
 
 
 ## Built with Ink
-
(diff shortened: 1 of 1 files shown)


sindresorhus:
You linked to the wrong repo.

I think we'll pass on this. There's just too much to comment on and it's not something I would recommend in its current state.

We plan to make something official for this: https://github.com/vadimdemedes/ink/issues/143 Until then, the easiest way to get started is to just base it on of the existing CLIs using Ink.

Or better yet, https://github.com/vadimdemedes/pastel

PR #206, opened 2019-06-25
Add `tap` to the "Built with Ink" section

isaacs:


isaacs:
Sorry, this is kind of overdue *^_^*

vadimdemedes:
Proud to have tap on this list :)

PR #207, opened 2019-06-26
Add gatsby-cli to the list of "Built with Ink"

wardpeet:
We're in love with ink so we want to be on the list! 😄 ❤️


vadimdemedes:
Oh my, that's awesome to know! ❤️

PR #208, opened 2019-06-27
[props] Fix prop types for minWidth and minHeight

visigoth:
`minWidth` and `minHeight` are used as percentages in the tests, so the
acceptable PropTypes should include a string. This quiets a warning while
running tests.


vadimdemedes:
Percentage with `minWidth` is only used in a failing test: https://github.com/vadimdemedes/ink/blob/master/test/width-height.js#L51. It's not supported yet.

Issue #158, opened 2019-03-13
Clearing old renders fails when the render goes beyond the vertical space available


joejag:
Hi, thank you very much for creating this project, I love the concept. 

I noticed a bug in my own program and noticed it's perhaps a problem with Ink itself. When the rendered view goes beyond the space available and the user scrolls back up to see the output, then older renders are still shown.

Here is an example using the Jest built-in example. I've tried this on Mac and Linux. The first run is where vertical space is not an issue and the second one shows the problem.

![example](https://user-images.githubusercontent.com/57625/54297700-458d6b00-45af-11e9-9f1e-3427411c5061.gif)

Notice old renders are still present.

Is this expected? Or is this something we could do something about?

vadimdemedes:
Hi, thanks! :)

Unfortunately, this is expected behavior with terminals. Erasing lines doesn't work beyond what's visible in terminal's viewport. There's one workaround I know of - erasing all of terminal session, including previous output from all commands (all history). Jest is doing this, but I know they end up reducing frequency of renders because clearing all of history leads to flickering. I think Ink could detect that there are lines beyond terminal's viewport, then erase terminal and reduce frequency of renders by itself. Afterwards, if terminal's height increases, this mode could be turned back off and go back to normal rendering speed and behavior.

@SimenB Could you correct me if I'm wrong on Jest's behavior here? :)

joejag:
Aha, yes, it seems this is a constraint with terminals.

Given the use case of my application, obliterating everything might be an option. Is there some code in Ink that I can use to achieve this?

Thanks.
Joe.

vadimdemedes:
Not yet, but I'm considering including it perhaps as an opt-in behavior. I will keep you posted on this issue.

vadimdemedes:
I'm going to keep it open for visibility, since other people may run into the same thing ;) Also, for myself to remember to implement that erase-all solution.

vadimdemedes:
If anyone could test #210 with their apps, that'd be great!

vadimdemedes:
Released in `2.5.0`, you can enable this feature via [experimental mode](https://github.com/vadimdemedes/ink#experimental-mode).

LeeCheneler:
> Not yet, but I'm considering including it perhaps as an opt-in behavior. I will keep you posted on this issue.

[...]

PR #210, opened 2019-06-29
Clear terminal if output is taller than viewport

vadimdemedes:
Fixes https://github.com/vadimdemedes/ink/issues/158.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,6 +41,7 @@
 	],
 	"dependencies": {
 		"@types/react": "^16.8.6",
+		"ansi-escapes": "^4.2.1",
 		"arrify": "^1.0.1",
 		"auto-bind": "^2.0.0",
 		"chalk": "^2.4.1",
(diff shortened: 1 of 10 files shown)


scotthovestadt:
@vadimdemedes This is something that would be especially useful for my case. Is there anything I can do to help push this forward? If there are open issues with the implementation, I can take a look and fix.

I checked out the CI complaints and it's just complaining about some minor lint rules, as far as I can tell.

vadimdemedes:
@scotthovestadt If you could test it out in your Ink apps, that'd be a great help!

scotthovestadt:
@vadimdemedes I (finally) applied this patch and tested it. It worked for my use-case.

LeeCheneler:
Is there any way to turn this behaviour off? Been tearing my hair out since we realised the terminal history gets wiped when the render height is taller than the console height. We're deep into development and all use tall terminals so hadn't noticed this behaviour 😱 its not documented anywhere I've looked unless I've missed it and basically means a complete rewrite of our console apps rendering layer 😭

PR #219, opened 2019-07-25
Don't error on empty string input to <Color>

benfletcher:
Even though only dev error, it's convenient to not have to guard against temporarily empty strings getting passed to <Color>.

Obviously, this is just a suggestion and one of many ways to implement. It's easy enough to write a wrapper around <Color> to handle this if that's preferred.

The specific use case is rendering of fields in color that may, at times, be empty strings.

I don't use Typescript, so sorry if this breaks something in that world.

### DIFF
--- a/src/components/Color.js
+++ b/src/components/Color.js
@@ -19,6 +19,10 @@ const methods = [
 ];
 
 const Color = ({children, ...colorProps}) => {
+	if (children === '') {
+		return null;
+	}
+	
 	const transformChildren = children => {
 		Object.keys(colorProps).forEach(method => {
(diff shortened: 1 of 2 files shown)


vadimdemedes:
Could you add a test for it? Also, CI is failing.

benfletcher:
I think everything looks good now. I added a test adjacent to another Color test. Not sure if that's the best place for it. Let me know if any other problems. Sorry for the multiple commits. Wasn't in a position to easily manage them.

vadimdemedes:
LGTM, thanks!

PR #221, opened 2019-08-24
Update scheduler version.

rakannimer:
Hello,

Ink's dependency on the `0.13.2` version of the scheduler causes an error when using some React hooks : [documented here](https://github.com/facebook/react/issues/15647).

This PR bumps the scheduler version to `0.15.0`


vadimdemedes:
Thanks!

PR #223, opened 2019-09-03
Update log-update

### LINKED ISSUE #218: Terminalwidth is off by 1 character on windows.
It seems like stdout.columns is off by 1 character. It could be an issue with yogalayout on windows. It works perfectly on mac osx.

https://github.com/vadimdemedes/ink/blob/6889f47e5c67210e780c357f0fbb51297a25f5e2/src/instance.js#L21-L23

terminal picture

Component can be found here:
https://github.com/gatsbyjs/gatsby/blob/0e78f563a48092309ba56819c09ed8d398dc8764/packages/gatsby-cli/src/reporter/reporters/ink/components/develop.js#L68-L80

This issue got reported in https://github.com/gatsbyjs/gatsby/issues/15466. I'm happy to help out on a fix for this if you can give me any pointers where to look.

herecydev:
Resolves #218

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
 		"cli-truncate": "^1.1.0",
 		"is-ci": "^2.0.0",
 		"lodash.throttle": "^4.1.1",
-		"log-update": "^3.0.0",
+		"log-update": "^3.3.0",
 		"prop-types": "^15.6.2",
 		"react-reconciler": "^0.20.0",
 		"scheduler": "^0.13.2",


vadimdemedes:
Thanks!

vadimdemedes:
Shouldn't it already install latest `log-update` with `^3.0.0` semver?

herecydev:
Well that's a good question. At least for yarn: new installs will receive the latest versions, but running a `yarn upgrade` [has no effect](https://github.com/yarnpkg/yarn/issues/4986) unless you had a direct dependency on `log-update`. The only workaround is to delete `yarn.lock` and run install again, which isn't great. Do you have any experiences with this?

PR #227, opened 2019-09-15
Add `useInput` hook to handle user input

vadimdemedes:
Follow up to https://github.com/vadimdemedes/ink/pull/147. Thanks to @jedahan for implementing the initial `useKeypress` hook. This PR changes it to listen to `data` event instead of `keypress` and renames the hook to `useInput`.


index.d.ts
```diff
@@ -44,6 +44,21 @@ export type Instance = {
...
 
+/**
+ * Hook that calls inputHandler callback with input that program received.
+ * Additionally contains helpful metadata for detecting when arrow keys were pressed.
+ */
+export function useInput(
+	inputHandler: (input: string, meta: Meta) => void
+): void;
+
+export interface Meta {
```
sindresorhus: I don't think `Meta` is clear enough as an export name. I would name it `MetaKey`.

index.d.ts
```diff
@@ -44,6 +44,21 @@ export type Instance = {
 
 export type Unmount = () => void;
 
+/**
+ * Hook that calls inputHandler callback with input that program received.
+ * Additionally contains helpful metadata for detecting when arrow keys were pressed.
+ */
+export function useInput(
+	inputHandler: (input: string, meta: Meta) => void
```
sindresorhus:
```suggestion
	inputHandler: (input: string, metaKey: MetaKey) => void
```

index.d.ts
```diff
@@ -44,6 +44,21 @@ export type Instance = {
 
 export type Unmount = () => void;
 
+/**
+ * Hook that calls inputHandler callback with input that program received.
```
sindresorhus:
```suggestion
 * Hook that calls the `inputHandler` callback with the input that the program received.
```

sindresorhus: I would clarify when the inputHandler is called. It's not clear whether it's for each keypress, each line, or something else.

index.d.ts
```diff
@@ -44,6 +44,21 @@ export type Instance = {
...
+ * Additionally contains helpful metadata for detecting when arrow keys were pressed.
+ */
```
sindresorhus: Include a usage example like in the readme. I would recommend just using all the text from the readme here. It makes it much easier to keep both in sync.

index.d.ts
sindresorhus: `meta` can be confused as `meta` keys, which is something different.

readme.md
```diff
@@ -838,6 +839,57 @@ Usage:
...
+		if (meta.left) {
+			// Left arrow key pressed
+		}
+	});
+
+	return …
+};
+```
+
+Handler function that you pass to `useInput` receives two arguments:
```
sindresorhus:
```suggestion
The handler function that you pass to `useInput` receives two arguments:
```

readme.md
```diff
@@ -838,6 +839,57 @@ Usage:
...
+
+#### input
+
+Type: `string`
+
+Input that program received.
```
sindresorhus:
```suggestion
The input that the program received.
```

readme.md
```diff
@@ -838,6 +839,57 @@ Usage:
...
+
+##### meta.left
+##### meta.right
+##### meta.up
+##### meta.down
+
+Type: `boolean`<br>
+Default: `false`
+
+If an arrow key was pressed, corresponding property will be `true`.
```
sindresorhus:
```suggestion
If an arrow key was pressed, the corresponding property will be `true`.
```

index.d.ts
sindresorhus: Or maybe just call it `key` as that's what it really is. Not worth optimizing for a potential future where you add other things there.

index.d.ts
```diff
@@ -44,6 +44,21 @@ export type Instance = {
...
+export function useInput(
+	inputHandler: (input: string, meta: Meta) => void
+): void;
+
+export interface Meta {
+	up: boolean
+	down: boolean
+	left: boolean
+	right: boolean
```
sindresorhus: Would be useful to also add `ctrl`, `meta`, `shift`. The original `keypress` event contains that.

https://thisdavej.com/making-interactive-node-js-console-apps-that-listen-for-keypress-events/

index.d.ts
```diff
@@ -44,6 +44,21 @@ export type Instance = {
...
+/**
+ * Hook that calls inputHandler callback with input that program received.
+ * Additionally contains helpful metadata for detecting when arrow keys were pressed.
+ */
+export function useInput(
+	inputHandler: (input: string, meta: Meta) => void
+): void;
+
+export interface Meta {
+	up: boolean
```
sindresorhus: `upArrow`, etc. For clarity.

examples/useinput/useinput.js
```diff
@@ -0,0 +1,41 @@
...
+		}
+
+		if (meta.down) {
+			setY(Math.min(10, y + 1));
+		}
+	});
+
+	return (
+		<Box flexDirection="column">
+			<Box>Use arrow keys to move the face. Press "q" to exit.</Box>
```
sindresorhus:
```suggestion
			<Box>Use arrow keys to move the face. Press “q” to exit.</Box>
```

src/hooks/useInput.js
```diff
@@ -0,0 +1,30 @@
...
+				left: input === '\u001B[D',
+				right: input === '\u001B[C'
+			};
+
+			inputHandler(input, meta);
+		};
+
+		stdin.on('data', handleData);
+
+		return () => stdin.off('data', handleData);
```
sindresorhus:
```suggestion
		return () => {
			stdin.off('data', handleData);
		};
```

You don't want to return the event emitter to the user, which `stdin.off` returns. That's why I only use inline arrow functions when I actually want to use the return value, not as a tiny syntactic shortcut.

src/hooks/useInput.js
```diff
@@ -0,0 +1,30 @@
+import {useLayoutEffect, useContext} from 'react';
+import {StdinContext} from '..';
+
+export default inputHandler => {
+	const {stdin, setRawMode} = useContext(StdinContext);
+
+	useLayoutEffect(() => {
+		setRawMode(true);
+
+		return () => setRawMode(false);
```
sindresorhus:
```suggestion
		return () => {
			setRawMode(false);
		};
```

vadimdemedes:
@sindresorhus Great feedback, fixed everything! Regarding https://github.com/vadimdemedes/ink/pull/227#discussion_r324512232, I was unable to reproduce `meta` detection, no idea how to trigger it. Google says it's Cmd key on Mac, but terminal seems to ignore it.

sindresorhus:
@vadimdemedes Look at how Node.js does the detection: https://github.com/nodejs/node/blob/ab841d5fbab1a9d2f8323d1ae3f71f37c6f636a1/lib/internal/readline/utils.js#L168-L172

vadimdemedes:
I meant I wasn't able to trigger the meta key, not to implement the detection of it. I tried pressing Cmd + letter, but terminal seems to ignore that combination.

vadimdemedes:
@sindresorhus Meta key is in.

index.d.ts
```diff
@@ -51,6 +51,46 @@ export type Instance = {
 
 export type Unmount = () => void;
 
+/**
+ * This hook is used for handling user input.
+ * It's a more convienient alternative to using `StdinContext` and listening to `data` events.
+ * The callback you pass to `useInput` is called for each character when user enters any input.
+ * However, if user pastes text and it's more than one character, the callback will be called only once and the whole string will be passed as `input`.
+ *
+ * ```js
```
sindresorhus: It's TS, not JS, but it's also implied, so no language tag needed.

```suggestion
 * ```
```

readme.md
```diff
@@ -846,6 +847,86 @@ Usage:
...
+Default: `false`
+
+##### key.meta
+
+Type: `boolean`<br>
+Default: `false`
+
+[Meta key](https://en.wikipedia.org/wiki/Meta_key) was pressed.
+
+Shift key was pressed.
```
sindresorhus: This is in the wrong place.

PR #231, opened 2019-09-19
Add initial type tests

sindresorhus:
I will add more type tests later on when I have more time. Just wanted to get it in so we can ask for type tests too for new features.


vadimdemedes:
Great, thanks!

Issue #183, opened 2019-04-17
Avoid calling renderNodeToOutput for each call of Component.render.


acrazing:
I'm writing a terminal game(gomoku) with ink, and I find its performance is poor when I print a `15 * 15` board with `mobx`.

I use mobx to store the focused piece, and then, when any arrow input, all the pieces will rerender, but its output is not changed. But the program is paused, after inspect the cpu profile, I find ink will call `renderNodeToOutput` after any Component updated(call render function).

![image](https://user-images.githubusercontent.com/10029810/56257583-4a3dc580-60ff-11e9-9baf-f89d5e080d87.png)

You can see the source code at <https://github.com/acrazing/gomoku-terminal>.

vadimdemedes:
There's definitely a room for progress in rendering / building layout performance. I can't promise any ETAs, but I'm planning to look into this more soon.

P.S. Nice app!

acrazing:
Perhaps you could consider caching the render result of the component and not re-rendering it to stdout if the render function does not modify the result (I'm just guessing).

vadimdemedes:
This might be possible. Currently Ink doesn't render to output if last output remained unchanged, but it still rebuilds the layout though.

Issue #228, opened 2019-09-16
`Box` `height` doesn't reset on re-render


colinking:
It appears that the `height` property might be updated when re-rendering a component, though I may be misunderstanding something!

Here's a quick example:

```jsx
// We're waiting for an async process, so we'll want a loading state.
const [results, setResults] = useState([])
useEffect(() => {
  loadResults().then(output => {
    setResults(output)
  })
}, [])

// Loading state if we're still waiting on the results
if (results.length === 0) {
  return (
    <Box height={3}>
      <Color grey>Loading...</Color>
    </Box>
  )
}

// Once the results load, show the list:
return (
  <Box flexDirection="column">
    <Color>1</Color>
    <Color>2</Color>
    <Color>3</Color>
    <Color>4</Color>
    <Color>5</Color>
  </Box>
)
```

I would expect to see all five numbers printed out, but only see the first three:

```sh
$ yarn dev
1
2
3
```

When run with `debug=true`:

> PS -- would be nice if `debug` mode added in a newline in between renders!

```
$ yarn dev --debug
Loading...

1
2
31
2
```

Note: I just swapped `height` out with `marginTop`/`marginBottom` here and it works fine!

vadimdemedes:
Could you try out https://github.com/vadimdemedes/ink/pull/232 and let me know if it fixes the issue for you?

colinking:
Put this example in [`colinking/ink-testing`](https://github.com/colinking/ink-testing), so it can be run like so:

```sh
$ git checkout https://github.com/colinking/ink-testing.git
$ npm i
$ npx ts-node src/height.tsx
```

Just tried running it with the latest version from your branch, and it appears that the CLI exits without waiting. For example:

```sh
// "ink": "^2.3.0",
$ npx ts-node src/height.tsx
1
2
3

$ 
```

```sh
// "ink": "vadimdemedes/ink#experimental-reconciler",
$ npx ts-node src/height.tsx
Loading...


$ 
```

vadimdemedes:
Yeah, I had the exact same problem few days ago. Replacing `useEffect` with `useLayoutEffect` has solved it for me. Check out https://github.com/vadimdemedes/ink/pull/147#issuecomment-531598933. TLDR is `useEffect` doesn't run soon enough after Ink renders a frame, so process exits before `setTimeout` is executed in your code. `useLayoutEffect` however, ensures to run the effect immediately after render.

vadimdemedes:
[...]

PR #232, opened 2019-09-21
Add new experimental reconciler

vadimdemedes:
This work has started after issue on Gatsby's repo was raised about performance implications of using Ink - https://github.com/gatsbyjs/gatsby/issues/15505.

Currently Ink regenerates the entire layout on each render and rebuilds the output as well. This involves recreating all Yoga nodes, which can become expensive as number and frequency of renders grows.

This PR introduces a new experimental reconciler, which only updates necessary bits of the Yoga layout tree on every render. It still rebuilds output from layout every time, but now Ink limits amount of writes into terminal output to 60 frames per second, which should be more than enough for CLIs, which should mitigate that tradeoff.

Author of the Gatsby issue above reported a significant speedup when using this reconciler and I'd appreciate if you also tried it out in your project. This is as simply as passing `experimental` option:

```js
render(<JSX/>, {experimental: true});
```

I want to test it a little more to ensure there are no visual regressions and then I'll replace current reconciler with this one.

As a side effect, this PR also fixes https://github.com/vadimdemedes/ink/issues/183.

### DIFF
--- a/src/instance.js
+++ b/src/instance.js
@@ -5,8 +5,11 @@ import logUpdate from 'log-update';
 import isCI from 'is-ci';
 import signalExit from 'signal-exit';
 import reconciler from './reconciler';
+import experimentalReconciler from './experimental/reconciler';
 import createRenderer from './renderer';
-import {createNode} from './dom';
+import createExperimentalRenderer from './experimental/renderer';
+import * as dom from './dom';
+import * as experimentalDom from './experimental/dom';
 import instances from './instances';
 import App from './components/App';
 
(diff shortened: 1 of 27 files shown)


src/instance.js
```diff
@@ -38,7 +57,11 @@ export default class Instance {
 		// so that it's rerendered every time, not just new static parts, like in non-debug mode
 		this.fullStaticOutput = '';
 
-		this.container = reconciler.createContainer(this.rootNode, false, false);
+		if (options.experimental) {
```
colinking: Could you add this to `index.d.ts`, too? Thanks! :)

vadimdemedes: Good catch!

colinking:
Just wanted to surface this on this PR, in case it's a bug with the new reconciler: https://github.com/vadimdemedes/ink/issues/228#issuecomment-533925036

PR #136, opened 2019-03-04
Add TypeScript definition

SimenB:
Closes #1 

Just a copy of https://github.com/rozzzly/template-ts-jest-webpack/blob/watch-cli/stubs/ink/index.d.ts with https://github.com/rozzzly/template-ts-jest-webpack/pull/1 and comments from the readme

PR #236, opened 2019-10-15
Remove `@types/react` from dependencies

SleeplessByte:
Introduced when adding typescript support https://github.com/vadimdemedes/ink/pull/136/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R43.

Adding `@types/react` to the `dependencies` is _great_. It allows users of the `ink` package to not worry about types. The react types are included automagically. However, having **multiple** `@types/react` in the dependency chain can cause havoc in a project, as `@types/react` tries to modify globals (specifically JSX globals). `tsc` will fail, noting that there are multiple declarations of X, Y and Z, where these are all the globals that are modified.

Having multiple `@types/react` in the chain happens for example when hoisting, when using workspaces, when using monorepos etc. Additionally, a user _should_ be able to provide their own `@types/react` version independent of this package.

A possible solution is to make this a peer dependency: it won't be provided by the `ink` library unless it's not present.

----

Note that in npm@3, it's likely that `peerDependencies` won't be automatically installed as they themselves can cause dependency hell. In this case, since typescript usage is optional, I would recommend adding this as an optional dependency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,9 +39,6 @@
 		"colors",
 		"text"
 	],
-	"peerDependencies": {
-		"@types/react": "^16.8.6"
-	},
 	"dependencies": {
 		"ansi-escapes": "^4.2.1",
 		"arrify": "^1.0.1",
--- a/package.json
+++ b/package.json
@@ -84,6 +84,7 @@
 		"xo": "^0.24.0"
 	},
 	"peerDependencies": {
+		"@types/react": ">=16.8.0",
 		"react": ">=16.8.0"
 	},
 	"babel": {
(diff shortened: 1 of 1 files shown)


sindresorhus:
It's already a peer dependency too, so I think it's just a mistake that it ended up in `dependencies`.

SimenB:
> It's already a peer dependency too, so I think it's just a mistake that it ended up in `dependencies`.

`react` is a peer dep, not `@types/react`

sindresorhus:
My mistake.

vadimdemedes:
Thanks!

PR #244, opened 2019-11-20
Add an `active` parameter to `useInput`

arcanis:
The `useInput` hook is nice, but it's always attaching itself to `stdin`. This may cause issues when too many elements are calling it, as Node will then print a warning that will glitch the output:

```
(node) warning: possible EventEmitter memory leak detected. 11 listeners added. Use emitter.setMaxListeners() to increase limit.
```

In my case it happens because each row of a table has a `useInput` registration in order to navigate from left to right - but only when the row is selected in the first place.

A fix that works for me is to simply disable the `useInput` hook for components that don't need it right now (ie rows that aren't currently selected). Unfortunately, since hooks don't work when inside control flows, it means the check must be internalised into the hook - hence this PR.

### DIFF
--- a/src/hooks/use-input.js
+++ b/src/hooks/use-input.js
@@ -1,18 +1,26 @@
 import {useEffect, useContext} from 'react';
 import {StdinContext} from '..';
 
-export default inputHandler => {
+export default (inputHandler, {active = true} = {}) => {
 	const {stdin, setRawMode} = useContext(StdinContext);
 
-	useEffect(() => {
+    useEffect(() => {
+		if (!active) {
+			return;
+		}
+
 		setRawMode(true);
 
 		return () => {
 			setRawMode(false);
 		};
-	}, [setRawMode]);
+    }, [active, setRawMode]);
 
-	useEffect(() => {
+    useEffect(() => {
+		if (!active) {
+			return;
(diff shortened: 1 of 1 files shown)


vadimdemedes:
Good catch, could you add a test for it in https://github.com/vadimdemedes/ink/blob/master/test/hooks.js?

sindresorhus:
I think it should be named `isActive`.

sindresorhus:
You also need to document it in the readme and index.d.ts.

src/hooks/use-input.js
```diff
@@ -1,18 +1,26 @@
...
+
 		setRawMode(true);
 
 		return () => {
 			setRawMode(false);
 		};
-	}, [setRawMode]);
+    }, [active, setRawMode]);
+
+    useEffect(() => {
```
sindresorhus: Use tab-indentation.

sindresorhus:
@arcanis Friendly ping :)

vadimdemedes:
Updated and merged this via https://github.com/vadimdemedes/ink/commit/f4190283c3956bd2a4cf69ba11cd678cc66976d2 to master. Thanks @arcanis!

PR #260, opened 2020-02-05
Fix error when squashing empty text nodes

### LINKED ISSUE #241: Yoga error with empty Text inside Box and Color
This issue seems new with the latest version and it crosses two libraries. 

Things that hide the error:
- Remove `<Color/>`
- Remove `<Box/>`
- Set value to a string of length 1 or greater

```
import React from 'react';

import TextInput from 'ink-text-input';
import { render, Color, Box } from 'ink';

const Test = () => {
    const [val, setVal] = React.useState('');

    return (
        <Box>
            <Color green>
                <TextInput
                    value={val}
                    onChange={setVal}
                    mask="*"
                />
            </Color>
        </Box>
    );
};

render(<Test />);
```

Error is `TypeError: Cannot read property 'yogaNode' of undefined`

Same error with `<Text />` in place of the TextInput component.

loilo:
Ink currently (since the [improvements to alignment](https://github.com/vadimdemedes/ink/compare/v2.4.0...v2.5.0?diff=unified#diff-38693372290cabd092ab6ade4a4db655) in 2.5.0) throws an error when trying to squash text nodes with zero child nodes. This PR avoids that error by bailing out early on empty text nodes.

Fixes #241

### DIFF
--- a/src/render-node-to-output.js
+++ b/src/render-node-to-output.js
@@ -27,6 +27,10 @@ const isAllTextNodes = node => {
 // Also, this is necessary for libraries like ink-link (https://github.com/sindresorhus/ink-link),
 // which need to wrap all children at once, instead of wrapping 3 text nodes separately.
 const squashTextNodes = node => {
+	if (node.childNodes.length === 0) {
+		return '';
+	}
+
 	// If parent container is `<Box>`, text nodes will be treated as separate nodes in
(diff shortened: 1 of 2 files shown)


vadimdemedes:
Thanks!

loilo:
My pleasure. 🙂

PR #262, opened 2020-02-28
Focus management

### LINKED ISSUE #242: Add `useFocus` hook to manage input focus
Do you have any recommandation for managing focus for multiple input component.
My current use case is to be able to display multiple SelectInput and be able to tab thru them. The component already have a focus prop, but how would I manage the focus when the multiple SelectInput are not within the same react component?

jdeniau:
This PR tries to start a focus management for ink.
This is a simple start but it should be extensible without breaking the API.

I really tried to write tests for this, but either I got issue with babel and "import React" errors or I did not managed to make hook works 🤔

If you can give me an hint about that ?

Fixes #242

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -79,6 +79,14 @@ export function useInput(
 	inputHandler: (input: string, key: Key) => void
 ): void;
 
+export function useFocus(): boolean;
+
+interface WithFocusProps {
+	hasFocus: boolean;
+}  
+
+export function withFocus(WrappedComponent: React.ComponentClass): React.ComponentClass<WithFocusProps>;
+
 export interface Key {
 	upArrow: boolean
 	downArrow: boolean
--- a/src/hooks/use-focus-selector.js
+++ b/src/hooks/use-focus-selector.js
@@ -1,0 +1,84 @@
+import {useContext, useEffect, useState} from 'react';
+import readline from 'readline';
+import {StdinContext} from '..';
+
+let FOCUS_ID = 0;
+function nextFocusId() {
+	FOCUS_ID++;
+
+	return FOCUS_ID;
+}
+
+export default function useFocusSelector() {
+	const [currentFocusedIndex, setCurrentFocusedIndex] = useState(0);
+	const [elementList, setElementList] = useState([]);
+
+	const register = () => {
+		const focusId = nextFocusId();
+		elementList.push(focusId);
+		setElementList(elementList.slice(0));
+
+		return focusId;
+	};
+
+	function unregister(focusId) {
+		setElementList(prev => prev.filter(element => element !== focusId));
+	}
+
+	function hasFocus(focusId) {
+		return elementList[currentFocusedIndex] === focusId;
+	}
+
+	function focusPrevious() {
+		let newIndex = currentFocusedIndex - 1;
+		if (newIndex < 0) {
+			newIndex = elementList.length - 1;
+		}
+
+		setCurrentFocusedIndex(newIndex);
+	}
+
+	function focusNext() {
+		let newIndex = currentFocusedIndex + 1;
+		if (newIndex >= elementList.length) {
+			newIndex = 0;
+		}
+
+		setCurrentFocusedIndex(newIndex);
+	}
+
+	const {stdin, isRawModeSupported, setRawMode} = useContext(StdinContext);
+
+	useEffect(() => {
+		if (isRawModeSupported) {
+			setRawMode(true);
+		}
+
+		return () => {
+			if (isRawModeSupported) {
+				setRawMode(false);
+			}
+		};
+	}, [isRawModeSupported, setRawMode]);
+
+	useEffect(() => {
+		const handleData = (ch, key) => {
+			if (key.name === 'tab') {
+				if (key.shift) {
+					focusPrevious();
+				} else {
+					focusNext();
+				}
+			}
+		};
+
+		readline.emitKeypressEvents(stdin);
+		stdin.on('keypress', handleData);
+
+		return () => {
+			stdin.off('keypress', handleData);
+		};
+	}, [stdin, focusNext, focusPrevious]);
+
+	return {register, unregister, hasFocus, elementList, currentFocusedIndex};
+}
--- a/src/hooks/use-focus-selector.js
+++ b/src/hooks/use-focus-selector.js
@@ -1,6 +1,8 @@
 import {useContext, useEffect, useState} from 'react';
-import readline from 'readline';
 import {StdinContext} from '..';
+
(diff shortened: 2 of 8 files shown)


src/hooks/use-focus-selector.js
```diff
@@ -0,0 +1,84 @@
+import {useContext, useEffect, useState} from 'react';
+import readline from 'readline';
```
jdeniau: IIRC you remove readline on a previous version.

I may remove this using `stdin.on('data')` directly

index.d.ts
```diff
@@ -79,6 +79,14 @@ export function useInput(
 	inputHandler: (input: string, key: Key) => void
 ): void;
 
+export function useFocus(): boolean;
+
+interface WithFocusProps {
+	hasFocus: boolean;
+}  
+
+export function withFocus(WrappedComponent: React.ComponentClass): React.ComponentClass<WithFocusProps>;
```
jdeniau: I'm not really sure about this line, I am not really used to typescript :man_shrugging:

taras:
@jdeniau thank you for working on this. My team and I tried to implement this over the last few days because didn't see this PR. We ended up with a very similar solution with the one exception that we're treating focus as a tree of focusable components rather than a flat list. 

Have you considered this approach?

jdeniau:
I did not.
I was inspired by the browser "tabindex" capacity.
Does a tree approach gives other interesting possibilities?

Is your code available somewhere to give a look at it?


Le ven. 6 mars 2020 à 22:29, Taras Mankovski <notifications@github.com> a
écrit :

taras:
Tree based focus management is helpful when you want to implement UI where new focus items are added to the screen after initial list of items was rendered. This is common with Detail List view. I made a diagram to show what happens when you manage focus as a tree.

Tree Focus in Detail View 001

Without a tree,

At step 4, new items will be appended to the end of the list. Which means that getting to details will require tabbing through all previously rendered items. At step 7, it's difficult to know that you have to return to item 2 without doing some explicit tracking of where the user was previously. 

We have a spike of an implementation of Focus Management but it's not a good demo yet and the code is quite rough. Feel free to poke around https://github.com/thefrontside/bigtest/blob/cli-mob/packages/ui/src/FocusManager.tsx#L106

taras:
@jdeniau just checking with you. Did you have any thoughts on this? If you like this approach, I'll be happy to pick it up after #264

jdeniau:
@taras I did not watch your code for the moment.

About the tree approach, I am torn between using a tabindex-like usage, to be consistent with the browsers, or to use a more complex but evolutive approach 🤔 

Maybe @vadimdemedes could share what he thinks about how he want to drive ink ?

taras:
> @taras I did not watch your code for the moment.

No worries. It's no hurry.

> About the tree approach, I am torn between using a tabindex-like usage, to be consistent with the browsers, or to use a more complex but evolutive approach 🤔

tabindex works fine if after the initial render, you're only adding focusable items at the end of the list. This is one of challenges of making modern SPAs accessible. If you need more granular control of focusable areas that change over the lifetime of the application then you need another data structure. Something like a tree of tab indexes which is what I described above. So, it's not so much that this approach doesn't use the same ideas, it just expands on them to accommodate more use cases.

jdeniau:
I know the limitations of what I will say, but you can also insert tabindex between two tabindex as you can see here : https://codepen.io/jdeniau/pen/bGdKmMJ?editors=1010

It is of course limited as you need to know how many items you will insert between two main items 😃

I read briefly your code by the way, but it seems quite complex at 21:00 without knowing bigtest. Il will give it another chance in the next days 😉

taras:
It's definitely doable because you can express any tree as an array. 

> I read briefly your code by the way, but it seems quite complex at 21:00 without knowing bigtest. Il will give it another chance in the next days 😉

It's more complicated than it needs to be because we're essentially building a tree in React with Context. It would be a lot simpler if we implemented this functionality at a lower level and just use the [DOM tree](https://github.com/vadimdemedes/ink/blob/master/src/dom.js) that's built into Ink.

jdeniau:
What do you think to be a good public API then ?

I was thinking of improving the current API like this :
```
useFocus({ tabindex: 1 });
withFocus(Element, { tabindex: 1 })
```

But how whould this work with a tree ? Do you intend to keep an API like this ? A possible solution is to use a semver equivalent : `1.1.8.2` or as an array `[1, 1, 8, 2]`, but it seems complex to maintain 😃

taras:
> What do you think to be a good public API then ?

We might not even need to specify an index. The render position in the DOM tree might be an acceptable default. We could start by shipping a `<Focusable>` component and/or `useFocusable` hook that returns a ref that can be passed into an element. Both of these would be sufficient without using any arguments. We can expand the API as we discover use cases that this API is not able to satisfy.

vadimdemedes:
I'm not sure we need complicated solutions to be as close as possible to browser behavior, as CLIs are supposed to be simple, clean and easy to use. 

My gut tells me we should go with the simplest solution possible at first and perhaps expose it as `unstable` prefixed feature, so people could test drive it, but also we could retain the freedom to change the API as we go and discover issues and use cases.

I'm going to dive deeper into this topic to have more context and will ping you guys back soon. We should hold off on this PR for a bit anyway, as there's currently a massive PR pending to convert codebase to TypeScript.

What do you all think?

taras:
Sounds good to me. I’m going to spike this in the CLI that we are building for BigTest. It’ll be easier to discuss specifics once there is a sufficiently complicated example to look at.

jdeniau:
@vadimdemedes looks good to me. The `unsable` tag seems a good point, as we are not sure of where to go here.

vadimdemedes:
@jdeniau Yeah, I think we should have that in Ink. By the way, have you seen https://github.com/reactjs/rfcs/issues/104? Perhaps it might be interesting for you to check out!

taras:
RFC for an RFC - explains why there is no actual RFC :)

jdeniau:
@vadimdemedes I did not. If this PR is on hold, let's can see what comes out of [the RFC](https://github.com/reactjs/rfcs/pull/109).

As a personnal level, I do not really like of React (dom version) overriding the browser, but it may seems interesting for the "large" react community.

apatrida:
> Sounds good to me. I’m going to spike this in the CLI that we are building for BigTest. It’ll be easier to discuss specifics once there is a sufficiently complicated example to look at.

@taras do you have this public and working?

vadimdemedes:
Closing in favor of https://github.com/vadimdemedes/ink/pull/295. Thanks @jdeniau for your work! I'm going to make sure to give you credit in release notes, as my implementation is based on yours.

Issue #151, opened 2019-03-11
React devtools


duncanbeevers:
Is there a way to point the standalone React devtools at an ink app?

I tried pulling in the standalone tools to connect to the devtools server running at 8097, but the standalone launcher still appears married to the browser.

```js
import { default as ReactDevtoolsCore } from 'react-devtools-core';
ReactDevtoolsCore.connectToDevTools();
```

```
ReferenceError: window is not defined
```

vadimdemedes:
I haven't looked into it, but I'm pretty sure it would be a challenging task to integrate Ink with devtools. I don't think I have enough time to work on something that big right now. Feel free to experiment with it and let me know how it goes!

vadimdemedes:
Closing as duplicate of https://github.com/vadimdemedes/ink/issues/279.

PR #264, opened 2020-03-18
Refactor codebase to TypeScript

taras:
**Note**: I created an [integration package](https://github.com/thefrontside/ink/pull/5#issuecomment-611232133) that plugin developers can use to test this new package before it's published. Please, test it by installing `@frontside/ink@next` and let me know how it goes. I'll deprecate this package after this branch is merged.
 
## Motivation

I refactored the codebase to TypeScript to make it easier to replace existing DOM with JSDOM and expose JSDOM for testing in more complex UI scenarios. This could also provide a path to #151. 

## Backstory

My team at [Frontside](https://github.com/thefrontside) and I are starting to work on the CLI for our new [BigTest test runner](https://github.com/thefrontside/bigtest). We wanted to use Ink but we wanted the CLI UI that could compete with Cypress' web UI. 

To test that kind of Ink UI, we'll want to have the ability to assert on a DOM like tree. Ink has DOM like representation but it's not a real DOM so we can't use existing DOM traversal tools to select elements to assert on. I started looking at replacing Ink's DOM with JSDOM but I found it very difficult to navigate the code because it's not typed.

So, I wanted to add types before diving further.

## Approach

I had to make quite a few changes.

1. Added typescript@3.8.3
2. Split up Ink instance into 2 separate constructors because they use different DOM structures.
3. Typed all of the files
4. Added types to [yoga-layout-prebuilt](https://github.com/vadimdemedes/yoga-layout-prebuilt/pull/7) and applied types everywhere it's used
5. Changed xo to use typescript
6. Upgraded Ava to 3.5.0 because I needed debugging functionality and all ava documentation for typescript references the new version
7. BREAKING: change Node requirements to 10 & 12
8. Configured GitHub Actions to use matrix of Node 10|12 & Experimental true|false
9. Updated tests to typescript (but didn't type them completely)
10. I used ts-node for tests but used version 7.0.0 of ts-node because it seems to have a [bug in new versions that makes it slow](https://github.com/TypeStrong/ts-node/issues/754).  
12. Removed the static index.d.ts in favour of generated types
13. Added src directory to the distribution
14. Changed props for Color component to consume types from Chalk

## Questions/TODO

- [x] Can we remove PropTypes in favour of just using TypeScript?
- [ ] Does this need to be a major release because of the Node dependency change?
- [x] Do you want me to revert to default or can we used named imports?
- [x] How should we ignore intentional unnecessary curly braces? https://github.com/vadimdemedes/ink/pull/264#discussion_r399668789


taras:
BTW, tests are passing on my fork but are not running here https://github.com/taras/ink/actions/runs/58115570

package.json
```diff
@@ -55,21 +57,39 @@
...
+		"@types/wrap-ansi": "3.0.0",
+		"@types/yoga-layout": "1.9.1",
+		"@typescript-eslint/eslint-plugin": "2.24.0",
+		"@typescript-eslint/parser": "2.24.0",
+		"ava": "^3.5.0",
 		"babel-eslint": "^10.0.1",
 		"delay": "^4.1.0",
+		"eslint-config-xo": "0.29.1",
 		"eslint-config-xo-react": "^0.22.0",
+		"eslint-config-xo-typescript": "0.26.0",
```
sindresorhus: Upgrade to the latest version `xo`. It has builtin TS support. No config needed. 👌

taras: Done!

taras:
I squashed the commits. I'll start working on replacing DOM with JSDOM in another branch as I wait for feedback on this PR. No rush.

test/flex-justify-content.tsx
```diff
@@ -18,7 +17,7 @@ test('row - align text to center', t => {
 test('row - align multiple text nodes to center', t => {
 	const output = renderToString(
 		<Box justifyContent="center" width={10}>
-			A{'B'}
+			AB
```
twick00: `AB` is not the same as `A{'B'}`. `A{'B'}` is rendered as two different text nodes and then combined. Why is this test changing?

src/components/Color.tsx
```diff
@@ -0,0 +1,126 @@
...
+	'bgRgb',
+	'bgKeyword'
+];
+
+const Color: FC<ColorProps & { children: ReactNode }> = ({children, ...colorProps}) => {
+	if (children === '') {
+		return null;
+	}
+
+	// TODO: Figure out what's going on here and write the type definition
```
twick00: Is there something better we can do here other than `@ts-ignore` everything?

test/flex-justify-content.tsx
taras: prettier got a little too helpful here. I'll revert it back.

src/components/Color.tsx
taras: I was going to return to it but forgot. Thank you for bringing it to my attention. I'll look into it in the morning.

test/flex-justify-content.tsx
taras: It was actually `xo --fix` that formatted it. I restored it back as well as another instances in the same file. https://github.com/vadimdemedes/ink/pull/264/commits/eaf6c6d8ae156def8a1351276022c64f295cfc29

src/components/Color.tsx
taras: @twick00 I went back and looked at over the types in Color component. I was able to improve types slightly by consuming types from Chalk instead of having a hardcoded interface for Color props. 

I had to use `as any` in 2 places because no matter what I tried, I couldn't get the correct call signatures to match the prop names. I have a sneaking suspicion that chalk's types are not correct. I looked at trying to fix chalk's types but it's a pretty nuanced API for such a small library. I decided to leave that yak unshaved. I rewrote `transformChildren` to use reduce because I believe it conveys the intention better than `forEach` that was there.

Does this work for you?

package.json
```diff
@@ -93,43 +112,50 @@
...
+		"require": [
+			"ts-node/register/transpile-only"
 		]
 	},
 	"xo": {
-		"parser": "babel-eslint",
 		"extends": [
-			"xo-react"
+			"xo-react",
+			"xo-typescript"
```
sindresorhus: This is not needed. It's bundled by XO.

package.json
```diff
@@ -93,43 +112,50 @@
...
+				"test/tsconfig.json"
+			]
+		},
 		"plugins": [
 			"react"
 		],
+		"extensions": [
+			"ts",
+			"tsx"
+		],
```
sindresorhus: This is also the default of XO.

package.json
```diff
@@ -93,43 +112,50 @@
...
-			"xo-react"
+			"xo-react",
+			"xo-typescript"
 		],
+		"parserOptions": {
+			"project": [
+				"tsconfig.json",
+				"test/tsconfig.json"
+			]
+		},
```
sindresorhus: This should also not be needed.

src/build-layout.ts
```diff
@@ -22,7 +29,7 @@ const buildLayout = (node, options) => {
 
 			for (const [index, childNode] of Object.entries(childNodes)) {
 				const childYogaNode = buildLayout(childNode, options).yogaNode;
-				yogaNode.insertChild(childYogaNode, index);
+				yogaNode.insertChild(childYogaNode, parseInt(index, 10));
```
sindresorhus:
```suggestion
				yogaNode.insertChild(childYogaNode, Number.parseInt(index, 10));
```

sindresorhus:
https://github.com/vadimdemedes/ink/blob/master/index.d.ts includes a lot of doc comments, can you ensure they're correct moved to the TS source?

test/flex-justify-content.tsx
sindresorhus: Can you open an issue on https://github.com/yannickcr/eslint-plugin-react if there isn't already one?

src/components/Color.tsx
```diff
@@ -15,31 +16,36 @@ const methods = [
...
-			if (colorProps[method]) {
-				if (methods.includes(method)) {
-					// @ts-ignore
-					children = chalk[method](...arrify(colorProps[method]))(children);
-					// @ts-ignore
-				} else if (typeof chalk[method] === 'function') {
-					// @ts-ignore
-					children = chalk[method](children);
+	const transformChildren = (children: string) => {
+		return Object.keys(colorProps).reduce(
```
sindresorhus: The forEach was much more readable. Generally, `reduce` is not the right solution unless you actually want to reduce numbers. Use a `for-of` loop.

.github/workflows/tests.yml
```diff
@@ -3,11 +3,12 @@ on: [push]
 
 jobs:
   test:
-    name: Node.js ${{ matrix.node_version }}
+    name: Node.js ${{ matrix.node_version }} & Experimental ${{ matrix.experimental }}
```
sindresorhus: What is the experimental matrix for?

src/components/Box.tsx
```diff
@@ -0,0 +1,123 @@
...
+	readonly flexShrink?: number;
+	readonly flexDirection?: 'row' | 'row-reverse' | 'column' | 'column-reverse';
+	readonly flexBasis?: string | number;
+	readonly alignItems?: 'flex-start' | 'center' | 'flex-end';
+	readonly justifyContent?:
+	| 'flex-start'
+	| 'center'
+	| 'flex-end'
+	| 'space-between'
+	| 'space-around';
```
sindresorhus:
```suggestion
	readonly justifyContent?:
		| 'flex-start'
		| 'center'
		| 'flex-end'
		| 'space-between'
		| 'space-around';
```

src/components/Box.tsx
```diff
@@ -0,0 +1,123 @@
...
+	readonly textWrap?:
+	| 'wrap'
+	| 'truncate'
+	| 'truncate-start'
+	| 'truncate-middle'
+	| 'truncate-end';
+	readonly unstable__transformChildren?: (x: any) => any;
```
sindresorhus: Don't use `any`. Use `unknown`.

src/components/Color.tsx
```diff
@@ -0,0 +1,85 @@
...
+	'bgHwb',
+	'bgRgb',
+	'bgKeyword',
+	'ansi',
+	'ansi256',
+	'bgAnsi',
+	'bgAnsi256'
+];
+
+const Color: FC<ColorProps & { children: ReactNode }> = ({
```
sindresorhus:
```suggestion
const Color: FC<ColorProps & {children: ReactNode}> = ({
```

src/components/Text.tsx
```diff
@@ -1,10 +1,24 @@
...
 import chalk from 'chalk';
 
-const Text = ({bold, italic, underline, strikethrough, children, unstable__transformChildren}) => {
-	const transformChildren = children => {
+interface TextProps {
+	readonly bold?: boolean;
+	readonly italic?: boolean;
+	readonly underline?: boolean;
+	readonly strikethrough?: boolean;
+	readonly unstable__transformChildren?: (x: string) => string;
```
sindresorhus: Don't use one character variable names. Use descriptive names.

src/dom.ts
```diff
@@ -0,0 +1,89 @@
+import Yoga from 'yoga-layout-prebuilt';
+import {Styles} from './styles';
+import {OutputTransformer} from './render-node-to-output';
+export type NodeNames = 'root' | 'div' | 'span';
```
sindresorhus:
```suggestion

export type NodeNames = 'root' | 'div' | 'span';
```

src/ink.tsx
```diff
@@ -0,0 +1,182 @@
...
+import {createNode, DOMNode} from './dom';
+import instances from './instances';
+import App from './components/App';
+
+export interface InkOptions {
+	stdout: NodeJS.WriteStream;
+	stdin: NodeJS.ReadStream;
+	debug: boolean;
+	exitOnCtrlC: boolean;
+	waitUntilExit?: () => Promise<any>;
```
sindresorhus: Don't use `any`. Also applies to all the other places `any` is used. Try to use a more explicit type.

src/styles.ts
```diff
@@ -0,0 +1,62 @@
...
+	minWidth?: number | string;
+	minHeight?: number | string;
+}
+
+export type Styles = PaddingStyles &
+MarginStyles &
+FlexStyles &
+DimensionStyles &
+PositionStyles &
+WrapTextStyles;
```
sindresorhus:
```suggestion
export type Styles = PaddingStyles &
	MarginStyles &
	FlexStyles &
	DimensionStyles &
	PositionStyles &
	WrapTextStyles;
```

test/components.tsx
```diff
@@ -166,8 +156,8 @@ test('squash multiple nested text nodes', t => {
 
 test('squash empty `<Text>` nodes', t => {
 	const output = renderToString(
-		<Box unstable__transformChildren={string => `[${string}]`}>
-			<Box unstable__transformChildren={string => `{${string}}`}>
+		<Box unstable__transformChildren={(str: string) => `[${str}]`}>
+			<Box unstable__transformChildren={(str: string) => `{${str}}`}>
```
sindresorhus: Don't use abbreviations for variable names.
```suggestion
			<Box unstable__transformChildren={(string: string) => `{${string}}`}>
```

tsconfig.json
```diff
@@ -0,0 +1,14 @@
+{
```
sindresorhus: Extend https://github.com/sindresorhus/tsconfig and remove the properties here that are already defined in that config.

test/render.tsx
```diff
@@ -2,17 +2,17 @@ import {serial as test} from 'ava';
...
+	let reject: (error: Error) => void;
 
 	// eslint-disable-next-line promise/param-names
 	const exitPromise = new Promise((resolve2, reject2) => {
 		resolve = resolve2;
 		reject = reject2;
 	});
 
-	const ps = spawn('node', ['./fixtures/run', `./${fixture}`, ...args], {
+	const ps = spawn('ts-node', [`./fixtures/${fixture}.tsx`, ...args], {
```
sindresorhus:
```suggestion
	const ps = spawn(process.execPath, [`./fixtures/${fixture}.tsx`, ...args], {
```

?

package.json
```diff
@@ -7,23 +7,26 @@
...
-		"node": ">=8"
+		"node": ">=10"
 	},
 	"scripts": {
-		"build": "babel src --out-dir=build",
+		"type-check": "tsc --noEmit",
+		"type-check:watch": "npm run type-check -- --watch",
+		"build": "npm run build:types && npm run build:js",
+		"build:types": "tsc --emitDeclarationOnly",
+		"build:js": "babel src --out-dir=build --extensions \".js,.ts,.tsx\" --source-maps inline",
```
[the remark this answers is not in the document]
taras: Good call, removed.

package.json
```diff
@@ -55,21 +57,38 @@
 		"signal-exit": "^3.0.2",
 		"slice-ansi": "^3.0.0",
 		"string-length": "^3.1.0",
+		"type-fest": "0.12.0",
+		"typescript": "^3.8.3",
 		"widest-line": "^3.1.0",
 		"wrap-ansi": "^6.2.0",
-		"yoga-layout-prebuilt": "^1.9.3"
+		"yoga-layout-prebuilt": "1.9.5"
```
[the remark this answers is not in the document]
taras: I think it was the result of calling `npm install <dep>`. It was not intentional. The only one that's intentional is `ts-node` because after 7.0.0 it got slower.

src/components/Color.tsx
taras: I don't agree with both of those statements but I'll use `for-of` because it's not my codebase.

taras: I'm not able to use `for-of` because of https://github.com/Microsoft/TypeScript/issues/3500, so I restored the original forEach loop.

src/components/Color.tsx
taras: I restored the original forEach loop because of https://github.com/vadimdemedes/ink/pull/264#discussion_r395977218

.github/workflows/tests.yml
taras: I'm just thought I'd run the 2 experiments in parallel instead of sequence but I'm happy to change it back if you prefer the original setup. I just realized that with this setup I'm not running `xo` in CI which is not great. Should I restore the original script that uses `npm test` or should I add `xo` as a step here?

test/components.tsx
taras: I don't usually but there are 6 other places in this file that use `str` as an argument to `unstable__transformChildren` prop. I changed it here to match. I'll remove the abbreviation in the other 6 places.

test/render.tsx
taras: This doesn't work unfortunately because `process.execPath` is path to `node` not `ts-node`. Maybe because the process is being ran by ava?

src/components/Box.tsx
taras: I just realized that this is caused by `xo --fix`. Is this a bug?

src/components/Box.tsx
taras: I can actually do better because we know that it's a `ReactNode`.

src/components/Text.tsx
taras: Changed to `children`

src/components/Box.tsx
sindresorhus: Definitely a bug in ESLint-plugin-react or our config for it. Most likely the former.

src/components/Color.tsx
sindresorhus: > I don't agree with both of those statements but I'll use `for-of` because it's not my codebase.

https://twitter.com/jaffathecake/status/1213077702300852224?s=21

.github/workflows/tests.yml
sindresorhus: I would prefer just a single job with `npm test`, but I'll leave it to @vadimdemedes to  decide.

test/render.tsx
sindresorhus: Ah, my fault. I missed that you were using `ts-node/register/transpile-only` and not `ts-node/register`.

tsconfig.json
```diff
@@ -0,0 +1,13 @@
...
+	"extends": "@sindresorhus/tsconfig",
+	"compilerOptions": {
+		"esModuleInterop": true,
+		"noImplicitAny": true,
+		"declaration": true,
```
sindresorhus: This is already defined in https://github.com/sindresorhus/tsconfig/blob/master/tsconfig.json

tsconfig.json
```diff
@@ -0,0 +1,13 @@
+{
+	"extends": "@sindresorhus/tsconfig",
+	"compilerOptions": {
+		"esModuleInterop": true,
+		"noImplicitAny": true,
```
sindresorhus: This is already implied by the `strict` option.

tsconfig.json
```diff
@@ -0,0 +1,13 @@
...
+		"declaration": true,
+		"allowSyntheticDefaultImports": true,
```
sindresorhus: From experience, this option causes lots of weird problems. Better to leave it out.

test/render.tsx
```diff
@@ -2,17 +2,17 @@ import {serial as test} from 'ava';
 import {spawn} from 'node-pty';
 import ansiEscapes from 'ansi-escapes';
 
-const term = (fixture, args = []) => {
-	let resolve;
-	let reject;
+const term = (fixture: string, args: string[] = []) => {
+	let resolve: (value?: any) => void;
```
sindresorhus: Use `unknown`.

test/helpers/run.ts
```diff
@@ -1,8 +1,11 @@
 import {spawn} from 'node-pty';
 
-export default (fixture, {env} = {}) => {
+const run: (fixture: string, props?: { env?: Record<string, unknown> }) => Promise<any> = (
```
sindresorhus: Don't use `any`.

test/helpers/render-to-string.ts
```diff
@@ -16,14 +19,20 @@ class Stream {
 	}
 }
 
-export default (node, {columns} = {}) => {
+const renderToString: (
+	node: ReactNode,
+	options?: { columns?: number }
```
sindresorhus:
```suggestion
	options?: {
		columns?: number
	}
```

test/helpers/render-to-string.ts
```diff
@@ -1,13 +1,16 @@
...
+
 	constructor({columns}) {
-		this.output = '';
 		this.columns = columns || 100;
 	}
 
-	write(str) {
+	write(str: string) {
 		this.output = str;
 	}
```
sindresorhus:
```suggestion
	write(string: string) {
		this.output = string;
	}
```

test/fixtures/exit-on-finish.tsx
```diff
@@ -1,15 +1,12 @@
-'use strict';
-const React = require('react');
-const {render, Box} = require('../..');
+import React from 'react';
+import {render, Box} from '../../src';
 
-class Test extends React.Component {
-	constructor() {
-		super();
+class Test extends React.Component<Record<string, unknown>, { counter: number }> {
```
sindresorhus:
```suggestion
class Test extends React.Component<Record<string, unknown>, {counter: number}> {
```

src/render.ts
```diff
@@ -0,0 +1,96 @@
...
+import instances from './instances';
+import {ExperimentalDOMNode} from './experimental/dom';
+import {DOMNode} from './dom';
+import {Stream} from 'stream';
+
+export interface RenderOptions {
+	stdout?: NodeJS.WriteStream;
+	stdin?: NodeJS.ReadStream;
+	debug?: boolean;
+	exitOnCtrlC?: boolean;
```
sindresorhus: Would be nice to include doc comments from all these properties. You can just copy-paste from the readme.

src/reconciler.ts
```diff
@@ -1,20 +1,28 @@
...
 } from './dom';
 
 const NO_CONTEXT = true;
 
-const hostConfig = {
+interface Props {
+	children: ReactNode;
+}
+
+const hostConfig: HostConfig<NodeNames, Props, DOMNode, any, any, any, any, any, any, any, any, any> = {
```
sindresorhus: Can we use `unknown` instead of `any` here?

src/ink.tsx
```diff
@@ -0,0 +1,184 @@
...
+		lastOutput: '',
+		rootNode,
+		render,
+		container,
+		exitPromise,
+		resolveExitPromise,
+		rejectExitPromise,
+		unmount,
+		unsubscribeExit,
+		waitUntilExit: options.waitUntilExit || waitUntilExit
```
sindresorhus:
```suggestion
		waitUntilExit: options.waitUntilExit ?? waitUntilExit
```

src/ink.tsx
```diff
@@ -0,0 +1,184 @@
...
+	// This variable is used only in debug mode to store full static output
+	// so that it's rerendered every time, not just new static parts, like in non-debug mode
+	fullStaticOutput: string;
+	render: <Props>(tree: React.ReactElement<Props>) => void;
+	renderer: InkRenderer;
+	onRender: () => void;
+	waitUntilExit: () => Promise<void>;
+	exitPromise: Promise<void>;
+	unmount: () => void;
+	resolveExitPromise: (value?: void) => void;
```
sindresorhus:
```suggestion
	resolveExitPromise: () => void;
```

?

src/ink.tsx
```diff
@@ -0,0 +1,184 @@
...
+	rejectExitPromise: (reason?: unknown) => void;
```
sindresorhus:
```suggestion
	rejectExitPromise: (reason?: Error) => void;
```

src/build-layout.ts
```diff
@@ -48,7 +55,7 @@ const buildLayout = (node, options) => {
 
 		for (const [index, childNode] of Object.entries(childNodes)) {
 			const {yogaNode: childYogaNode} = buildLayout(childNode, options);
-			yogaNode.insertChild(childYogaNode, index);
+			yogaNode.insertChild(childYogaNode, parseInt(index, 10));
```
sindresorhus:
```suggestion
			yogaNode.insertChild(childYogaNode, Number.parseInt(index, 10));
```

taras:
@sindresorhus After applying your tsconfig, there are now >80 type errors. I'm going through and fixing all of them which requires working some of the types. Please, hold off on type related feedback until I finish. I'll let you know as soon as I'm done.

src/components/Color.tsx
twick00: 👍 I'm ok with this.

tsconfig.json
twick00: ➕

taras:
Just a quick update, I fixed all of the type errors but now I have 80 failed tests :) I'll work on fixing all of them over next few days. Hopefully have it ready for review by the end of the week.

src/ink.tsx
taras: I removed any everywhere except for in Color component.

sindresorhus:
Why are you changing all the imports to be named? https://github.com/vadimdemedes/ink/pull/264/commits/dabfb2a6520f2a736bb1a5392f062c1ff1127740

taras:
> Why are you changing all the imports to be named? [dabfb2a](https://github.com/vadimdemedes/ink/commit/dabfb2a6520f2a736bb1a5392f062c1ff1127740)

It's a convention in TypeScript which eliminates the need for `esModuleInterop` and makes imports more predictable. I just pushed a commit that removed it everywhere except the imports from dependencies. I can undo it if you feel strongly about it.

src/ink.tsx
taras: Done

src/reconciler.ts
taras: Done

src/build-layout.ts
taras: I applied this everywhere.

test/render.tsx
taras: Done.

taras:
@sindresorhus I didn't fix all of the type definitions in tests because there are lots of places where arguments used to test results with args that are not allowed by typescript. They would still be allowed in JavaScript so I didn't change it. Does that work for you?

taras:
@sindresorhus please answer my questions from the description when you get a chance. Also, I'm done with type changes so you can continue with feedback.

sindresorhus:
> It's the default in TypeScript. It eliminates the need for esModuleInterop and makes imports more predictable. I just pushed a commit that removed it everywhere except the imports from dependencies. I can undo it if you feel strongly about it.

That's not entirely correct. It's just a convention in TS to use named imports. You can still use default import from CommonJS. You just have to use `import foo = require('foo')`. And using named imports for internal imports have no technical benefit other than autocomplete. But I don't feel strongly about this.

test/flex-justify-content.tsx
taras: Based on https://github.com/yannickcr/eslint-plugin-react/blob/7dce90a4f570847b33eaefddb8d9bb7d6eca9dc5/docs/rules/jsx-curly-brace-presence.md, `xo --fix`correcting `A{'B'}`to `AB`is correct behavior. 

In our case, we're intentionally causing this scenario. To prevent fix from correcting it, we should either disable globally as I've done or locally on each line/file. What would you prefer?

.github/workflows/tests.yml
taras: Restored.

src/components/Box.tsx
taras: This is formatted by `@typescript-eslint/indent`. I get `Expected indentation of 1 tab but found 2.` when I use your suggested format. I'm going to clone `typescript-eslint`package to see if I can replicate this in their tests.

taras: I believe this is the issue that we're experiencing https://github.com/typescript-eslint/typescript-eslint/issues/880

examples/counter/counter.js
```diff
@@ -7,7 +7,7 @@ const Counter = () => {
 
 	React.useEffect(() => {
 		const timer = setInterval(() => {
-			setCounter(prevCounter => prevCounter + 1);
+			setCounter(prevCounter => prevCounter + 1); // eslint-disable-line unicorn/prevent-abbreviations
```
vadimdemedes: Perhaps it's better to either rename `prevCounter` to `previousCounter` to satisfy that linter rule or disable that rule for `examples` folder.

package.json
```diff
@@ -7,23 +7,23 @@
 	"author": {
 		"name": "vdemedes",
 		"email": "vdemedes@gmail.com",
-		"url": "github.com/vadimdemedes"
+		"url": "https://github.com/vadimdemedes"
```
vadimdemedes: Shouldn't it work without `https://` prefix?

examples/counter/counter.js
sindresorhus: > previousCounter

👍

package.json
```diff
@@ -55,33 +54,42 @@
...
-		"@babel/plugin-proposal-class-properties": "^7.8.3",
-		"@babel/plugin-proposal-object-rest-spread": "^7.8.3",
-		"@babel/preset-react": "^7.8.3",
+		"@sindresorhus/tsconfig": "0.7.0",
+		"@types/is-ci": "^2.0.0",
+		"@types/lodash.throttle": "^4.1.6",
+		"@types/node": "^13.9.1",
 		"@types/react": "^16.9.19",
-		"ava": "^2.4.0",
+		"@types/react-reconciler": "^0.18.0",
```
vadimdemedes: I may be wrong, but should types for modules in `dependencies` be included there as well, instead of `devDependencies`?

package.json
sindresorhus: Unfortunately it does not, so this change is correct.

src/components/Static.tsx
```diff
@@ -0,0 +1,84 @@
+import React, {Component, ReactNode, ReactNodeArray} from 'react';
+import PropTypes from 'prop-types';
+import {Styles} from '../styles';
+
+const childrenToArray = (children: ReactNode) =>
+	Array.isArray(children) ? children : [children];
```
vadimdemedes:
```suggestion
const childrenToArray = (children: ReactNode) => {
	return Array.isArray(children) ? children : [children];
}
```

src/components/AppContext.ts
```diff
@@ -0,0 +1,12 @@
+import {createContext} from 'react';
+
+export interface AppContextValue {
```
vadimdemedes:
```suggestion
export interface AppContextProps {
```

src/components/StdinContext.ts
```diff
@@ -0,0 +1,26 @@
+import {createContext} from 'react';
+
+export interface StdinProps {
```
vadimdemedes:
```suggestion
export interface StdinContextProps {
```

src/components/StdoutContext.ts
```diff
@@ -0,0 +1,15 @@
+import {createContext} from 'react';
+
+export interface StdoutProps {
```
vadimdemedes:
```suggestion
export interface StdoutContextProps {
```

src/components/Color.tsx
```diff
@@ -0,0 +1,84 @@
...
+	'ansi',
+	'ansi256',
+	'bgAnsi',
+	'bgAnsi256'
+];
+
+/**
+ * The `<Color>` compoment is a simple wrapper around the `chalk` API. It supports all of the `chalk`'s methods as `props`.
+ */
+export const Color: FC<ColorProps & {children: ReactNode}> = ({
```
vadimdemedes: I think `React.FC` type already includes a type for `children` prop, you don't need it to include it here. It can just be `FC<ColorProps>`.

src/components/Text.tsx
```diff
@@ -1,10 +1,27 @@
...
+	readonly italic?: boolean;
+	readonly underline?: boolean;
+	readonly strikethrough?: boolean;
+	readonly unstable__transformChildren?: (children: ReactNode) => ReactNode;
+}
+
+/**
+ * This component can change the style of the text, make it bold, underline, italic or strikethrough.
+ */
+export const Text: FC<TextProps & { children: ReactNode }> = ({
```
vadimdemedes: Same as https://github.com/vadimdemedes/ink/pull/264/files#r399674079.

src/experimental/apply-style.ts
```diff
@@ -1,32 +1,66 @@
...
 
-const applyMarginStyles = (node, style) => {
-	node.setMargin(Yoga.EDGE_START, style.marginLeft || style.marginX || style.margin || 0);
-	node.setMargin(Yoga.EDGE_END, style.marginRight || style.marginX || style.margin || 0);
-	node.setMargin(Yoga.EDGE_TOP, style.marginTop || style.marginY || style.margin || 0);
-	node.setMargin(Yoga.EDGE_BOTTOM, style.marginBottom || style.marginY || style.margin || 0);
+const applyMarginStyles = (node: Yoga.YogaNode, style: MarginStyles) => {
+	node.setMargin(
+		Yoga.EDGE_START,
+		(style.marginLeft ?? style.marginX) ?? style.margin ?? 0
```
vadimdemedes: Could you keep the old syntax with `||` operator? I find `??` confusing when used multiple times at once.

src/experimental/createExperimentalInk.tsx
```diff
@@ -0,0 +1,164 @@
...
+import React, {ReactNode} from 'react';
+import signalExit from 'signal-exit';
+import {App} from '../components/App';
+import {Ink, InkOptions} from '../ink';
+import {instances} from '../instances';
+import {reconciler} from './reconciler';
+import {createRenderer} from './renderer';
+import {createNode} from './dom';
+
+export function createExperimentalInk(options: InkOptions): Ink {
```
vadimdemedes: What's the reason here for not keeping the `Instance` class, in favor of a function?

src/experimental/reconciler.ts
```diff
@@ -28,15 +52,22 @@ const hostConfig = {
 		// trigger an immediate render to ensure <Static> children are written to output before they get erased
 		if (rootNode.isStaticDirty) {
 			rootNode.isStaticDirty = false;
-			rootNode.onImmediateRender();
+			if (rootNode.onImmediateRender) {
```
vadimdemedes:
```suggestion
			if (typeof rootNode.onImmediateRender === 'function') {
```

src/experimental/reconciler.ts
```diff
@@ -28,15 +52,22 @@ const hostConfig = {
...
+				rootNode.onImmediateRender();
+			}
+
 			return;
 		}
 
-		rootNode.onRender();
+		if (rootNode.onRender) {
```
vadimdemedes:
```suggestion
		if (typeof rootNode.onRender === 'function') {
```

src/experimental/apply-style.ts
taras: This was auto corrected by xo. I would need to add an ignore or change config. Would you like me to change it?

src/ink.tsx
```diff
@@ -0,0 +1,184 @@
...
+export interface InkOptions {
+	stdout: NodeJS.WriteStream;
+	stdin: NodeJS.ReadStream;
+	debug: boolean;
+	exitOnCtrlC: boolean;
+	experimental: boolean;
+	waitUntilExit?: () => Promise<void>;
+}
+
+export interface Ink {
```
vadimdemedes: Only these props should be publicly available - https://github.com/vadimdemedes/ink#instance. I commented above on having an old `Instance` class, let's discuss there.

test/components.tsx
```diff
@@ -141,10 +133,8 @@ test('transform children', t => {
 
 test('squash multiple text nodes', t => {
 	const output = renderToString(
-		<Box unstable__transformChildren={str => `[${str}]`}>
-			<Box unstable__transformChildren={str => `{${str}}`}>
-				hello{' '}world
-			</Box>
+		<Box unstable__transformChildren={(string: string) => `[${string}]`}>
+			<Box unstable__transformChildren={(string: string) => `{${string}}`}>hello world</Box>
```
vadimdemedes: I think it was mentioned before already, but pinging just in case, it's important to keep `hello{' '}world` as-is.

test/fixtures/use-input.tsx
```diff
@@ -1,8 +1,7 @@
-'use strict';
-const React = require('react');
-const {render, useInput, AppContext} = require('../..');
+import React from 'react';
+import {render, useInput, AppContext} from '../../src';
 
-const UserInput = ({test}) => {
+const UserInput: (props: { test: string }) => JSX.Element | null = ({test}) => {
```
vadimdemedes:
```suggestion
const UserInput: React.FC<{ test: string }> = ({test}) => {
```

vadimdemedes:
@taras Fantastic work! As for JSDOM, I'd be cautious of this change, as it's not a very lightweight library and I've seen Gatsby (they use Ink for their CLI) struggle with performance already. Perhaps we should discuss how we can improve https://github.com/vadimdemedes/ink-testing-library instead?

@sindresorhus Thanks so much for reviewing this!

vadimdemedes:
Didn't see you already created an issue - https://github.com/vadimdemedes/ink-testing-library/issues/10#issue-577190746, cool, we'll discuss there.

src/experimental/apply-style.ts
vadimdemedes: Yes please, I find it more noisy than the original `||` version :)

package.json
taras: I believe it's only necessary when the time is exposed as part of the API. I don't believe that react-reconciler is.

vadimdemedes: I could be wrong, but I remember similar issues, where TS couldn't find typings for modules used internally by a dependency I installed.

taras:
@vadimdemedes do you want to keep the proptypes?

package.json
taras: Ok, I'm happy to include them just to be sure.

src/components/Color.tsx
taras: TypeScript fails to compile with errors because over defaultValue and propTypes. I can remove propTypes and defaultValue and children or keep the children. I would prefer to remove propTypes.

image

src/experimental/apply-style.ts
taras: Done in https://github.com/vadimdemedes/ink/pull/264/commits/bd1c5100cbc45cdab3326334f2629b9744deb903

src/experimental/reconciler.ts
taras: Done in https://github.com/vadimdemedes/ink/pull/264/commits/14805dd8ee591c4fb772877e2a05d7219ea844ee

src/experimental/reconciler.ts
taras: Done in https://github.com/vadimdemedes/ink/pull/264/commits/4c740b887cb1138d5615076d8c9acea0e266ce32

test/components.tsx
taras: Overly eager `xo --fix`. Stored in https://github.com/vadimdemedes/ink/pull/264/commits/92575f79e4eb69560494ecd5780c64746a19db40

test/fixtures/use-input.tsx
taras: Resolved in https://github.com/vadimdemedes/ink/pull/264/commits/afc427c554aab7b4f763fdcdac7c9e3320f78d73

src/components/Text.tsx
taras: FC includes optional children prop. According to propTypes, children is required so children needs to be defined to make it required. I moved chidren prop into TextProps in https://github.com/vadimdemedes/ink/pull/264/commits/0e678de0de66fe570cfb3830baca325af82191d8

src/components/StdoutContext.ts
taras: Done in https://github.com/vadimdemedes/ink/pull/264/commits/e75f4d21ccc379e28ab2352359343206314fed9d

src/components/AppContext.ts
taras: Done in https://github.com/vadimdemedes/ink/pull/264/commits/ce79b7560a3ce9c6aff96778caae9c037b583af5

src/components/StdinContext.ts
taras: Changed in https://github.com/vadimdemedes/ink/pull/264/commits/d09a2f3de338bc49064c467827ba17a04087b834

src/experimental/createExperimentalInk.tsx
taras: I was learning about how Ink works and refactoring to TypeScript at the same time. It was easier for me to trust a closure with internal state than a mutable object. Now that I know how Ink works, I can revert back to class/instance if you prefer. Please let me know.

vadimdemedes: Yes please, I think it makes more sense to have `Instance` class as it was before.

src/components/Color.tsx
vadimdemedes: What's the error that is being reported here? It's not visible in the screenshots.

taras: @vadimdemedes if I removed children props, I get a typescript error on propTypes - you can see it at the bottom of the first screenshot. 

Since, FC type has children as optional by default, I think the best would be to add children to ColorProps type because children are not optional.

vadimdemedes: I meant the error message from TypeScript, it's not in the screenshot.

If it works with `children` in `ColorProps` while still keeping `propTypes`, if I understood correctly, that'd be great.

src/experimental/createExperimentalInk.tsx
taras: Ok, I'll work on this later today.

taras: Ok, this is done. I called the class Ink instead of Instance. I left Instance type as result of render so the API doesn't change.

taras:
@vadimdemedes I believe everything in this branch is done based on feedback.

vadimdemedes:
Fantastic work, @taras, thank you! There are some tiny styling places to fix up, but I'll do that myself after this PR is merged. Considering the tremendous amount of change here, I'm going to take this opportunity to create a major release, including some other changes I wanted to make for a long time.

vadimdemedes:
Btw, is there a reason to include `src` folder in npm package?

vadimdemedes:
`typescript` is in `dependencies` of package.json, should it be in dev dependencies instead?

sindresorhus:
> Btw, is there a reason to include src folder in npm package?

No

> typescript is in dependencies of package.json, should it be in dev dependencies instead?

Yes

vadimdemedes:
Cool, fixed both in master.

PR #266, opened 2020-03-20
Use the same rendering approach on CI and non-CI environment by default

Fatme:
### Current behaviour
 
The components are rendered using different approaches on `CI` and `non-CI` environment.

- `CI` - Only the last frame is rendered on `CI` env. The output is rendered before exiting e.g before unmounting the component (for non-static output).
- `non-CI` - The components are rendered by erasing specified amount of lines (using `ascii` symbols) from terminal.

### The problem with current behaviour

We have `semi-integration` tests with the following setup:
* API calls are mocked
* A particular command is executed by spawning an interactive child process
* [An interactive inputs](https://github.com/vadimdemedes/ink-text-input) are shown and their values are entered by writing to the `stdin` of the spawned child process

So, we have interactive input fields (for example, with label `Enter your username`) and tests wait for that particular label (`Enter your username`) to be rendered before writing the input value.
However, the output on `CI` environment is rendered when unmounting the component. But the component is never unmounted as it waits for user input (`unmount` life cycle method is never called). All of this resulted in `timeout` issue **only on our `CI`**, locally tests are green.

After spending a lot of time investigating the issue and trying different things (changing `CI` build setup, running the tests on `self-hosted` agents in `AWS` and others) we found the problem comes from `ink` library and rendering approach on `CI` environment.

### Proposed behaviour

This PR changes the default behaviour and renders the same output on `CI` and `non-CI` environment by default. In case we want to render only last frame on `CI`, we can pass `CIMode ` option. This way, only the last frame will be rendered  on `CI` and all frames will be rendered on `non-CI` environment. Specifying `CIMode` as part of rendering options will confirm that we're expecting different output on `CI` and `non-CI`.

> NOTE:  If we agree on this approach and on the name of the new option (`CIMode`), I'll add a description in `readme`  as well.

### DIFF
--- a/src/instance.js
+++ b/src/instance.js
@@ -92,7 +92,7 @@ export default class Instance {
 			return;
 		}
 
-		if (isCI) {
+		if (this.options.CIMode && isCI) {
 			if (hasStaticOutput) {
 				this.options.stdout.write(staticOutput);
 			}
@@ -155,7 +155,7 @@ export default class Instance {
 
 		// CIs don't handle erasing ansi escapes well, so it's better to
 		// only render last frame of non-static output
-		if (isCI) {
+		if (this.options.CIMode && isCI) {
 			this.options.stdout.write(this.lastOutput + '\n');
 		} else if (!this.options.debug) {
 			this.log.done();


vadimdemedes:
Have you tested how output looks in various CIs after applying this change? In my experience, they have a lot of trouble handling ansi escapes that erase screen or parts of it.

Fatme:
Hey @vadimdemedes,

I've tested it only with [Azure Pipelines](https://docs.microsoft.com/en-us/azure/devops/pipelines/get-started/what-is-azure-pipelines?view=azure-devops).

In case if CI doesn't handle ansi escapes well, you can provide `CIMode` option and this way you'll receive only the last rendered frame.

taras:
CI environment variable is already quite abstract. If I may suggest that an alternative API is considered that would make the environment variables more specific. For example, `RENDER_LAST_FRAME` flag can be used in CI or regular environments.

vadimdemedes:
I think the best path forward here would be to check if `CI` is explicitly set to `false`. Currently this PR changes default behavior and that would be a breaking change for little to no benefit.

@Fatme would you be able to update it to work like that?

Also, I don't know how your test suite works, but I assume it spawns a process when running Ink app with input components. Another solution would be to not pass all environment variables to that process and instead pass only the ones you need. Then child process won't detect that it's running in a CI system.

@taras As for `RENDER_LAST_FRAME` environment variable, I'm afraid the amount of such custom variables could accumulate over time, which is going to lead to worse developer experience. I believe in good defaults, and since having CLIs that wait for user input is extremely uncommon, I think displaying only last frame is a better default.

taras:
@vadimdemedes setting CI=false is probably going to work fine enough for our use case. I'll make an issue if it doesn't.

Fatme:
@vadimdemedes,

I believe in good defaults as well. However, on my perspective having different behaviour on CI  **is not** a good default. It leads to confusion and people spend a lot of time debugging the unexpected behavior.

Regarding to your suggestion to pass only the needed variables - currently, we **hacked** it using this approach. The fact that we need to **hack** the environment variables shows that we have bad defaults.

In my opinion, we should have the same behavior on both `CI` and `non CI` environment. And we should be able to switch to the `"last rendered frame"` behavior only by passing an option. We shouldn't connect `"last rendered frame"` behaviour with the `CI` environment variable by default.

vadimdemedes:
> I believe in good defaults as well. However, on my perspective having different behaviour on CI is not a good default. It leads to confusion and people spend a lot of time debugging the unexpected behavior.

As I mentioned before, CIs do not handle clearing of previous output well (usage of [log-update](https://github.com/sindresorhus/log-update) module), that's why we can't make output be the same on CI and in terminal. Displaying last frame by default is at least ensuring output from Ink looks good on all CIs.

> Regarding to your suggestion to pass only the needed variables - currently, we hacked it using this approach. The fact that we need to hack the environment variables shows that we have bad defaults.

It's not common to use CLIs that require interactive input in CIs. Since you need to test such a tool in CI, it's ok to make it seem that this CLI is not running in one to show interactive input. I proposed a change to check for `CI=false` environment variable, which is going to allow such a case effortlessly.

vadimdemedes:
Closing in favor of https://github.com/vadimdemedes/ink/commit/a056565157de1049340bbf65e3ffc568a6759288.

matteodepalo:
@vadimdemedes I agree with having a sane default behavior, but could we opt out of this with a more specific Ink environment variable? We rely on `CI=true` for other purposes in our test suite so setting it to `false` is not an option for us. Could we have something like `INK_CI` or `RENDER_LAST_FRAME` to override whatever is the default behavior if set? I read your concern about not having too many environment variables, and I understand it, but in this case we might need an escape hatch for people that don't have the option of changing `CI` which unfortunately is often used for other things in CI environments and cannot be changed so easily.

vadimdemedes:
@matteodepalo Do you share the same use case as the one in PR description? If not, could you share yours?

matteodepalo:
@vadimdemedes not exactly, we mostly encountered the issue of the last frame not being consistent in CI and non-CI output. We first went with trying to isolate the test with something like `node-pty` and passing `CI: false`, but found it difficult to integrate it because of the way we run tests in parallel. In the end what we ended up doing is this https://github.com/Shopify/cli/blob/main/packages/cli-kit/src/private/node/testing/ui.ts#L101

For inputs, we're able to use the function `sendInputAndWaitForContent` defined in that file to achieve what this issue is claiming is not possible in CI.

I still think this behavior is a bit surprising and I've followed the update to React 18 process close enough to see that it was also surprising for you and caused you do debug things for a while so I'd be inclined towards giving developers the option of opting out of the surprising behavior.

matteodepalo:
```
For context, we want to make Ink's tests think they're not running in CI, so that we can replicate behavior of Ink in real-world usage.
```

I interpreted this as wanting to make Ink think that it's being run in CI so the output is what we'd get in a real-world scenario, but I might have misinterpreted this, apologies if so.

PR #268, opened 2020-03-28
Add support for `stderr`

milesj:
@vadimdemedes 

I'm working on a CLI that buffers console logs/errors/warns/etc while ink is rendering, and injects them into `Static` to be rendered above the output. This is working, however, console errors/warns are being output as `stdout` since ink does not support `stderr` currently.

This PR adds support for `stderr` and ideally the ability to render multiple `Static` components in experimental mode.

I'll add more tests once I review this PR a bit more.


vadimdemedes:
Hey @milesj, I'm curious what's the reason for writing static output to `stderr` instead?

Also, there's currently a monster PR pending to convert the entire codebase to TypeScript, so unfortunately I can't merge this PR just yet to avoid causing massive merge conflicts, sorry!

milesj:
I can provide some examples. So basically I wrap all `console.*` methods (https://github.com/milesj/boost/blob/cli/packages/cli/src/LogBuffer.ts) and have a custom logger (https://github.com/milesj/boost/blob/cli/packages/cli/src/Program.tsx#L62) so that console output does not disrupt the `log-update` render loop of ink.

I then redirect this buffered output to `Static` here: https://github.com/milesj/boost/blob/cli/packages/cli/src/Wrapper.tsx#L65 This works, but now `console.error` (and others) are no longer `stderr`, which this PR aims to fix.

This isn't super critical to land right now, but would be nice to have in the future. Totally down to wait for the TS stuff to land first.

src/experimental/renderer.js
```diff
@@ -5,20 +5,24 @@ import Output from './output';
...
 
 	for (const childNode of node.childNodes) {
 		if (childNode.unstable__static) {
-			return childNode;
+			nodes.push(childNode);
 		}
 
 		if (Array.isArray(childNode.childNodes) && childNode.childNodes.length > 0) {
-			return findStaticNode(childNode);
+			nodes.push(...findStaticNodes(childNode));
```
vadimdemedes, line 21: I'm a bit concerned about the impact of this change in terms of performance. Removing the restriction on having just one `<Static>` component means that now we have to traverse entire component tree to find all `<Static>` components.

I think there are few things we can do here:

1. Allow `<Static>` component to only be placed at top-level of your app (can't be nested deep in the tree)
2. Remember references to `<Static>` components from the first time we find them and avoid calling `findStaticNodes` on every render
3. Add `useStatic()` hook which, when called, tells Ink that there's a new `<Static>` component in the tree and sends a reference to it.

However, all of these don't sound like long-term solutions, there's something hacky about each of them. Now that I think of it, we can't actually allow multiple `<Static>` components through CLI app, because there's no way for developer to know in which order they will be written to stdout/stderr. That basically means we should allow only 2 `<Static>` components, one for stdout and one for stderr. Perhaps we could have either `<Static stream="stdout|stderr">` or `<Stdout>` + `<Stderr>` components. 

Sorry for the brain dump! I'll experiment with that idea and let you know ;)

milesj: Yeah that makes sense. I have another idea. What if static took either a string, or another component that targets which stream?

```
<Static>
  <Log type="stderr">Foo</Log>
</Static>
```

vadimdemedes:
Hey @milesj, I pushed an alternative solution to manually push output to stdout/stderr, perfectly suited for logs - https://github.com/vadimdemedes/ink#usestderr. It's not released yet, but it will be as part of Ink 3. Thanks for your work, wouldn't have happened otherwise!

milesj:
Awesome, glad to hear!

PR #276, opened 2020-04-25
Measure text using Yoga during layout calculation

vadimdemedes:
Before this change, Ink had to do 3 steps to measure and wrap text correctly.
First step would calculate the layout, second step would traverse the whole tree
and find text nodes that need wrapping and third step would recalculate layout
once more with "real" text dimensions assigned during the second step.

This change does all of this in one go using Yoga's `setMeasureFunc()` method.
This method assigns a function to calculate the dimensions of each node.
Ink can assign this function to all text nodes to efficiently calculate their
dimensions during layout and let Yoga reposition nodes accordingly.

Blocked by https://github.com/DefinitelyTyped/DefinitelyTyped/pull/44222. Once this PR is merged, I'll remove `// @ts-ignore`.


vadimdemedes:
I'm actually going to merge this now, because a PR to DefinitelyTyped can take a long time to get it.

vadimdemedes:
Opened https://github.com/vadimdemedes/yoga-layout-prebuilt/issues/9, so that I don't forget to get back to this.

PR #281, opened 2020-05-01
Implement a new <Static> component

vadimdemedes:
Previous implementation of `<Static>` component was performing poorly for apps
with thousands of items to render. Ink's rendering process slowed down
proportionally to the number of children in `<Static>` component, because
React has to call `React.createElement()` for each child, even if it was already
rendered. See https://github.com/gatsbyjs/gatsby/pull/21955#discussion_r387703540
for more context.

This change changes the API of a `<Static>` component to call `React.createElement()`
and render only new items, ignoring the items that were already rendered.
The new API is looking very similar to virtual list libraries.
You pass a list of items to render via `items` prop and specify a function
that is responsible for rendering each item, the rest is handled for you.

```jsx
<Static items={['a', 'b', 'c']}>
	{item => <Box key={item}>Item: {item}</Box>}
</Static>
```

This change also removes the need to traverse entire node tree on every render
to find where `<Static>` is located and extract its contents. Reconciler
is going to save a reference to `<Static>` node as soon as it encounters one,
saving precious CPU cycles and thus making rendering less expensive.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -644,25 +644,21 @@ Usage:
 
 #### `<Static>`
 
-`<Static>` component allows permanently rendering output to stdout and preserving it across renders.
-Components passed to `<Static>` as children will be written to stdout only once and will never be rerendered.
-`<Static>` output comes first, before any other output from your components, no matter where it is in the tree.
-In order for this mechanism to work properly, at most one `<Static>` component must be present in your node tree and components that were rendered must never update their output. Ink will detect new children appended to `<Static>` and render them to stdout.
+`<Static>` component permanently renders its output above everything else.
+It's useful for displaying activity like completed tasks or logs - things that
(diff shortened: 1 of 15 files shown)


readme.md
```diff
@@ -671,7 +667,52 @@ Jest continuously writes the list of completed tests to the output, while updati
...
+**Note:** `<Static>` only renders new items in `items` prop and ignores items
+that were previously rendered. This means that when you add new items to `items`
+array, changes you make to previous items will not trigger a rerender.
+
+See [examples/jest](examples/jest/jest.js) for a basic implementation of Jest's
+UI using `<Static>` component.
+
+##### items
+
+Type: `array`
```
sindresorhus, line 679:
```suggestion
Type: `Array`
```

To adhere to TS types.

PR #287, opened 2020-05-07
Add support for React Devtools

vadimdemedes:
This change adds support for debugging Ink apps using React Devtools. All you need to do is run your CLI with `DEV=true` environment variable set.

```
$ DEV=true my-cli
```

Then, start up React Devtools itself to debug:

```
$ npx react-devtools
```

After it starts up, you should see the component tree of your CLI.
You can even inspect and change the props of components, and see the results immediatelly in the CLI, without restarting it.


SimenB:
This is super cool! 👍 

Thoughts on a less generic name than `DEV`? E.g. `INK_REACT_DEVTOOLS` or something?

SimenB:
Fair enough. I'm still a bit worried something else will pick it up, what about `INK_DEV`? I might also be worried about something's that not an issue 😀

vadimdemedes:
Doubt it should be a problem, since `DEV` is not such a common env variable. If it will become one, it will be simple to swap with something else.

PR #295, opened 2020-06-07
Add focus management

vadimdemedes:
This PR adds a simple focus management mechanism to Ink. It's based on @jdeniau's work in https://github.com/vadimdemedes/ink/pull/262 and [Focus Management RFC](https://github.com/reactjs/rfcs/issues/104) in React.

You can check out the documentation for changes made here in https://github.com/vadimdemedes/ink/tree/focus#usefocusoptions, but the gist is that components can call `useFocus()` hook to notify Ink that they can be focused and when user presses <kbd>Tab</kbd>, Ink is going to move focus between focusable components.

```js
import {useFocus} from 'ink';

const FocusableComponent = () => {
  const {isFocused} = useFocus();
  // `isFocused` will equal `true` after user presses Tab and this component receives focus

  return <Text>{isFocused ? 'I am focused' : 'I am not focused'}</Text>;
};
```

### DIFF
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/prefer-nullish-coalescing */
 import React, {PureComponent} from 'react';
 import type {ReactNode} from 'react';
 import PropTypes from 'prop-types';
@@ -6,6 +7,11 @@ import AppContext from './AppContext';
 import StdinContext from './StdinContext';
 import StdoutContext from './StdoutContext';
 import StderrContext from './StderrContext';
+import FocusContext from './FocusContext';
+
+const TAB = '\t';
+const SHIFT_TAB = '\u001B[Z';
+const ESC = '\u001B';
 
 interface Props {
 	children: ReactNode;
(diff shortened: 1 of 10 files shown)


vadimdemedes:
@jdeniau @taras Let me know what you think!

src/components/App.tsx
```diff
@@ -69,7 +92,21 @@ export default class App extends PureComponent<Props> {
...
+									add: this.addFocusable,
+									remove: this.removeFocusable,
+									activate: this.activateFocusable,
+									deactivate: this.deactivateFocusable,
+									enableFocus: this.enableFocus,
+									disableFocus: this.disableFocus,
+									focusNext: this.focusNext,
+									focusPrevious: this.focusPrevious
+								}}
+							>
```
jdeniau, line 107: Personally, I'd move this context provider and all the following methods into a separate file because this file becomes more complicated with those 200 supplementary lines. (ie. someone that want to improve ink might be lost in those lines whereas he does not need to understand focus management)

vadimdemedes: Agree and this is what I wanted to do initially, but decided to leave it in `App.tsx` for now to reuse the `data` event handler on `stdin` stream, instead of having two separate ones. I think a better move here is to rewrite this file using hooks and extract individual functionality (focus management, exiting, raw mode) into hooks, which will clean this whole thing up.

PR #299, opened 2020-06-13
Rewrite text rendering

vadimdemedes:
This is going to be the biggest change in Ink 3 release so far. TLDR is that all text must be wrapped in `<Text>` or `<Color>` components from now on. I know it may not be a popular move, but I assure you - it's necessary.

```jsx
// Before
<Box>Hello World</Box>

// After
<Text>Hello World</Text>
```

## Problem

I started adding support for borders to `<Box>` component and when I started testing text wrapping inside those boxes, I realized that there's a constant mismatch between Yoga node measurements and how Ink renders text to output. I started digging and realized that complicated logic I've had for text squashing and wrapping was at fault here.

Text squashing is basically a [`Node.normalize()`](https://developer.mozilla.org/en-US/docs/Web/API/Node/normalize) behavior implemented in Ink. If there's a component like this:

```jsx
<Box>Failed tests: {5}</Box>
```

Ink would join two text nodes here ("Failed tests: " and "5") into one *after* Yoga finished its layout, wrap text if necessary and then render this text to the terminal.

Problems start with the fact that Ink previously created a new Yoga node for each text node:

Since Yoga doesn't support any text handling and all nodes are essentially `<div style="display: flex">`, correct text wrapping becomes impossible when each text node is a box under the hood. Consider a case when there's a `<Box>` with limited width and two long text nodes inside of it.

```jsx
<Box width={10}>
  {'Helloooooooo'}{' World'}
</Box>
```

This is how Yoga would build such a layout:

Untitled-2020-06-13-1600

You would expect to see "World" node on the same line as the last "ooo" characters, but those are two boxes, and second box can't just start where first box finished.

Code for various workarounds in handling of text layout in Ink is complicated and it's hard to predict the consequences of changes I make. This is not maintainable long-term. That's why I started looking into alternative ways to solve text squashing and wrapping once and for all.

## Solution

Fortunately, I didn't need to look for a better way long enough. I've recently used React Native and one thing that immediatelly caught my attention was `<Text>` components. All text in React Native must be wrapped in `<Text>` component otherwise app is going to fail.

Ink and React Native have a common limitation where we don't have a full-featured browser rendering engine at our disposal, hence why both Ink and React Native use Yoga for layout. That's why I figured React Native just must have a solution for this problem. After digging into a custom React reconciler that React Native uses and React Native codebase itself, I found how reconciler works with text nodes and how React Native works with Yoga to work with text.

First, their reconciler always creates text nodes instead of checking if text can be assigned as `textContent` to parent node. For example:

```js
const div = document.createElement('div');

// Assign text directly to container without explicitly creating a text node
div.textContent = 'Hello World';
```

In Ink, figuring out where to look for text during rendering and accounting for all possible combinations is a pain and a source of rendering bugs. Forcing text to always be contained in text nodes helped remove this burden.

Second, the fact that text is always wrapped into `<Text>` lets React Native apply special logic to the contents of these text nodes. No matter how many nested nodes a `<Text>` component might have, there's only one Yoga node behind all of it. This is brilliant, because as contents of `<Text>` change, we can just remeasure the dimensions and update the Yoga node to correctly reflect the changes.

Ink adopted both of these techniques, which greatly simplified the codebase and improved my confidence in how the new layout mechanism works. Now if we look at our last example with long "Helloooooooo World" text, this is how it works now:

Ink creates only one "real" text node for the `<Text>` component and all its child nodes become "virtual" text nodes that don't have individual Yoga nodes, and instead they join together. Now text is wrapped similarly to how a browser would wrap it, where inline nodes are supported. This results in a system, where text measurements in Yoga and Ink always match, predictable and simple to debug.

Rendering process has also been simplified (see https://github.com/vadimdemedes/ink/blob/b2cab926fbbfe801664ebd028ec175d4527ea7fb/src/render-node-to-output.ts), which is a step towards a future where Ink doesn't need to rerender the entire layout on each update, but only rerender the changed parts.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -234,7 +234,7 @@ Ink uses [Yoga](https://github.com/facebook/yoga) - a Flexbox layout engine to b
 It's important to remember that each element is a Flexbox container.
 Think of it as if each `<div>` in the browser had `display: flex`.
 See `<Box>` built-in component below for documentation on how to use Flexbox layouts in Ink.
-Note that all text must be wrapped in `<Text>` component.
+Note that all text must be wrapped in a `<Text>` component.
 
 ### Built-in Components
 
--- a/readme.md
+++ b/readme.md
@@ -257,28 +257,28 @@ const Example = () => (
 
 ##### bold
 
-Type: `boolean`<br>
+Type: `boolean`\
 Default: `false`
 
 ##### italic
 
-Type: `boolean`<br>
+Type: `boolean`\
 Default: `false`
 
 ##### underline
 
-Type: `boolean`<br>
+Type: `boolean`\
 Default: `false`
 
 ##### strikethrough
 
-Type: `boolean`<br>
+Type: `boolean`\
 Default: `false`
 
 ###### wrap
 
-Type: `string`<br>
-Values: `wrap` `truncate` `truncate-start` `truncate-middle` `truncate-end`<br>
+Type: `string`\
+Values: `wrap` `truncate` `truncate-start` `truncate-middle` `truncate-end`\
 Default: `wrap`
 
 This property tells Ink to wrap or truncate text if its width is larger than container.
--- a/readme.md
+++ b/readme.md
@@ -112,9 +112,9 @@ Don't forget to import `React` into every file that contains JSX:
 
 ```jsx
 import React from 'react';
-import {render, Box} from 'ink';
+import {render, Text} from 'ink';
 
-const Demo = () => <Box>Hello World</Box>;
+const Demo = () => <Text>Hello World</Text>;
 
 render(<Demo />);
 ```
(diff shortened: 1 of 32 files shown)


vadimdemedes:
After this PR is in, I'm going to work on removing `<Color>` component in favor of a universal `color` prop in `<Text>` component instead. I think this is a logical move, since color is only ever applied to text, because there's nothing else you can render in Ink.

It would look like:

```jsx
<Text color="green">
<Text color="rgb(255, 255, 255)">
<Text color="#fff">
<Text backgroundColor="white">
<Text backgroundColor="rgb(0, 0, 0)">
<Text backgroundColor="#000">
```

When I'm going to add support for borders to `<Box>`, color could be applied to border like this:

```jsx
<Box borderColor="green">
  <Text color="green">5 tests passed</Text>
</Box>
```

readme.md
```diff
@@ -234,9 +234,80 @@ Ink uses [Yoga](https://github.com/facebook/yoga) - a Flexbox layout engine to b
 It's important to remember that each element is a Flexbox container.
 Think of it as if each `<div>` in the browser had `display: flex`.
 See `<Box>` built-in component below for documentation on how to use Flexbox layouts in Ink.
+Note that all text must be wrapped in `<Text>` component.
```
sindresorhus, line 237:
```suggestion
Note that all text must be wrapped in a `<Text>` component.
```

readme.md
```diff
@@ -234,9 +234,80 @@ Ink uses [Yoga](https://github.com/facebook/yoga) - a Flexbox layout engine to b
...
+		<Text italic>I am italic</Text>
+		<Text underline>I am underline</Text>
+		<Text strikethrough>I am strikethrough</Text>
+	</>
+);
+```
+
+##### bold
+
+Type: `boolean`<br>
```
sindresorhus, line 260: Unrelated to this PR, but just FYI, you can now use `\` instead of `<br>` to force a linebreak.
```suggestion
Type: `boolean`\
```

PR #302, opened 2020-06-15
Extract console patching and make it configurable

vadimdemedes:
This PR extracts console patching code into [patch-console](https://github.com/vadimdemedes/patch-console) module and exposes a `patchConsole` option to disable it (it's enabled by default). It's useful when users want to implement custom loggers and don't want Ink's handling of it.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -167,6 +167,13 @@ Default: `true`
 Configure whether Ink should listen to Ctrl+C keyboard input and exit the app.
 This is needed in case `process.stdin` is in [raw mode](https://nodejs.org/api/tty.html#tty_readstream_setrawmode_mode), because then Ctrl+C is ignored by default and process is expected to handle it manually.
 
+###### patchConsole
+
+Type: `boolean`\
+Default: `true`
+
+Patch console methods to ensure console output doesn't mix with Ink output.
+
 ###### debug
 
 Type: `boolean`<br>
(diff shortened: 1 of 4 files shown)


readme.md
```diff
@@ -167,6 +167,13 @@ Default: `true`
 Configure whether Ink should listen to Ctrl+C keyboard input and exit the app.
 This is needed in case `process.stdin` is in [raw mode](https://nodejs.org/api/tty.html#tty_readstream_setrawmode_mode), because then Ctrl+C is ignored by default and process is expected to handle it manually.
 
+###### patchConsole
+
+Type: `boolean`\
+Default: `true`
+
+Patch console methods to ensure console output doesn't mix with Ink output.
```
mAAdhaTTah, line 175: More explanation as to where the console output actually goes would be helpful. I'm not 100% clear on the implementation, but it kinda looks like the call to `writeToStdout` clears the console, then writes the data. Would a `console.log` call in a `patchConsole` enabled app result in Ink's output being clear, outputting the log message, then Ink overwriting said log message on the next render?

mAAdhaTTah:
I like this approach and this works for me since if I want to handle log messages myself. Maybe pointing to `patch-console` in the docs so if personal handling is desired, we know to use that?

readme.md
vadimdemedes: Good point!

vadimdemedes: You are right that `writeToStdout` clears console, then writes the data. After that happened, Ink output is rerendered again at the end. This ensures both outputs are visible and are not mixed with each other.

PR #305, opened 2020-06-19
Don't override existing global.window or global.WebSocket

taowen:
I found import ink will override the global.window provided by my jsdom setup.  It would be nice to check first before set it globally.


vadimdemedes:
Thanks!

Issue #168, opened 2019-03-20
Get width of <Box>


maticzav:
Is it possible to obtain the width of the `Box` element in Ink 2.0?

vadimdemedes:
There's a way, but I'm still thinking how to properly expose it in a way that can live on long term. Hopefully I'm going to finish it soon and get back to you in this thread.

vadimdemedes:
I forgot to ask, what's the use case for this? When would you need to know a specific width of `<Box>`?

maticzav:
I want to create a general table component with `<tr>` and `<td>` - similar to how tables work in plain HTML where everything is aligned by default.

vadimdemedes:
Hm, would you be able to use flexbox properties like `flexGrow` to achieve the same layout?

jdeniau:
A possible use case is to "hack" something to have border as if I'm right, there is no way to have a bordered box for now.

danikaze:
Anyways, if anyone is creating layouts inside `Box` with fancy component that depends on the parent's dimensions, I guess both `width` and `height` should be exposed...
i.e.
```ts
<Box height={10} width={20}>
  <FancyComponent that wants to use 100% of available space (parent) />
</Box>
```

jdeniau:
I think the flexbox comportement of Box should work in that case

danikaze:
> I think the flexbox comportement of Box should work in that case

@jdeniau  I'm sorry I'm a bit new with this library but... how would it be done if the `<FancyComponent>` actually wants to render like a `<Card>`? Like, in this case, 100% = 20x10 _tiles_ with background color?
How card knows it's actual available size?
Using other `<Box>` components inside so they rely on the outter one? but again, if only `Box` doesn't support what you want to do (in this case, background color) how to do it?

maticzav:
I don't think `flexGrow` is an option. I haven't experimented with flex-box limits, however, to my understanding that wouldn't solve the problem because `flexGrow` is only row-agnostic which leads to well-sized rows, stretched to fill the horizontal space, but doesn't account for the size of other cells in the column.

danikaze:
> I don't think `flexGrow` is an option. I haven't experimented with flex-box limits, however, to my understanding that wouldn't solve the problem because `flexGrow` is only row-agnostic which leads to well-sized rows, stretched to fill the horizontal space, but doesn't account for the size of other cells in the column.

exactly my point

jdeniau:
[...]

PR #307, opened 2020-06-20
Add measureElement() API

vadimdemedes:
Fixes https://github.com/vadimdemedes/ink/issues/168. cc @maticzav 

From readme:

> Use `measureElement()` function if you need to measure dimensions of a particular `<Box>` element.
It returns an object with `width` and `height` properties.
This function is useful when your component needs to know the amount of available space it has before it can start rendering or when you need to change your layout based on the length of the content.

```jsx
import {Box, measureElement} from 'ink';

const Example = () => {
	const ref = useRef();

	useEffect(() => {
		const {width} = measureElement(ref.current);
	}, []);

	return (
		<Box width={100}>
			<Box ref={ref}>{/* This box will stretch to 100 width */}</Box>
		</Box>
	);
};
```

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -1598,6 +1598,7 @@ This function is useful when your component needs to know the amount of availabl
 Type: `MutableRef`
 
 Reference to `<Box>` element captured via `ref` property.
+See [Refs](https://reactjs.org/docs/refs-and-the-dom.html) for more information on how to capture references.
 
 ```jsx
 import {Box, measureElement} from 'ink';
--- a/readme.md
+++ b/readme.md
@@ -1587,11 +1587,11 @@ clear();
 
 #### measureElement(ref)
 
-Measure dimensions of a particular `<Box>` element.
+Measure the dimensions of a particular `<Box>` element.
 It returns an object with `width` and `height` properties.
 This function is useful when your component needs to know the amount of available space it has before it can start rendering or when you need to change your layout based on the length of the content.
 
-**Note:** `measureElement()` returns correct results after initial render, when layout has been calculated.
+**Note:** `measureElement()` returns correct results only after the initial render, when layout has been calculated. Until then, `width` and `height` equal to zero. It's recommended to call `measureElement()` in a `useEffect` hook, which fires after component has rendered.
 
 ##### ref
 
--- a/readme.md
+++ b/readme.md
@@ -1601,7 +1601,7 @@ Reference to `<Box>` element captured via `ref` property.
 See [Refs](https://reactjs.org/docs/refs-and-the-dom.html) for more information on how to capture references.
 
 ```jsx
-import {Box, measureElement} from 'ink';
+import {render, measureElement, Box} from 'ink';
 
 const Example = () => {
 	const ref = useRef();
@@ -1616,6 +1616,8 @@ const Example = () => {
 		</Box>
 	);
 };
+
+render(<Example />);
 ```
 
 ## Testing
--- a/readme.md
+++ b/readme.md
@@ -1591,7 +1591,7 @@ Measure the dimensions of a particular `<Box>` element.
 It returns an object with `width` and `height` properties.
 This function is useful when your component needs to know the amount of available space it has before it can start rendering or when you need to change your layout based on the length of the content.
 
-**Note:** `measureElement()` returns correct results only after the initial render, when layout has been calculated. Until then, `width` and `height` equal to zero. It's recommended to call `measureElement()` in a `useEffect` hook, which fires after component has rendered.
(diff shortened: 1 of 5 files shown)


readme.md
```diff
@@ -837,6 +837,32 @@ Accepts the same values as [`color`](#color) in `<Text>` component.
 
 <img src="media/box-borderColor.jpg" width="228">
 
+##### Measuring dimensions
+
+Use `measureElement()` function if you need to measure dimensions of a particular `<Box>` element.
```
sindresorhus, line 842:
```suggestion
Use the `measureElement()` function if you need to measure dimensions of a particular `<Box>` element.
```

readme.md
```diff
@@ -837,6 +837,32 @@ Accepts the same values as [`color`](#color) in `<Text>` component.
...
+It returns an object with `width` and `height` properties.
+This function is useful when your component needs to know the amount of available space it has before it can start rendering or when you need to change your layout based on the length of the content.
+
+**Note:** `measureElement()` returns correct results after initial render, when layout has been calculated.
```
sindresorhus, line 846: The clear enough what implications this has for usage. What does it return before? How should the user handle this?

sindresorhus, line 846:
```suggestion
**Note:** `measureElement()` returns correct results only after the initial render, when layout has been calculated.
```

src/measure-element.ts
```diff
@@ -0,0 +1,21 @@
...
+	width: number;
+
+	/**
+	 * Element height
+	 */
+	height: number;
+}
+
+/**
+ * Measure dimensions of a particular `<Box>` element.
```
sindresorhus, line 16:
```suggestion
 * Measure the dimensions of a particular `<Box>` element.
```

src/measure-element.ts
```diff
@@ -0,0 +1,21 @@
+import type {DOMElement} from './dom';
+
+interface Output {
+	/**
+	 * Element width
+	 */
+	width: number;
+
+	/**
+	 * Element height
```
sindresorhus, line 10:
```suggestion
	 * Element height.
```

src/measure-element.ts
```diff
@@ -0,0 +1,21 @@
+import type {DOMElement} from './dom';
+
+interface Output {
+	/**
+	 * Element width
```
sindresorhus, line 5:
```suggestion
	 * Element width.
```

vadimdemedes:
Thanks @sindresorhus, updated!

readme.md
```diff
@@ -1587,11 +1587,11 @@ clear();
 
 #### measureElement(ref)
 
-Measure dimensions of a particular `<Box>` element.
+Measure the dimensions of a particular `<Box>` element.
 It returns an object with `width` and `height` properties.
 This function is useful when your component needs to know the amount of available space it has before it can start rendering or when you need to change your layout based on the length of the content.
 
-**Note:** `measureElement()` returns correct results after initial render, when layout has been calculated.
+**Note:** `measureElement()` returns correct results only after the initial render, when layout has been calculated. Until then, `width` and `height` equal to zero. It's recommended to call `measureElement()` in a `useEffect` hook, which fires after component has rendered.
```
sindresorhus, line 1594:
```suggestion
**Note:** `measureElement()` returns correct results only after the initial render, when layout has been calculated. Until then, `width` and `height` equal to zero. It's recommended to call `measureElement()` in a `useEffect` hook, which fires after the component has rendered.
```

readme.md
```diff
@@ -1598,6 +1598,7 @@ This function is useful when your component needs to know the amount of availabl
 Type: `MutableRef`
 
 Reference to `<Box>` element captured via `ref` property.
+See [Refs](https://reactjs.org/docs/refs-and-the-dom.html) for more information on how to capture references.
```
sindresorhus:
```suggestion
A reference to a `<Box>` element captured with a [`ref` property](https://reactjs.org/docs/refs-and-the-dom.html).
```

maticzav:
Oh wow! It looks fantastic.

I have some exams left in June; expect the `<table>` component in early July if somebody doesn't surpass me.

Thanks @vadimdemedes, Ink is so good. ❤️

vadimdemedes:
@maticzav Awesome, good luck on the exams!

readme.md
```diff
@@ -1585,6 +1585,44 @@ const {clear} = render(<MyApp />);
 clear();
 ```
 
+#### measureElement(ref)
+
+Measure the dimensions of a particular `<Box>` element.
+It returns an object with `width` and `height` properties.
+This function is useful when your component needs to know the amount of available space it has before it can start rendering or when you need to change your layout based on the length of the content.
```
maticzav, line 1592: This line sounds a bit confusing to me, especially the "before it can start rendering part".

My suggestion would be:

"This function is useful when your component needs to know the amount of available space it has. You could use it, for example, when you need to change layout based on the length of its content."

Issue #320, opened 2020-07-02
Ink v3 contains `@types` packages as dependencies


G-Rath:
Currently the `package.json` for v3 contains `@types/` packages as dependencies:

https://github.com/vadimdemedes/ink/blob/9ebfe314e865506ccb37b48723a27485346d950b/package.json#L41-L51

These should be listed as `devDependencies` :)

SimenB:
the ones present in the `.d.ts` files should remain as deps, but I doubt all of these are (if any?)

bartlangelaan:
I do think that it's better if they are listed as `*` instead of a version range, to make sure that only a single version per package is installed. That's also what the `@types/` packages themselves do.

vadimdemedes:
> These should be listed as devDependencies :)

Could you elaborate why?

G-Rath:
Sure! It's primarily for the same reasons that other dependencies are listed as `devDependencies`: they're not required by the library at runtime.

In theory none of these packages ever need to be listed as dependencies, but sometimes a library will use types from one or more of the packages in their typed API so those packages are sometimes included as `dependencies`. In that case, the definitions should usually be included using a looser version, depending on the type in question.

The other way you can do this is like what you've done with `@types/react`: mark the package as `optional` using `peerDependenciesMeta`.

Theres pros and cons to these two methods: the first (using `dependencies`) means that everyone will always get the `@types` packages that are required for TypeScript users to use your libraries types regardless of if they're using TypeScript or not; the second (using `peerDependenciesMeta`) means that the users typically have to manually include the `@types/` packages themselves if they're using TypeScript.

Looking over the codebase, I believe the only `@types` that should be in `dependencies` is `@types/react-reconciler`.

Constraint wise, typically `'*'` is recommended as mixing type versions can be messy, and it's usually easier to resolve problems with 1 package than with 2. The type packages are also not always 1:1 with the libraries they're typing in terms of major versions, so using `'*'` saves you from having to manage the overhead of doing i.e `24 | 28 | 29` as your constraint, nor will you have to make new releases updating those constraints because a new version of the `@types` was released & people are getting dependency warnings.

You should also always use `'*'` as the constraint for `@types/node` (if it's in `dependencies`, which it rarely needs to be), as it's the one dependency you can't actually control, and having two versions of it in your dependency tree can lead to some very confusing errors.

Hope that helps clarify things - let me know if you've got more questions! 😄

vadimdemedes:
[...]

PR #322, opened 2020-07-05
Move @types/* packages to dev dependencies

vadimdemedes:
Fixes https://github.com/vadimdemedes/ink/issues/320.

cc @G-Rath


package.json
```diff
@@ -75,6 +64,17 @@
 	},
 	"devDependencies": {
 		"@sindresorhus/tsconfig": "0.7.0",
+		"@types/is-ci": "^2.0.0",
+		"@types/lodash.throttle": "^4.1.6",
+		"@types/node": "*",
```
G-Rath, line 69:
```suggestion
		"@types/node": "^10.17.26",
```

Ideally this should be set to the lowest major supported by your `engines` field :)

package.json
```diff
@@ -75,6 +64,17 @@
 	},
 	"devDependencies": {
 		"@sindresorhus/tsconfig": "0.7.0",
+		"@types/is-ci": "^2.0.0",
+		"@types/lodash.throttle": "^4.1.6",
```
sindresorhus, line 68: You cannot move all types. This type, for example, is used in the source code: https://github.com/vadimdemedes/ink/blob/293d40fb6f85478ae0b8120709d71bf530d23eb2/src/ink.tsx#L3

G-Rath: That is only used in the compiled js, not in the TypeScript definitions, so it's not required by this package at runtime nor at downstream compile time :)

vadimdemedes: Ok I think I'm confused now. What's a question I need to ask myself to determine when a `@types/*` dependency should be included in `dependencies` or `devDependencies`? My understanding was that this question is "Are types from @types/* package are re-exported from my module in any way?".

G-Rath: > "Are types from @types/* package re-exported from my module in any way?".

That is the correct question to ask. 

I'm not sure why @sindresorhus thinks this type is required in `dependencies`, since it fails the above question.

SimenB: You need to keep at least `react`, `node`, and `yoga-layout` (the latter seems to ship its own types)

image

sindresorhus: Yeah, sorry. I picked a bad example.

sindresorhus: @SimenB `@types/node` should be a dev dependency, and maybe also an optional peer dependency, although I usually don't do that as it's expected that the user has that anyway if they use TypeScript.

sindresorhus: I really wish TS had diagnostics to warn when `@types/x` is in the wrong place or missing based on what types are used in the final package.

SimenB: If not a dep it should be an optional peer, yeah. Might make sense for react as well in this specific case?

SimenB: I've looked briefly at [`api-extractor`](https://api-extractor.com/), I _believe_ it can warn

milesj: Rule of thumb for when something is dev/normal dep. If you are referencing a type from `@types` as an argument, return, etc type, then it must be a normal dep (since it shows up in `.d.ts`), otherwise it's a dev dep. 

For example, this is would make `@types/debug` a normal dep:

```ts
import debug from 'debug';

class Example {
  debugger: debug.Debugger = debug('example');
}
```

vadimdemedes: @milesj Thanks for clarifying, that's a lot clearer! So it doesn't matter if I'm exporting types from 3rd-party packages in Ink's API? It should be a normal dependency, as long as it's referenced in Ink's source, even if only in "internal" code?

milesj: If you're re-exporting types that you expect a consumer to use, then it should be a normal dep. If there any specific examples in the source code, feel free to link them.

vadimdemedes: So if I'm not re-exporting those types, should they still be normal dependencies?

milesj: Only if you explicitly use a type from the package and use it in your public API (argument type, return type, property type, etc). When in doubt, you can look at the compiled `.d.ts` files.

But yeah, it would be nice if there was an eslint plugin for this.

vadimdemedes: Ok, In that case I think this PR is good to go as-is? None of the types from dependencies I've moved to `devDependencies` are exported from Ink.

SimenB: See https://github.com/vadimdemedes/ink/pull/322#discussion_r449919189, you need at least node and react - possibly optional peer deps tho

vadimdemedes: > @SimenB @types/node should be a dev dependency, and maybe also an optional peer dependency, although I usually don't do that as it's expected that the user has that anyway if they use TypeScript.

I think comment by @sindresorhus makes sense to me, if developer is using TypeScript, they most likely already have (and should) `@types/node` and `@types/react` installed.

vadimdemedes: @milesj seems to share the same opinion on this too, sounds logical to me.

milesj: Yeah, that's how I do it for most of my stuff. But if you want to add them, just do `*` versions to not conflict with consumer versions.

vadimdemedes:
Thanks everyone for feedback, very useful!

PR #326, opened 2020-07-06
Make sure that stack lines that cannot be parsed don't break the ErrorOverview

bartlangelaan:
I was testing Ink 3.0.0-4 and I had an error in my application. However, instead of showing the new ErrorOverview it showed an Ink error:

image

(TypeError: Cannot read property 'function' of null at ink/build/components/ErrorOverview.js:78:106)

Looking at the source code, it seems that a parseLine() function can return null, but that null values are not handled.

After this MR, it shows the line unparsed if it cannot be parsed - which makes sure that no error data gets lost.

image

Looks like that there is still some stuff that can be done to handle React errors better (and also parse Component data). But at least this handles the case where parseLine returns null, which should be handled either way.

### DIFF
--- a/src/components/ErrorOverview.tsx
+++ b/src/components/ErrorOverview.tsx
@@ -83,7 +83,19 @@ const ErrorOverview: FC<Props> = ({error}) => {
 						.split('\n')
 						.slice(1)
 						.map(line => {
-							const parsedLine = stackUtils.parseLine(line)!;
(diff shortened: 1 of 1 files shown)


vadimdemedes:
Released in [`v3.0.0-6`](https://github.com/vadimdemedes/ink/releases/tag/v3.0.0-6).

PR #330, opened 2020-07-12
Fix type of `<Box>` to include `ref` prop

### LINKED ISSUE #329: Types for `<Box>` don't include the `ref` prop
Currently the `Box` component is explicitly typed as `React.FC`, which doesn't include the `ref` prop. Instead typings should be inferred from the `forwardRef()` function.

flut1:
By explicitly typing `Box` as `React.FC` the `ref` prop is removed from typings. Instead let TypeScript infer the type from the return type of the `forwardRef()` function. (in this case `ForwardRefExoticComponent`)

To make sure `Box` still has the "children" prop, wrap `Props` in `PropsWithChildren<>`.

Fixes #329

### DIFF
--- a/src/components/Box.tsx
+++ b/src/components/Box.tsx
@@ -1,6 +1,6 @@
 /* eslint-disable @typescript-eslint/prefer-nullish-coalescing */
 import React, {forwardRef} from 'react';
-import type {FC} from 'react';
+import type {PropsWithChildren} from 'react';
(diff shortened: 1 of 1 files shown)


vadimdemedes:
Good catch, thanks!

vadimdemedes:
Landed in [3.0.0-7](https://github.com/vadimdemedes/ink/releases/tag/v3.0.0-7).

PR #331, opened 2020-07-14
Skip unparseable error lines

elmpp:
The `stack-utils` library doesn't support React error traces which caused a further error within `ErrorOverview`, masking the true cause

Before:
Screenshot 2020-07-14 at 23 46 03

After:
Screenshot 2020-07-14 at 23 45 40


bartlangelaan:
This should already be fixed with #326 (which is available since [the 3.0.0-6 release](https://github.com/vadimdemedes/ink/releases/tag/v3.0.0-6)).

PR #334, opened 2020-07-19
docs: add links to top with jump2header

strdr4605:
Generated with:

```
jump2header --header contents --start "Getting Started" --end "Examples" --maxLevel 3
```

[Demo](https://github.com/strdr4605/ink/blob/jump2header/readme.md#getting-started).

If you don't like the PR but interested in links to top see [jump2header](https://github.com/strdr4605/jump2header#readme) for more options.


sindresorhus:
Thanks for the suggestion, but this is not something we want.

PR #338, opened 2020-07-28
Improve auto-complete for `color` and `backgroundColor` props for `Text`

iamssen:
# Changes

image

In the JetBrains IDEs would suggest colors like the above image.

image

But, in VSCode it does not work. 😨

This may not be complete, but it would help JetBrains IDE users.

### DIFF
--- a/src/components/Text.tsx
+++ b/src/components/Text.tsx
@@ -3,17 +3,19 @@ import type {FC, ReactNode} from 'react';
 import chalk from 'chalk';
 import colorize from '../colorize';
 import type {Styles} from '../styles';
+import type {LiteralUnion} from 'type-fest';
+import type {ForegroundColor, BackgroundColor} from 'chalk';
 
 export interface Props {
 	/**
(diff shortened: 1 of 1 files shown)


sindresorhus:
You could use https://github.com/sindresorhus/type-fest/blob/master/source/literal-union.d.ts to make it work in VS Code too. I use this type a lot.

Example: https://github.com/sindresorhus/boxen/blob/0985f4de83cfe4a6b67327099a7a55cae7b1297a/index.d.ts#L40-L60

iamssen:
@sindresorhus Thank you for your guide!

image

In the VSCode work fine.

image

😢 But, JetBrains IDEs not work...

I think this is better because there are more VSCode users.

SimenB:
Probably worth reporting to them so they can align: https://youtrack.jetbrains.com/newIssue

iamssen:
@SimenB <https://youtrack.jetbrains.com/issue/WEB-46671>

I have created an issue in JetBrains. I hope this will be a chance to support better code assist.

vadimdemedes:
Thanks for the improvement!

PR #339, opened 2020-07-28
chore: remove `esModuleInterop` and add `importsNotUsedAsValues: 'error'`

SimenB:
I noticed `import type` was used, so went to add a compiler option to enforce it. Ending up fixing a few stragglers which didn't use it. Also removed `esModuleInterop` so it's not enforced downstream for consumers. Happy to split this up into 2 PRs if you want.


readme.md
```diff
@@ -154,7 +154,7 @@ $ npm install --save-dev @babel/preset-react
 Next, create a file `source.js`, where you'll type code that uses Ink:
 
 ```jsx
-import React from 'react';
+import * as React from 'react';
```
SimenB, line 157: Caught by my search&replace, but also recommended in "ESM", albeit under discussion: https://github.com/facebook/react/issues/11503#issuecomment-662647468

sindresorhus:
Did you see https://twitter.com/drosenwasser/status/1218253620077981696 ?

SimenB:
Why is it supposed to be painful?

sindresorhus:
But what's the benefit?

vadimdemedes:
I used it only to be explicit what is being imported and thought it was a good idea after reading https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-exports. I don't have a strong opinion on it though.

SimenB:
> But what's the benefit?

Main benefit in my mind is mostly unblocking `isolatedModules` - turning on isolated modules _without_ type imports is nearly impossible (at least in my experience). And the benefit to using isolated modules is that it works correctly with other tools such as Babel if you ever switch to using that (due to wanting to use some extra babel transforms or whatever).

vadimdemedes:
@SimenB Thanks a lot for working on this PR, but I was researching this topic and `import type` doesn't indeed seem to have a lot of benefits, especially if `esModuleInterop` has to be turned off.

vadimdemedes:
We can certainly revisit in the future.

SimenB:
No problem at all! 🙂

PR #343, opened 2020-07-30
Removed 'which is unreleased yet' phrase

elijahmanor:
Removed "which is unreleased yet" phrase from readme.md since v3 has been released!


vadimdemedes:
Thanks!

PR #355, opened 2020-08-05
Fix `Static` example code

lucas-lm:
Variable `timer` is undefined.
Variable `timer` should get the setTimeout return in order to be used in clear timeout on clean up function.


lucas-lm:
Minor fix

vadimdemedes:
Thanks!

PR #327, opened 2020-07-06
Fall back to 80 columns if the width of the terminal cannot be determined

bartlangelaan:
Fixes #325 by using 80 columns if `process.stdout.columns` is `undefined`.

Issue #325, opened 2020-07-05
"Invalid value undefined for setWidth" in standard-version


jaredLunde:
This yoga error occurs when a command using Ink v3 happens in husky + [standard-version](https://github.com/conventional-changelog/standard-version). I would assume this happening because the hook runs in an environment where there is no terminal width to read.

Perhaps there should be a check/default on the `terminalWidth` argument here?
https://github.com/vadimdemedes/ink/blob/155e1a042b12e82929bdb26571385b5dea0f0b39/src/renderer.ts#L12-L13

```
Error: Invalid value undefined for setWidth
    at Bound.<anonymous> (/Users/jared/Code/oss/lundle/node_modules/yoga-layout-prebuilt/yoga-layout/dist/entry-common.js:167:17)
    at Bound.prototype.<computed> [as setWidth] (/Users/jared/Code/oss/lundle/node_modules/yoga-layout-prebuilt/yoga-layout/dist/entry-common.js:133:22)
    at Object.exports.default (/Users/jared/Code/oss/lundle/node_modules/ink/build/renderer.js:11:19)
    at Ink.onRender (/Users/jared/Code/oss/lundle/node_modules/ink/build/ink.js:49:78)
    at Ink.unmount (/Users/jared/Code/oss/lundle/node_modules/ink/build/ink.js:172:14)
    at EventEmitter.emit (events.js:315:20)
    at emit (/Users/jared/Code/oss/lundle/node_modules/signal-exit/index.js:78:11)
    at processEmit [as emit] (/Users/jared/Code/oss/lundle/node_modules/signal-exit/index.js:156:5)
npm ERR! code ELIFECYCLE
npm ERR! errno 7
npm ERR! lundle@0.4.8 check-types: `FORCE_COLOR=1 node dist/main/bin check-types`
npm ERR! Exit status 7
npm ERR! 
npm ERR! Failed at the lundle@0.4.8 check-types script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/jared/.npm/_logs/2020-07-05T21_38_36_580Z-debug.log
husky > pre-commit hook failed (add --no-verify to bypass)
```

sindresorhus:
It's common to default to 80x24 when the terminal size cannot be determined.

bartlangelaan:
I added this workaround to my own app, which removes the error.

```process.stdout.columns = process.stdout.columns || 80;```

PR #368, opened 2020-08-31
Fallback to 80 if number of columns isn't set

### LINKED ISSUE #366: CI rendering is broken since 3.0.0
@vadimdemedes, as I said in https://github.com/vadimdemedes/ink/issues/275#issuecomment-682007306 earlier, I've been trying to narrow down an issue with rendering related to continuous integration services. I managed to narrow it to the Ink 3 release.

I ran `npx create-ink-app`, pasted the counter from the README into `ui.js`, and added an effect which calls exit once the count reaches 10;

```javascript
const React = require("react");
const { useState, useEffect } = React;
const { render, Text, useApp } = require("ink");

const Counter = () => {
  const { exit } = useApp();
  const [counter, setCounter] = useState(0);

  useEffect(() => {
    const timer = setInterval(() => {
      setCounter((previousCounter) => previousCounter + 1);
    }, 100);

    return () => {
      clearInterval(timer);
    };
  }, []);

  useEffect(() => {
    if (counter === 10) {
      exit();
    }
  }, [counter, exit]);

  return <Text color="green">{counter} tests passed</Text>;
};

module.exports = Counter;
```

Locally that works as expected; the count is incremented and then it exits. 10 is rendered.

On CircleCI however, this is what's outputted:

image

No final frame is rendered and there's a bunch of extra newlines.

And it works fine on CircleCI with ink@2.7.1;

image

Please let me know if I'm using `exit` incorrectly or something like that. I'm happy to make a PR but I just need some feedback

adam-lynch:
Fixes #366. 

Side note: As I said in that issue, `node.yogaNode.getComputedWidth` returns `0`, but I didn't touch that. This single change solves my issue.

Related:

- #325
- #327

cc @bartlangelaan

### DIFF
--- a/src/ink.tsx
+++ b/src/ink.tsx
@@ -115,9 +115,9 @@ export default class Ink {
 
 		const {output, outputHeight, staticOutput} = render(
 			this.rootNode,
-			// The 'columns' property can be undefined when not using a TTY.
+			// The 'columns' property can be undefined or 0 when not using a TTY.
 			// In that case we fall back to 80.
-			this.options.stdout.columns ?? 80
+			this.options.stdout.columns || 80
 		);
 
 		// If <Static> output isn't empty, it means new children have been added to it


vadimdemedes:
Great catch, thanks!

vadimdemedes:
CI is failing due to an unrelated error fixed via https://github.com/vadimdemedes/ink/commit/9f6343a35d8b0b792f68f25794ea5b83898dc0cb.

adam-lynch:
No prob

PR #372, opened 2020-09-09
Fix error of text string outside text box in example

seanjmurray:
ERROR  Text string "Press Tab to focus next element, Shift+Tab to focus previous element, Esc to reset focus." must be rendered inside <Text> component


vadimdemedes:
Thanks!

PR #379, opened 2020-09-19
Fix `borderColor` type

matejbudzel:
Type of `borderColor` property in `Styles` interface can (and should) match the `color` property of `Text` component, but it did not. It was limited to basic foreground colors without the option use string values.


sindresorhus (approved):

vadimdemedes:
Thanks!

PR #380, opened 2020-09-19
Add TypeScript setup guide

LitoMore:
BTW, I think TypeScript can be used as the preferred option for creating apps in the future.

I also created a PR https://github.com/vadimdemedes/create-ink-app/pull/5 for creating TypeScript Ink app  `create-ink-app`.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -126,6 +126,9 @@ $ npx create-ink-app
 ```
 
 <details><summary>Manual setup</summary>
+
+<h3>Node.js</h3>
+
 <p>
 Ink requires the same Babel setup as you would do for regular React-based apps in the browser.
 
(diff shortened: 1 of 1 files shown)


vadimdemedes:
@sindresorhus Do you think we need something like this? My thinking is because Ink doesn't require any special treatment to work with TypeScript, we don't need a guide for it. Normally libraries don't include guides like this, because it looks more like an intro to TypeScript itself, which in my opinion doesn't belong to Ink or other projects' readme.

LitoMore:
I think it's useful for some TypeScript beginners.

sindresorhus:
Yeah, I don't see the need for this either. Every single project shouldn't have to document how to use TypeScript. If you instead publish it as a blog post with some more content, we could link to it.

LitoMore:
Make sense, thanks.

PR #385, opened 2020-09-25
Import devtools only when DEV=true

### LINKED ISSUE #384: Ink defines global window object
Ink defines a global `window` object. As far as I can tell, this is done exclusively [for devtools support](https://github.com/vadimdemedes/ink/blob/master/src/devtools.ts#L12-L14).

Defining a global `window` breaks other packages which assume the existence of a web browser based on the existence of `window` (in my case this is [`vue-template-compiler`](https://npmjs.com/package/vue-template-compiler)).

Could we somehow avoid setting a global `window` variable, or only do it if `DEV=true` is set?

loilo:
As discussed in #384.

However, after also converting the `ws` module import to `require()` to avoid loading unused code, I basically had the whole `devtools.ts` wrapped in an `if` block. At that point, I decided to roll back my changes to that file and just go for conditionally importing the whole devtools file.

Fixes #384

### DIFF
--- a/src/devtools.ts
+++ b/src/devtools.ts
@@ -1,6 +1,8 @@
 // Ignoring missing types error to avoid adding another dependency for this hack to work
 // @ts-ignore
 import ws from 'ws';
+// @ts-ignore
+import {connectToDevTools} from 'react-devtools-core';
 
 const customGlobal = global as any;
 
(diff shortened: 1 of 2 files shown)


vadimdemedes:
Great, thanks!

PR #388, opened 2020-10-02
Added ink-select-input-horizontal to useful components

H3RSKO:
Added ink-select-input-horizontal to useful components on Readme

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -1727,7 +1727,7 @@ You can even inspect and change the props of components, and see the results imm
 - [ink-text-input](https://github.com/vadimdemedes/ink-text-input) - Text input.
 - [ink-spinner](https://github.com/vadimdemedes/ink-spinner) - Spinner.
 - [ink-select-input](https://github.com/vadimdemedes/ink-select-input) - Select (dropdown) input.
-- [ink-select-input-horizontal](https://github.com/H3RSKO/ink-select-input-horizontal) - Mod of the above component adding color and orientation customization.
+- [ink-select-input-horizontal](https://github.com/H3RSKO/ink-select-input-horizontal) - Modification of the above component adding color and orientation customization.
 - [ink-link](https://github.com/sindresorhus/ink-link) - Link component.
 - [ink-gradient](https://github.com/sindresorhus/ink-gradient) - Gradient color component.
 - [ink-big-text](https://github.com/sindresorhus/ink-big-text) - Awesome text component.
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -1727,6 +1727,7 @@ You can even inspect and change the props of components, and see the results imm
 - [ink-text-input](https://github.com/vadimdemedes/ink-text-input) - Text input.
 - [ink-spinner](https://github.com/vadimdemedes/ink-spinner) - Spinner.
 - [ink-select-input](https://github.com/vadimdemedes/ink-select-input) - Select (dropdown) input.
+- [ink-select-input-horizontal](https://github.com/H3RSKO/ink-select-input-horizontal) - Mod of the above componenet adding color and orientation customization.
```
sindresorhus, line 1730: "mod" is not a word, and you have a typo.

sindresorhus:
Why not contribute to the original component instead?

readme.md
H3RSKO: Fixed typo and changed 'mod' to modification. Thought I would be going over line length but apparently not. 

Regarding the original component, we had put this together for our own project which had different default settings, and only after added in the customizability. 

I can work on modifying the original instead. Just give me a few days.

vadimdemedes:
@sindresorhus I think we've discussed this in the original repo somewhere and I advised to make it a separate package to keep the original one simple. I think the ideal strategy here is to keep these components separate, but extract their functionality into a React hook and reuse it between the two.

vadimdemedes:
Hey @H3RSKO, there are some issues in your readme:

1. Indentation is off in the `require()` code example
2. "Looking for a version compatible with Ink 2.x?" link points to `ink-text-input`.
3. Could you link to [ink-select-input](https://github.com/vadimdemedes/ink-select-input) in "This is a modified version of the original ink-select-input" text?
4. "Default" is on the same line as "Type" in [initialIndex](https://github.com/H3RSKO/ink-select-input-horizontal#initialindex) description.

sindresorhus:
Doesn't seem like it would add a lot of complexity though. It just a `direction` prop with some extra handling: https://github.com/H3RSKO/ink-select-input-horizontal/commit/27a1f085a7416c22058ffe451af4b31f7d692122

vadimdemedes:
That's true, didn't think it would be so simple. Sorry @H3RSKO, would you be able to add it to `ink-select-input` itself?

H3RSKO:
Hey guys, I posted a PR here: https://github.com/vadimdemedes/ink-select-input/pull/35#issue-498076509

As i mentioned there, I was having trouble with NPM link so please confirm it works before adding. 

Thanks!

vadimdemedes:
Cool, closing this one then, let's continue discussion in that PR.

PR #391, opened 2020-10-19
Make `patchConsole` option support a function

milesj:
This updates the `patchConsole` option to support a boolean (existing functionality) or a function.

I currently patch the console manually in my project, but it doesn't interweave correctly with Ink's output. But I can't use Ink's `patchConsole` because I need more control of the console overrides. However, if I use `patch-console` myself outside of Ink, I still have the same problems as before (I don't have access to Ink's stderr/stdout streams).

I feel like this is a nice compromise, as I can provide my own `patchConsole` function AND hook into Ink's rendering flow.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -1576,14 +1576,14 @@ This is needed in case `process.stdin` is in [raw mode](https://nodejs.org/api/t
 
 ###### patchConsole
 
-Type: `boolean`\
+Type: `boolean | (callback: (stream: 'stdout' | 'stderr', data: string) => void) => () => void`\
 Default: `true`
 
 Patch console methods to ensure console output doesn't mix with Ink output.
 When any of `console.*` methods are called (like `console.log()`), Ink intercepts their output, clears main output, renders output from the console method and then rerenders main output again.
 That way both are visible and are not overlapping each other.
 
-This functionality is powered by [patch-console](https://github.com/vadimdemedes/patch-console), so if you need to disable Ink's interception of output but want to build something custom, you can use it.
+This functionality is powered by [patch-console](https://github.com/vadimdemedes/patch-console), so if you need to disable Ink's interception of output but want to build something custom, you can use it. You can also provide your own patch function to `patchConsole` with the same API.
 
 ###### debug
 
--- a/readme.md
+++ b/readme.md
@@ -1583,7 +1583,28 @@ Patch console methods to ensure console output doesn't mix with Ink output.
 When any of `console.*` methods are called (like `console.log()`), Ink intercepts their output, clears main output, renders output from the console method and then rerenders main output again.
 That way both are visible and are not overlapping each other.
 
-This functionality is powered by [patch-console](https://github.com/vadimdemedes/patch-console), so if you need to disable Ink's interception of output but want to build something custom, you can use it. You can also provide your own patch function to `patchConsole` with the same API.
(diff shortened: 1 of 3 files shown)


readme.md
```diff
@@ -1576,14 +1576,14 @@ This is needed in case `process.stdin` is in [raw mode](https://nodejs.org/api/t
...
-Type: `boolean`\
+Type: `boolean | (callback: (stream: 'stdout' | 'stderr', data: string) => void) => () => void`\
 Default: `true`
 
 Patch console methods to ensure console output doesn't mix with Ink output.
 When any of `console.*` methods are called (like `console.log()`), Ink intercepts their output, clears main output, renders output from the console method and then rerenders main output again.
 That way both are visible and are not overlapping each other.
 
-This functionality is powered by [patch-console](https://github.com/vadimdemedes/patch-console), so if you need to disable Ink's interception of output but want to build something custom, you can use it.
+This functionality is powered by [patch-console](https://github.com/vadimdemedes/patch-console), so if you need to disable Ink's interception of output but want to build something custom, you can use it. You can also provide your own patch function to `patchConsole` with the same API.
```
sindresorhus, line 1586: Maybe include your use-case as it might not be clear to the user what they would use this capability for.

milesj: Added a basic example. Wasn't sure what to do without making it look overly complicated.

milesj:
Was that failing test expected? Seems unrelated to my change.

vadimdemedes:
Ink gets stdout and stderr streams from `process.stdout` and `process.stderr` respectively, so `patch-console` should work? Could you share more about your use case and how a `patchConsole` function solves it?

milesj:
@vadimdemedes I'm looking for the ability to intercept all `console.*` messages and format the message string by adding prefixes, suffixes, timestamps, etc. Basically more metadata.

vadimdemedes:
I see, I think you can achieve that by disabling `patchConsole` and using the `patch-console` module directly. To make sure it plays nice with Ink, you can catch the logs, format them, then use [`useStdout`](https://github.com/vadimdemedes/ink#usestdout) and [`useStderr`](https://github.com/vadimdemedes/ink#usestderr) hooks to output each message. That's how Ink patches console itself, so you'd get the exact same behavior.

milesj:
@vadimdemedes I'm actually doing something like that now, but the output gets quite messed up and out of sync. I'm assuming its a timing issue as I have to pass the logs from outside the React layer into it to write using the hooks. https://github.com/milesj/boost/blob/master/packages/cli/src/LogWriter.tsx

Could maybe be a bit easier if I had access to stdout/err write methods _outside_ of React? Maybe as a return of render?

```
const { writeToStdout, writeToStderr } = render(<App />);
```

milesj:
Did some more digging, and it turns out `debug` is the culprit for a lot of the UI trashing, since it writes to stderr directly. https://github.com/visionmedia/debug/blob/master/src/node.js#L194

I feel like this may be a problem for many packages but I'm not sure what a good solution is.

milesj:
I ended up patching over `debug` also to fix this problem. Seems to have resolved all of the issues so far. https://github.com/milesj/boost/pull/116/files#diff-8619ceacae3dd4e98002efd562ba51ceec745f1468e604da5ad4a0424b6d887eR87

I'll close this for now.

Issue #222, opened 2019-08-29
Scrolling


arcanis:
Hello! I've been playing with Ink yesterday, and I really like where you're going. I've worked on [something similar](https://manaflair.github.io/mylittledom/demo/#complex-form) before, but your approach seems much simpler and less labyrinthine. Kudos!

One particular thing I'm finding myself missing, however, are scrollable containers. I anticipate that printing a lot of data will eventually cause my display to go beyond the height of a regular terminal (particularly when working with lists of interactive elements), so scroll will be quite important.

I'm trying to think how would one implement this feature through userland blocks, but I don't see how it could be done - I remember implementing this in my library required to first support overflows, meaning computing intersections between the various nodes printed on screen.

Have you had some thoughts about?

vadimdemedes:
Hi, thanks! I want to merge https://github.com/vadimdemedes/ink/pull/210 to solve this problem, I just want to test it more extensively. With this PR seems you won't need scrollable containers anymore.

arcanis:
Hm but that doesn't solve the issue for interactive elements, right? For example, if we imagine a list of radio buttons, the ones beyond the top of the screen would never be visible without first-party scrolling (terminals support some basic scroll, but typing anything would put back the scroll to the bottom).

vadimdemedes:
Yep, in this case components have to implement their own scrolling, similar to https://github.com/vadimdemedes/ink-select-input.

zaripych:
Maybe what @arcanis means here is something of a general purpose scrolling component that can wrap any number of components as children and allow scrolling through it. I tried to write something like that and bumped into a few issues:

- there seem to be no way to measure height of the child nodes (it is possible in `react-native` that also uses `yoga`, using `measure` functions)
- there seem to be no way to determine height of the viewport (the scroll view component itself), the height must be fixed
- there is no parameter similar to `overflow` like in `react-native` where all sub-components of a Box would have been clipped. neither there is `zIndex`. as a result if the scroll view is just next to a header, scrolling through elements will render on top of the header

```javascript
interface IScrollViewProps {
  totalHeight: number;
  viewportHeight: number;
}

export const VerticalScrollView: React.FC<IScrollViewProps> = props => {
  const thumbHeight = 3;

  const [offset, setOffsetCore] = React.useState(0);

  const thumbPosition =
    (offset / (props.totalHeight - (props.viewportHeight - thumbHeight))) *
    props.viewportHeight;

[...]

PR #393, opened 2020-10-28
Add `overflow` prop to `Box` component

tin-t:
Adds an `overflow` property to `<Box>` with values of `"visible"` (default) and `"hidden"`. When set to `"hidden"`, any content not inside the element's border will not be rendered.

This can be useful for making scrolling elements for arbitrary content such as the implementation posted in #222.

### DIFF
--- a/src/styles.ts
+++ b/src/styles.ts
@@ -141,6 +141,11 @@ export interface Styles {
 	 * Accepts the same values as `color` in <Text> component.
 	 */
 	readonly borderColor?: LiteralUnion<typeof ForegroundColor, string>;
+
+	/**
+	 * Set this property to `hidden` to hide content overflowing the box.
+	 */
+	readonly overflow?: 'hidden' | 'visible';
 }
 
 const applyPositionStyles = (node: Yoga.YogaNode, style: Styles): void => {
--- a/readme.md
+++ b/readme.md
@@ -785,6 +785,14 @@ Default: `flex`
 
 Set this property to `none` to hide the element.
 
+##### overflow
+
+Type: `string`\
+Allowed values: `visible` `hidden`\
+Default: `visible`
+
+Set this property to `hidden` to hide content that overflows this element.
+
 #### Borders
 
 ##### borderStyle
--- a/src/output.ts
+++ b/src/output.ts
@@ -100,7 +100,7 @@ export default class Output {
 					}
 
 					if (endX - startX < length) {
-						line = sliceAnsi(line, startX - x, endX - startX);
+						line = sliceAnsi(line, startX - x, endX - x);
 					}
 				}
 
--- a/src/output.ts
+++ b/src/output.ts
@@ -1,6 +1,6 @@
 import sliceAnsi from 'slice-ansi';
 import stringLength from 'string-length';
-import {OutputTransformer} from './render-node-to-output';
+import {Bounds, OutputTransformer} from './render-node-to-output';
 
 /**
  * "Virtual" output class
@@ -21,6 +21,7 @@ interface Writes {
 	y: number;
 	text: string;
 	transformers: OutputTransformer[];
+	bounds?: Bounds;
 }
 
 export default class Output {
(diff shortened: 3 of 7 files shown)


prozacgod:
How close would something like this be to allowing a line start / offset prop and giving the box container the ability to scroll content?

tin-t:
Here's an example of both horizontal and vertical scrolling (using the technique from https://github.com/vadimdemedes/ink/issues/222#issuecomment-572450030):
```jsx
const React = require('react');
const {render, Box, Text} = require('ink');

const Scroller = ({ scrollX, scrollY, width, height, children, ...props }) => (
	<Box overflow="hidden" width={width} height={height} {...props}>
		<Box flexDirection="column" marginTop={-scrollY} marginLeft={-scrollX} flexGrow={0} flexShrink={0}>
			{children}
		</Box>
	</Box>
);

const Main = () => {
	const [scrollX, setScrollX] = React.useState(0);
	const [scrollXDir, setScrollXDir] = React.useState(1);
	const [scrollY, setScrollY] = React.useState(0);
	const [scrollYDir, setScrollYDir] = React.useState(1);
	
	React.useEffect(() => {
		if(scrollX + scrollXDir < -5 || scrollX + scrollXDir > 5){
			setScrollXDir(scrollXDir * -1);
			return;
		}
		
		setTimeout(() => setScrollX(scrollX + scrollXDir), 100);
	}, [scrollX, scrollXDir])
	
	React.useEffect(() => {
		if(scrollY + scrollYDir < -7 || scrollY + scrollYDir > 7){
			setScrollYDir(scrollYDir * -1);
			return;
		}
		
		setTimeout(() => setScrollY(scrollY + scrollYDir), 100);
	}, [scrollY, scrollYDir])
	
	return <Scroller scrollX={scrollX} scrollY={scrollY} width={24} height={12} borderColor="white" borderStyle="single">
		<Box width={10} height={5} borderColor="white" borderStyle="single" flexShrink={0}>
			<Text>box 1</Text>
		</Box>
		<Box width={20} height={10} borderColor="white" borderStyle="single" flexShrink={0}>
			<Text>box 2</Text>
		</Box>
	</Scroller>
}

render(<Main />);
```

prozacgod:
Excellent,  Have you seen the little snippet I made for tracking terminal size/fullscreen boxe's?  

https://github.com/vadimdemedes/goodness-squad/issues/5#issuecomment-765111320

Give your work a try with that, and see how it goes.

I may make a fullscreen context so the terminal size is within the react ecosystem - providing side to child components.

gnidan:
Any movement on this? Would love to have use of this feature!

gnidan:
Is it possible to use this implementation to compute the unbounded overflow dimension?

To show what I mean, I want to have an inner box that exceeds the bounds of an `overflow="hidden"` outer box:
```
   .- outer
   |
   v
.-------.
|       |
|   .---+- -.
|   |###| / !
|   |###|/ /!
|   |###| / !
'---+---'/ /!
    ! / / / !
    !/ / / /!
    ! / / / !
    '- - - -'
       ^
       |
       '- inner
```

How can I measure the dimensions of the inner box? Is this possible at all? I've been messing around with the various methods defined by the `YogaNode` type (`getComputedBorder()`, etc.), but none of them seem to work for this. I'm following [the above example](https://github.com/vadimdemedes/ink/pull/393#issuecomment-765744123), which makes use of a 100ms `setTimeout`, but I'm not sure if this is correct here.

Any thoughts? Thanks in advance!

dustinlacewell:
Let's merge this

Songkeys:
Can we get this merged?

lordofthelake:
@vadimdemedes Is there anything that can be done to help merging this? It enables several useful patterns, it would be a really nice to have

src/styles.ts
```diff
@@ -141,6 +141,11 @@ export interface Styles {
 	 * Accepts the same values as `color` in <Text> component.
 	 */
 	readonly borderColor?: LiteralUnion<typeof ForegroundColor, string>;
+
+	/**
+	 * Set this property to `hidden` to hide content overflowing the box.
```
sindresorhus, line 146:
```suggestion
	 * Set this property to `hidden` to hide content overflowing the box.
	 *
	 * @default 'visible'
```

dustinlacewell:
> The most needed features are always not merged..
> 
> Here we go again

If there is enough interest, we can always create a community fork of this project and manage it that way. That said, the maintainer is probably just busy but holding out hope they will have the bandwidth to continue to maintain it. I've been in that position before, it's tough.

It really does pain me this library is not more actively developed and popular. It's fricken react on the terminal! Haha I love it.

Thanks again @vadimdemedes

PR #399, opened 2020-11-24
Added MemoryRouter Routing recipe with usage information

### LINKED ISSUE #396: Suggestion : Add react-router/MemoryRouter to list of useful components in the Docs
I've been looking for a solution to implement routing in `ink`, in particular I saw this https://github.com/jimmed/ink-router which looks abandoned but let me understand I was not the first one to take a look at this need,

I ended up looking at the source code of 'react-router' to implement my own quick and dirty solution, and then realized that `MemoryRouter` could do the job, and indeed, I started using it in Ink and it worked just fine. For the record I have been managing Router state trough history functions, in particular `history.push`, and consuming context state through `Switch/Route`.

advl:
Fixes #396 

+ Added a recipe for using `react-router/MemoryRouter` in the readme.md
+ Added an example for triggering routing using `vadimdemedes/ink-select-input`. I think this example is useful, but it requires an additional package.
~~+ Added an example of leveraging routing with the node `universal-analytics` package (Google Analytics). Also requires an additional package on top of the router and ink.~~
+ Both examples are tested.
~~+ I also allowed myself to include a link to my resume application which uses this pattern. I can remove if if it is not the right place.~~

I hope the recipe and the examples are clear and simple enough.


sindresorhus:
This should ideally have been multiple PRs.

sindresorhus:
> Added an example of leveraging routing with the node universal-analytics package (Google Analytics). Also requires an additional package on top of the router and ink.

I don't really see what this adds for Ink users. I think the `select-input` example already covers everything useful. I would also not want to encourage users to add analytics to their CLIs.

> I also allowed myself to include a link to my resume application which uses this pattern. I can remove if if it is not the right place.

We decided not to include such obscure CLIs. See: https://github.com/vadimdemedes/ink/commit/c9094c2ef57a1755bb89cbed27d77a617ae7dfec

readme.md
```diff
@@ -1752,6 +1753,81 @@ You can even inspect and change the props of components, and see the results imm
...
+      </Box>
+    </>
+  );
+};
+
+const AppWrapper = () => {
+
+  return(
+    <MemoryRouter>
+        <App />
```
sindresorhus, line 1815: Use tab-indentation.

advl:
Thank you for your review. I just committed the necessary fixes.

readme.md
```diff
@@ -1765,6 +1839,7 @@ You can even inspect and change the props of components, and see the results imm
 - [Write to stdout](examples/use-stdout/use-stdout.js) - Write to stdout bypassing main Ink output.
 - [Write to stderr](examples/use-stderr/use-stderr.js) - Write to stderr bypassing main Ink output.
 - [Static](examples/static/static.js) - Use `<Static>` to render permanent output.
+- [Router with Select Input](examples/router/select-input.js) - Change the url of the Router with a UI menu. Requires the additional packages `react-router` and [ink-select-input](https://github.com/vadimdemedes/ink-select-input).
```
sindresorhus, line 1842: I think the description here could be improved. The `Requires the additional packages `react-router` and [ink-select-input](https://github.com/vadimdemedes/ink-select-input).` part is also not relevant and can be documented in the file itself if needed.

examples/router/select-input.js
```diff
@@ -0,0 +1,57 @@
+import * as React from 'react';
```
sindresorhus, line 1: The example is not runnable. See the other examples.

sindresorhus (changes requested):

readme.md
```diff
@@ -1752,6 +1752,80 @@ You can even inspect and change the props of components, and see the results imm
 
 - [ink-use-stdout-dimensions](https://github.com/cameronhunter/ink-monorepo/tree/master/packages/ink-use-stdout-dimensions) - Subscribe to stdout dimensions.
 
+## Recipe : Routing using MemoryRouter
+
+If you desire to mirror the routing of a browser React app, the package `react-router` provides the useful `MemoryRouter`. 
+
+The following example provides a basic implementation.
+
+```jsx
```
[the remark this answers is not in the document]
advl: I tried to make it a little bit more simple. The most important is that the Routing Context needs to be provided higher in the tree, hence the need for a second component "wrapper".

readme.md
```diff
@@ -1752,6 +1752,80 @@ You can even inspect and change the props of components, and see the results imm
 
 - [ink-use-stdout-dimensions](https://github.com/cameronhunter/ink-monorepo/tree/master/packages/ink-use-stdout-dimensions) - Subscribe to stdout dimensions.
 
+## Recipe : Routing using MemoryRouter
```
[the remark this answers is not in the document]
advl: ✔️

examples/router/select-input.js
advl: The other examples in the `examples` folder are not enough since they don't require external packages. The best I could do, atm, was to make the example runnable out-the-box by using it to replace the `ui.js` file in a newly generated `npx creat-ink-app` project, after installing both dependencies `npm i react-router ink-select-input`. This seems a bit complicated to me. 

My suggestion is remove the example `router/select-input` and integrate it into `docs/recipes/routing-with-react-router.md`, so that gives a better chance at explaing the right runtime and usage

advl:
I just committed the necessary changes. The main question is now whether to integrate the routing example into the routing recipe 

> My suggestion is to remove the example `router/select-input` and integrate it into `docs/recipes/routing-with-react-router.md`, so that gives a better chance at explaing the right runtime and usage. What do you think ?

advl:
@sindresorhus What can I do to facilitate the merge ?

sindresorhus:
I needs a review from @vadimdemedes.

PR #412, opened 2021-01-25
Add `Fullscreen` component that tracks screen resizes

prozacgod:
This is a bare-minimum component for the task listed in 

https://github.com/vadimdemedes/goodness-squad/issues/5

### DIFF
--- a/src/components/Fullscreen.tsx
+++ b/src/components/Fullscreen.tsx
@@ -16,10 +16,10 @@ const Fullscreen: FC = ({children}) => {
 		}
 
 		process.stdout.on('resize', onResize);
-		process.stdout.write('\x1b[?1049h');
+		process.stdout.write('\u001B[?1049h');
 		return () => {
 			process.stdout.off('resize', onResize);
-			process.stdout.write('\x1b[?1049l');
+			process.stdout.write('\u001B[?1049l');
 		};
 	}, []);
 
--- a/src/components/Fullscreen.tsx
+++ b/src/components/Fullscreen.tsx
@@ -8,12 +8,12 @@ const Fullscreen: FC = ({children}) => {
 	});
 
 	useEffect(() => {
-		function onResize() {
+		const onResize = () => {
 			setSize({
 				columns: process.stdout.columns,
 				rows: process.stdout.rows
 			});
-		}
+		};
 
 		process.stdout.on('resize', onResize);
 		return () => {
--- a/src/components/Fullscreen.tsx
+++ b/src/components/Fullscreen.tsx
@@ -16,10 +16,8 @@ const Fullscreen: FC = ({children}) => {
 		}
 
 		process.stdout.on('resize', onResize);
-		process.stdout.write('\u001B[?1049h');
 		return () => {
 			process.stdout.off('resize', onResize);
-			process.stdout.write('\u001B[?1049l');
 		};
 	}, []);
 
--- a/src/components/Fullscreen.tsx
+++ b/src/components/Fullscreen.tsx
@@ -1,0 +1,33 @@
+import React, {FC, useState, useEffect} from 'react';
+import Box from './Box';
+
+const Fullscreen: FC = ({children}) => {
+	const [size, setSize] = useState({
+		columns: process.stdout.columns,
+		rows: process.stdout.rows
+	});
+
+	useEffect(() => {
+		function onResize() {
+			setSize({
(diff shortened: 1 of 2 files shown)


src/components/Fullscreen.tsx
```diff
@@ -0,0 +1,33 @@
+import React, {FC, useState, useEffect} from 'react';
+import Box from './Box';
+
+const Fullscreen: FC = ({children}) => {
+	const [size, setSize] = useState({
+		columns: process.stdout.columns,
```
sindresorhus, line 6: You are assuming `process.stdout`, but that could be changed in the options: https://github.com/vadimdemedes/ink#stdout-1

src/components/Fullscreen.tsx
```diff
@@ -0,0 +1,33 @@
...
+	useEffect(() => {
+		function onResize() {
+			setSize({
+				columns: process.stdout.columns,
+				rows: process.stdout.rows
+			});
+		}
+
+		process.stdout.on('resize', onResize);
+		process.stdout.write('\u001B[?1049h');
```
sindresorhus, line 19: This should use the `ansi-escapes` package. If these escapes don't exist there, they should be added.

src/components/Fullscreen.tsx
```diff
@@ -0,0 +1,33 @@
...
+import Box from './Box';
+
+const Fullscreen: FC = ({children}) => {
+	const [size, setSize] = useState({
+		columns: process.stdout.columns,
+		rows: process.stdout.rows
+	});
+
+	useEffect(() => {
+		function onResize() {
```
sindresorhus, line 11: Use arrow function.

sindresorhus (changes requested):

src/components/Fullscreen.tsx
prozacgod: Is this for techincal reasons or just the project's coding standard? (just curious)

src/components/Fullscreen.tsx
prozacgod: Good call, I noticed that should have sorted it - is there a mechanism by which the terminal is already captured in a context or anything?  If so, this seems like it may be a deeper dive than this "bare minimum".  If you have any tips, please!!! :D  If not, I'll look into it over the weekend.

prozacgod: .. Okay I was too curious, so I walked through the renderer code there's a StdinContext, StdoutContext I can look at.

src/components/Fullscreen.tsx
sindresorhus: Just a preference. Arrow functions are better as they don't get hoisted in the scope. Not a problem here, but in general better to be consistent.

prozacgod:
I remove the escape sequences as they aren't necessary, and I didn't want to have to sort out the "right way" as I'm not an ANSI expert, those were the 'alt-buffer' xterm sequence enter/return - while "nice to have" in some cases, are not at all necessary and likely better handled by another component or the implementor themselves.

I started to look into the stdout context,I'm not 100% of the path to take, as the generic "write stream" has optional rows/columns - so I'll have to do some testing on alternatives streams - maybe I'll make a socket server and telnet into it and see how it all behaves.

vadimdemedes:
Thanks for working on this! I think this functionality would be better off as a React hook than a component, more flexible that way. It could be called `useViewportSize` or `useTerminalSize`. Which name do you think makes more sense?

prozacgod:
Oh, I really like that idea, would love to see it - mm as far as function name, doesn't matter to me - as a tongue-in-cheek test I just tried typing both functions - useTerminalSize seems faster/easier to type :P  (totally objective test there)

dustinlacewell:
Really hope to see this.

vadimdemedes:
Hey @prozacgod, do you still have time to finish this PR? Totally cool if not ;)

prozacgod:
I could put some times into this, although I'm a bit... out of touch with the original task ... I think there was a request to do a useTerminalSize as a hook, but just glancing at this out of context I'm not sure how this would  look in practice... so I'll mock something up here and give me some feedback on this..

This is kinda what you were thinking of the use, if we just drop it and make it a hook?

`
  const example = () => {
    const size = useTerminalSize();

    return (
      <Box width={size.columns} height={size.rows}>
        {children}
      </Box>
    );
}
`

prozacgod:
I have fully abandoned this.

Issue #359, opened 2020-08-08
what to do when a view gets longer than the screen as it flickers badly on updates


KyleAMathews:
![ink-flickering](https://user-images.githubusercontent.com/71047/89714383-f63d9c80-d952-11ea-843b-4268be14cfe2.gif)

Is there any easy way to detect when this is happening so we can take corrective measures?

vadimdemedes:
Hm, not really, this is up to your terminal to render the text, so the flickering is on them. For example, while iTerm may flicker, built-in terminal app in macOS almost never does.

As for the code here, what is the UI supposed to do? I will probably be able to help better if I'm more familiar with your use case.

H3RSKO:
We noticed in our app (gitgud) that iTerm2 would cause it to become almost unusable due to flickering. If you switch terminals to hyper or the default mac terminal it will almost completely fix the issue. I think it is just an iTerm2 bug.

mAAdhaTTah:
FWIW, I feel like lately iTerm2 has been having issues like this that are definitely unrelated to Ink. Just streaming significant of output in general has been weird. Maybe the most recent update had a bug or something?

marcus13371337:
Hello,

I'm having real issues running my application in "fullscreen" on iTerm. The screen is flickering really badly when there are a lot of updates to the screen. It seems as iTerm queues up all the frames and shows them one by one instead of just showing the latest frame.

My application is a circle-ci dashboard. I was inspired by k9s (https://github.com/derailed/k9s) which runs really smoothly on iTerm (written in go though). Have a look at my project to see if you can find any gotchas:
https://github.com/marcus13371337/circle-dash

H3RSKO:
It seems to be an iTerm issue. Try using a different terminal.

marcus13371337:
Yes, it's definitely an iTerm issue, however, if I wanted to distribute my cli I want it to run smoothly in every terminal. Since there are applications running smoothly on iTerm (k9s as an example), I guess there are some hacks/features ink can implement in order to make the iTerm experience better

marcus13371337:
Think I found a solution to the problem which I've presented in the following PR:
https://github.com/vadimdemedes/ink/pull/413

@H3RSKO Would you be super kind to try it with gitgud in iTerm and confirm that it fixes the flickering, would be super nice. It's not 100% bug-free though

pie6k:
[...]

PR #413, opened 2021-01-26
POC: Incremental rendering

marcus13371337:
Did some digging on the following issue:
https://github.com/vadimdemedes/ink/issues/359

Found that what's causing the flickering on iTerm2 is that the whole terminal window is erased and then rewritten. IMO iTerm does the right thing, which is to show an empty terminal and then add back the content -> boom we have a flickering. Another approach for updating the terminal would be what I have in this POC PR, which is to update only the rows that have been changed. Tested in my cli, but with some bugs such that React console messages are displayed in the top and stuck there forever. This is the same approach as other terminal packages seem to do as well.

Needs some improvements, of course, just a POC!

### DIFF
--- a/src/ink.tsx
+++ b/src/ink.tsx
@@ -1,7 +1,6 @@
 import React, {ReactNode} from 'react';
 import {throttle, DebouncedFunc} from 'lodash';
 import logUpdate, {LogUpdate} from './log-update';
-import ansiEscapes from 'ansi-escapes';
 import originalIsCI from 'is-ci';
 import autoBind from 'auto-bind';
 import reconciler from './reconciler';
@@ -60,7 +59,7 @@ export default class Ink {
 		this.throttledLog = options.debug
 			? this.log
 			: throttle(this.log, undefined, {
-					leading: true,
+					leading: false,
 					trailing: true
 			  });
 
@@ -111,7 +110,7 @@ export default class Ink {
 			return;
 		}
 
-		const {output, outputHeight, staticOutput} = render(
+		const {output, staticOutput} = render(
 			this.rootNode,
 			// The 'columns' property can be undefined or 0 when not using a TTY.
 			// In that case we fall back to 80.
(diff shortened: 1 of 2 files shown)


src/ink.tsx
```diff
@@ -12,6 +11,7 @@ import * as dom from './dom';
 import {FiberRoot} from 'react-reconciler';
 import instances from './instances';
 import App from './components/App';
+import ansiEscapes from 'ansi-escapes';
```
vadimdemedes, line 14: Extra change.

src/ink.tsx
```diff
@@ -60,7 +60,7 @@ export default class Ink {
 		this.throttledLog = options.debug
 			? this.log
 			: throttle(this.log, undefined, {
-					leading: true,
+					leading: false,
```
vadimdemedes, line 63: Why change this?

src/ink.tsx
```diff
@@ -143,7 +143,7 @@ export default class Ink {
 			this.fullStaticOutput += staticOutput;
 		}
 
-		if (outputHeight >= this.options.stdout.rows) {
+		if (outputHeight > this.options.stdout.rows) {
```
vadimdemedes, line 146: Why change this?

src/log-update.ts
```diff
@@ -24,9 +24,22 @@ const create = (stream: Writable, {showCursor = false} = {}): LogUpdate => {
 			return;
 		}
 
-		previousOutput = output;
-		stream.write(ansiEscapes.eraseLines(previousLineCount) + output);
-		previousLineCount = output.split('\n').length;
+		const newLines = output.split('\n');
```
vadimdemedes, line 27: I'm a bit confused, doesn't this `updateSequence` variable contains ANSI escapes to erase all the previous lines anyway?

marcus13371337: Thanks for your review! Sorry for a bit of a messy PR, it's just a POC, and wanted to show that the flickering is probably caused by these lines.

Yes it does erase all lines, but it sends the commands to the terminal to do it row by row instead of wipe:ing all lines and then write them again.

marcus13371337: When I initially created the PR I updated only the lines that were changed, but then I tested to replace all lines at it seemed to work equally well.

sindresorhus:
Bump

marcus13371337:
@sindresorhus Hi again, sorry haven't had enough time to fix this PR to a proper state. However, the POC has been made IMO, that instead of erasing the whole window and then repaint it is what's causing the flickering. Updating one row at the time seems to fix that. Could someone pick it up from there?

PhamMinhHaiAu-12035071 (approved):

PR #422, opened 2021-03-21
docs: Alphabatical arrangement of libs using ink

pranshuchittora:


sindresorhus:
Thanks for the suggestion, but it's too much work to ensure contributors follow it alphabetically. I have tried it multiple times in the past on more popular projects and it always ended up me having to edit every single pull request.

PR #424, opened 2021-03-25
Export stdout/stderr writers from render.

milesj:
@vadimdemedes 

I've run into an instance where I need to write to `process.stdout/stderr` directly, but I don't want to break Ink, and it's currently not possible outside of using `console` (and its patching).

This is a simple change that simply re-exports the stdout/err writers from the `Ink` instance.

### DIFF
--- a/src/render.ts
+++ b/src/render.ts
@@ -55,8 +55,18 @@ export interface Instance {
 	 * Returns a promise, which resolves when app is unmounted.
 	 */
 	waitUntilExit: Ink['waitUntilExit'];
+	/**
+	 * Write to stdout without disrupting the render cycle.
+	 */
+	writeToStdout: Ink['writeToStdout'];
+	/**
+	 * Write to stderr without disrupting the render cycle.
+	 */
+	writeToStderr: Ink['writeToStderr'];
+	/**
+	 * Delete this instance.
+	 */
 	cleanup: () => void;
-
 	/**
 	 * Clear output.
 	 */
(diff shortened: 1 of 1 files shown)


vadimdemedes:
Could you share the use case when it's necessary? Curious when people would need this, since it replaces Ink's output just for one frame.

milesj:
@vadimdemedes It's been a while since I posted this but I think this was the problem.

Currently, I need access to stdout/stderr, and the only way to do that is through the React layer where I bubble the streams back up, as seen here: https://github.com/milesj/boost/blob/master/packages/cli/src/components/internal/LogWriter.ts#L10

I basically want to avoid waiting for React render to access these, and just access them directly off of `render()` here:  https://github.com/milesj/boost/blob/master/packages/cli/src/Program.ts#L284

vadimdemedes:
Thanks for the reply despite the old thread!

I see that you have `patchConsole: true` in your Ink options (https://github.com/milesj/boost/blob/master/packages/cli/src/Program.ts#L301). Is there a reason why you can't use `console.log` or `console.error`?

milesj:
I think it was to avoid the patching overhead and just write to the stream directly. I wanted it as fast as possible.

vadimdemedes:
The overhead of patching is pretty minimal, see https://github.com/vadimdemedes/patch-console/blob/master/source/index.ts.

Going to close this for now, since it's a pretty narrow use case and I'm not sure I'd want to expose additional APIs for something that Ink supports already.

Issue #426, opened 2021-04-05
Add programmatic focus option


rektdeckard:
Excellent lib! Was thinking it would be nice to be able to programmatically focus a *specific* element, instead of being at the whims of the layout order. A simple, non-breaking API change could solve this quite nicely, e.g.:

```tsx
const First = () => {
  // Pass an optional 'id' to useFocus
  const { isFocused } = useFocus({ id: "first" });
  return <Text>{isFocused ? "I am focused" : "I am not focused"}</Text>;
}

const Second = () => {
  // Return a 'focus' function that focuses the element with id matching the argument
  const { isFocused, focus } = useFocus();

  useInput((input) => {
      if (input === "f") focus("first");
    }, { isActive: isFocused }
  );

  return <Text>{isFocused ? "I am focused" : "I am not focused"}</Text>;
}
```

It looks like the machinery is already there, just not exposed in the API. WDYT?

PR #427, opened 2021-04-05
Add programmatic focus (#426)

rektdeckard:
This patch adds the ability to explicitly assign a focus `id` to an element, as well as to focus any element directly by calling the `focus(id: string) => void` function returned by both `useFocus` and `useFocusManager` hooks. #426.

```tsx
const First = () => {
  // Pass an optional 'id' to useFocus
  const { isFocused } = useFocus({ id: "first" });
  return <Text>{isFocused ? "I am focused" : "I am not focused"}</Text>;
}

const Second = () => {
  // Returns a 'focus' function that focuses the element with id matching the argument
  const { isFocused, focus } = useFocus();

  useInput((input) => {
      // Focus any element directly without having to cycle through focusables
      if (input === "f") focus("first");
    }, { isActive: isFocused }
  );

  return <Text>{isFocused ? "I am focused" : "I am not focused"}</Text>;
}
```


src/components/App.tsx
```diff
@@ -213,6 +214,17 @@ export default class App extends PureComponent<Props, State> {
 		});
 	};
 
+	focus = (id: string): void => {
+		this.setState(previousState => {
+			const elementWithIdOrNext =
+				previousState.focusables.find(e => e?.id === id)?.id ||
```
sindresorhus, line 220:
```suggestion
				previousState.focusables.find(event => event?.id === id)?.id ||
```

src/hooks/use-focus-manager.ts
```diff
@@ -25,6 +25,12 @@ interface Output {
 	 * If active component is the first in the list of focusable components, focus will be switched to the last component.
 	 */
 	focusPrevious: Props['focusPrevious'];
+
+	/**
+	 * Switch focus to the element with provided `id`.
+	 * If there's no element with thad `id`, focus will be given to the first focusable component.
```
sindresorhus, line 31: Typo

src/hooks/use-focus.ts
```diff
@@ -12,13 +12,23 @@ interface Input {
 	 * Auto focus this component, if there's no active (focused) component right now.
 	 */
 	autoFocus?: boolean;
+
+	/**
+	 * Assign an ID to this component, so it can be programmatically focused with `focus(id)`.
```
sindresorhus, line 17:
```suggestion
	 * Assign an ID to this component, so it can be programmatically focused with `.focus(id)`.
```

sindresorhus (changes requested):

src/hooks/use-focus-manager.ts
rektdeckard: Lol, whose Thad?

src/hooks/use-focus.ts
rektdeckard: It's not an instance method. This is the callback returned from `useFocus` and `useFocusManager`.

src/components/App.tsx
rektdeckard: Not an event here, it's a `Focusable`. Should I call is focusable?

sindresorhus: Ah, yes.

rektdeckard:
Anything left to do here or can this be merged?

sindresorhus:
It needs a review from @vadimdemedes.

rektdeckard:
@vadimdemedes Any chance of reviewing? 🙏  Would love to use it in a project I'm working on.

dustinlacewell:
Just want to report that this feature is working pretty well for me. Thanks @rektdeckard

src/hooks/use-focus.ts
```diff
@@ -29,29 +39,35 @@ interface Output {
...
+	id
+}: Input = {}): Output => {
 	const {isRawModeSupported, setRawMode} = useStdin();
-	const {activeId, add, remove, activate, deactivate} = useContext(
+	const {activeId, add, remove, activate, deactivate, focus} = useContext(
 		FocusContext
 	);
 
-	const id = useMemo(() => Math.random().toString().slice(2, 7), []);
+	const resolvedId = useMemo(() => id ?? Math.random().toString().slice(2, 7), [
```
vadimdemedes, line 52: Could you keep this variable name named as `id`, and rename `id` in arguments to `customId` as such?

```js
const useFocus = ({ id: customId }) => ...
```

src/components/App.tsx
```diff
@@ -213,6 +214,17 @@ export default class App extends PureComponent<Props, State> {
 		});
 	};
 
+	focus = (id: string): void => {
+		this.setState(previousState => {
+			const targetIdOrNext =
+				previousState.focusables.find(focusable => focusable.id === id)?.id ||
+				this.findNextFocusable(previousState);
```
vadimdemedes, line 221: Curious about this behavior, why do you think it should focus the next element if no element by ID is found? I'd think it would do nothing.

rektdeckard: Fair point, probably the expected behavior! I'll remove.

src/hooks/use-focus.ts
rektdeckard: Done!

src/components/App.tsx
rektdeckard: Done

PR #431, opened 2021-04-16
Control individual borders & custom border styles

kosciolek:
The following features are added:
* Control individual borders
* Provide custom border styles

Looks like this:

```ts
const Example = () => (
  <Box flexDirection={"column"}>
    <Box>
      <Box
        width={50}
        borderStyle={{
          bottomLeft: "+",
          bottomRight: "+",
          horizontal: "-",
          topLeft: "+",
          topRight: "+",
          vertical: "|"
        }}
        borderLeft={true}
      >
        <Text>I have a custom border.</Text>
      </Box>
      <Box borderStyle={"bold"} width={50} borderLeft={false}>
        <Text>I have no left border.</Text>
      </Box>
    </Box>
    <Box borderStyle="bold" borderTop={false}>
      <Text>I have no top border.</Text>
    </Box>
  </Box>
);
```

### DIFF
--- a/src/components/Box.tsx
+++ b/src/components/Box.tsx
@@ -79,11 +79,11 @@ Box.defaultProps = {
 	flexDirection: 'row',
 	flexGrow: 0,
 	flexShrink: 1,
-	
+
 	borderRight: true,
-	borderLeft:true,
-	borderBottom:true,
-	borderTop:true
+	borderLeft: true,
+	borderBottom: true,
+	borderTop: true
 };
 
 export default Box;
--- a/src/components/Box.tsx
+++ b/src/components/Box.tsx
@@ -78,7 +78,12 @@ Box.displayName = 'Box';
 Box.defaultProps = {
 	flexDirection: 'row',
 	flexGrow: 0,
-	flexShrink: 1
+	flexShrink: 1,
+	
+	borderRight: true,
+	borderLeft:true,
+	borderBottom:true,
+	borderTop:true
 };
 
 export default Box;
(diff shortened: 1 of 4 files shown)


ryan-waltz:
Tell me if it suits you so far, then I'll write the tests.

lodenrogue (approved):

lodenrogue:
Looks cool!

vadimdemedes:
I like the idea, but I'm not a fan of this "inverted" API where all borders are on by default and you need to disable individual sides instead of enabling borders you need. If I want to show only top border, I'd have to set `borderLeft={false} borderRight={false} borderBottom={false}`, which doesn't look good.

I can now see that relying only on `borderStyle` being present to show a border was a mistake on my end. Would love to find a way to ship this without introducing breaking changes.

lordofthelake:
Pitching in, as it is something that I'm also in need of. 

Maybe a simple solution to keep retro compatibility (at least in the 3.x series) would be to just render the borders that are specified in the `borderStyle` property, and to split the `horizontal` and `vertical` options into `top`/`bottom`/`left`/`right`?
```
<Box
        width={50}
        borderStyle={{
          bottomLeft: "+",
          bottomRight: "+",
          top: "-",
          bottom: "-",
          topLeft: "+",
          topRight: "+",
          left: "|" // rendering only the left border
        }}
      >
```

You lose the ability to use the presets and still render some of the borders (that could be reintroduced later, in a future 4.x version, with an overhauled border API), but it would not introduce any breaking change and still allow for the same flexibility (at the price of having to redefine all the border symbols if you want to use one of the predefined styles, but skipping some of the sides)

vadimdemedes:
Submitted https://github.com/vadimdemedes/ink/pull/568, which builds upon this PR to support showing/hiding individual border sides. I'll follow up with another PR that supports custom border styles as well.

Thank you @kosciolek for this contribution!

PR #436, opened 2021-05-06
Add link to ink-task-list in readme

privatenumber:
Added https://github.com/privatenumber/ink-task-list to the list


sindresorhus:
I don't think "task runner" is the correct description of this. It's a "task list".

privatenumber:
One of the exported components is called TaskList because it simply wraps a list of tasks. But collectively, I thought it's closer to the View of a task runner because it has dynamic states (eg. loading state).

Also, "task list" sounds synonymous to "todo list", and sounds like a static checklist.

sindresorhus:
But "task runner" makes it sound like it **runs** tasks, which it does not. It displays the status of tasks.

privatenumber:
I was hoping the `ink-` prefix would imply the "View" part, but I see your point that it's clearer to name it task list than runner.

I updated the package name 👍

---

P.S. While I have you here, just also want to say thank you for all the work that you do. ink-task-list would not have been so easy to make without your projects ❤️ I am constantly impressed to find you in so many corners of the ecosystem!

PR #445, opened 2021-05-23
Fix meta key

RichieAHB:
Addresses: https://github.com/vadimdemedes/goodness-squad/issues/3 - and removes `meta: true` from arrow / page events. In so doing this migrates the `useInput` hook to a potentially more sustainable keypress implementation, in line with Node's own implementation.

There are a few questions here like:
- Should the `meta` key input data be capitalised of always be lowercase?
- Should the `delete` input for `\u007f` actually be `backspace` in line with the Node implementation?


src/hooks/use-input.ts
```diff
@@ -132,49 +147,38 @@ const useInput = (inputHandler: Handler, options: Options = {}) => {
 			return;
 		}
 
-		const handleData = (data: Buffer) => {
-			let input = String(data);
+		const handleKeyPress = (data: string, keyData: ReadlineKeyInfo) => {
```
sindresorhus, line 150:
```suggestion
		const handleKeyPress = (data: string, keyInfo: ReadlineKeyInfo) => {
```

src/hooks/use-input.ts
```diff
@@ -78,6 +79,16 @@ export interface Key {
 
 type Handler = (input: string, key: Key) => void;
 
+// Taken from https://nodejs.org/api/readline.html#readline_rl_write_data_key
+interface ReadlineKeyInfo {
```
sindresorhus, line 83: Can you not import it from the Node.js types?

sindresorhus (changes requested):

sindresorhus:
> There are a few questions here like:

It would be great if you could share your opinion about the questions.

src/hooks/use-input.ts
RichieAHB: I assumed they'd exist aroudn the `keypress` event but they were in `readline` and I missed them, so will updated 👍

RichieAHB:
> It would be great if you could share your opinion about the questions.

In terms of the naming of `backspace` or `delete` for `\u007f`. Everywhere I looked seemed to have it down as `delete` (and that's what it is on Wikipedia: https://en.wikipedia.org/wiki/Delete_character) but as mentioned on Wikipedia (and as happens on my Mac), it actually appears when the backspace key is pressed. If I use `Fn-Backspace` (i.e delete forward), I see the `\x1B[3~` escape sequence which node maps to `delete`. Personally I think I'd prefer to have this mapped to `backspace` but it would be a breaking change.

As for the `meta` key handling capitalisation, I don't know of many use cases for it being important (although there may be) and I'd be inclined to leave it like this without knowing better!

RichieAHB:
There _is_ actually a problem with this method in that a sequence such as `Ctrl-h` gets sent to the terminal as `backspace` keypress. Whereas in the previous implementation the sequences were mapped to their _actual_ inputs with:

https://github.com/vadimdemedes/ink/blob/5abb6a12ff4fbf26532977cc064bf4a44f34094a/src/hooks/use-input.ts#L156-L161

One issue with the previous approach is that (at least on a Mac), the meta key (mapped to Esc) will always be sent as an escape sequence on its own as it doesn't seem to be considered a control key in that sense. Whereas with the `keypress` event the meta key can prefix other chars in one event, which is what this code expects (but will never see on my Mac):

https://github.com/vadimdemedes/ink/blob/5abb6a12ff4fbf26532977cc064bf4a44f34094a/src/hooks/use-input.ts#L163-L166

Perhaps a combination of these two approaches may be more effective. Will downgrade this PR to a draft.

vadimdemedes:
Hey @RichieAHB, I'm going to merge a [simpler solution](https://github.com/vadimdemedes/ink/pull/573) for this bug for now, but I'll probably use your code for the next major release of Ink. Thank you for this contribution!

RichieAHB:
No worries - I remember at the time not being 100% sure it worked so use it if it makes sense but thanks for continuing to work on it!

Issue #428, opened 2021-04-13
Using backgroundColor breaks ink


Twiggeh:
I get the error 
```
/mnt/Storage/projects/Webdev/CodeGeneration/node_modules/yoga-layout-prebuilt/yoga-layout/build/Release/nbind.js:53
        throw ex;
        ^

TypeError: chalk_1.default[method] is not a function
```
 with writing

```jsx
<Text backgroundColor='bgBlackBright'> 
```
Any of the suggested strings breaks with the same error, not jut `bgBlackBright`

<details>
<summary>Dependencies & Versions</summary>

```json
"@types/node": "^14.14.35",
"@types/react": "^17.0.3",
"ink": "^3.0.8",
"react": "^17.0.2",
"typescript": "^4.2.3"
```
</details>

<details>
<summary> Stack Trace </summary>

```
TypeError: chalk_1.default[method] is not a function
    at Object.exports.default (/mnt/Storage/projects/Webdev/CodeGeneration/node_modules/ink/build/colorize.js:21:39)
    at transform (/mnt/Storage/projects/Webdev/CodeGeneration/node_modules/ink/build/components/Text.js:24:42)
    at Output.get (/mnt/Storage/projects/Webdev/CodeGeneration/node_modules/ink/build/output.js:40:28)
    at Object.exports.default (/mnt/Storage/projects/Webdev/CodeGeneration/node_modules/ink/build/renderer.js:29:74)
    at Object.Ink.onRender [as onImmediateRender] (/mnt/Storage/projects/Webdev/CodeGeneration/node_modules/ink/build/ink.js:49:78)
    at invokeFunc (/mnt/Storage/projects/Webdev/CodeGeneration/node_modules/lodash/lodash.js:10401:23)
    at leadingEdge (/mnt/Storage/projects/Webdev/CodeGeneration/node_modules/lodash/lodash.js:10411:26)
    at Object.debounced [as onRender] (/mnt/Storage/projects/Webdev/CodeGeneration/node_modules/lodash/lodash.js:10478:20)
    at resetAfterCommit (/mnt/Storage/projects/Webdev/CodeGeneration/node_modules/ink/build/reconciler.js:42:22)
    at commitRootImpl (/mnt/Storage/projects/Webdev/CodeGeneration/node_modules/react-reconciler/cjs/react-reconciler.development.js:15066:5)
    ```
</details>

sindresorhus:
[...]

PR #451, opened 2021-06-27
Fix the TypeScript type for the `backgroundColor` prop

hmu332233:
Hello, 
I read the issue(#428) and would like to suggest a solution to this

114594138-334df180-9c8d-11eb-8150-aef2cb6a6f96

Chalk's backgroundColor type is declared as follows.
https://github.com/chalk/chalk/blob/main/index.d.ts#L31-L49
```js
export type BackgroundColor =
	| 'bgBlack'
	| 'bgRed'
	| 'bgGreen'
	| 'bgYellow'
	| 'bgBlue'
	| 'bgMagenta'
	| 'bgCyan'
	| 'bgWhite'
	| 'bgGray'
	| 'bgGrey'
	| 'bgBlackBright'
	| 'bgRedBright'
	| 'bgGreenBright'
	| 'bgYellowBright'
	| 'bgBlueBright'
	| 'bgMagentaBright'
	| 'bgCyanBright'
	| 'bgWhiteBright';
```

However, in ink, there is a code to attach bg prefix (https://github.com/vadimdemedes/ink/blob/master/src/colorize.ts#L8-L14),
so there is no need to attach bg.
Therefore, I suggest to use chalk's `ForegroundColor` type as auto-complete.

### DIFF
--- a/src/components/Text.tsx
+++ b/src/components/Text.tsx
@@ -1,5 +1,5 @@
 import React, {FC, ReactNode} from 'react';
-import chalk, {ForegroundColor, BackgroundColor} from 'chalk';
+import chalk, {ForegroundColor} from 'chalk';
 import colorize from '../colorize';
 import {Styles} from '../styles';
 import {LiteralUnion} from 'type-fest';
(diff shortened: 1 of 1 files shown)


vadimdemedes:
Thank you!

PR #456, opened 2021-07-19
wip: add title option to a border

ribelo:
No test and no extensive code, because I don't know if there's even a chance of merging such a feature. If there is, I can add the needed stuff.


vadimdemedes:
Looks like an interesting feature, but I wonder if it's possible to solve this with a user-land component instead?

PR #462, opened 2021-07-24
Remove unsupported color methods

cometkim:
`hsl`, `hsv`, `hwb`, `keyword`, and `ansi` seem to have been removed on chalk

See https://github.com/chalk/chalk/pull/433


sindresorhus:
You need to update the tests using these color types too.

PR #475, opened 2021-09-06
Add support for controlling borders individually in `<Box>`

### LINKED ISSUE #425: Control which side has a border
So far, it looks like it's possible to draw only a full box, where borders on all 4 sides are visible.

Is it possible to control which sides are rendered? If not, will it be in the future, by any chance?

lordofthelake:
This is a re-implementation of #431, addressing @vadimdemedes' [comment](https://github.com/vadimdemedes/ink/pull/431#issuecomment-882620548) about keeping API retro-compatibility, while adding some extra flexibility:

* Allows controlling which character is used to draw the border box individually for each side (adding the possibility of using different characters for left/right and top/bottom borders)
* Allows the user to pass only some keys in the `borderStyles` object to specify which sides are rendered, instead of adding additional boolean flags.

It comes already with tests and updated documentation.

Closes #425, #431


lordofthelake:
@vadimdemedes I fixed the build issue, now everything should pass (if the workflow gets approved)

lordofthelake:
@sindresorhus @vadimdemedes Updated again to incorporate #478, to fix the linter error that stopped the build. Please note that at the moment, on `master`, the test for "useInput - handle Ctrl+C when `exitOnCtrlC` is `false`" is flakey — can be disabled or a bit of delay can be added to give time to the sub-process buffer to be flushed.

vadimdemedes:
Shipped a slightly different approach for this in https://github.com/vadimdemedes/ink/pull/568 and https://github.com/vadimdemedes/ink/pull/570. Thank you!

PR #478, opened 2021-10-02
Fix CI

jodevsa:
it looks like **eslint** behaviour have changed lately and we are getting **react/default-props-match-prop-types** errors.
I would have fixed the lint errors, but turns out we need to change some boolean props names from **name** to **isName** which would be a breaking change. I decided to turn it off as we also have **react/prop-types** turned off.


jodevsa:
@vadimdemedes

PR #481, opened 2021-10-07
Make shell commands copyable in the readme

LitoMore:
### Changes

- Remove `$` prefix from shell commands
- Specify the language for code block

### DIFF
--- a/migrate.md
+++ b/migrate.md
@@ -5,8 +5,8 @@ If you're migrating your app and running into issues that this guide doesn't cov
 
 ## Install
 
-```
-$ npm install ink@next react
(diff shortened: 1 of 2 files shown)


LitoMore:
> I'd prefer to leave it there, I like how it makes it clear that it's a shell command.

Since GitHub launched the `copy-text` feature for code blocks, no `$` will more convenient. Users can copy-paste and run.

> Doesn't seem like `sh` language adds any highlighting?

Yes, it's not highlighted. But just for now. Even if it doesn't have any highlighting, we can still specify a language for it. Just in case.

PR #484, opened 2021-10-09
Add `truncationCharacter` prop for `Text` component

### LINKED ISSUE #480: Truncate without …
I'm trying to create an app with ink.js and I need to truncate a string, but without `…`.

I looked through the code of the `<Text>` component, but it just passes the `wrap` property on to an underlying `<ink-text>` component which AFAIK puts altering the way truncation works out of my reach.

I tried limiting the size of the box but leaving `wrap` set to its default value `"wrap"`, but that resulted in the text wrapping out of its container and overwriting components that came after it – not exactly ideal.

Is there some way to truncate text without adding the `…` character?

codemaster138:
This PR adds a `truncationCharacter` prop to the `Text` component:

```jsx
<Box width={7}>
	<Text wrap="truncate" truncationCharacter=".">Hello World</Text>
</Box>
//=> 'Hello.'

<Box width={7}>
	<Text wrap="truncate" truncationCharacter="+++">Hello World</Text>
</Box>
//=> 'Hel+++'
```
---
Resolves #480


codemaster138:
@vadimdemedes Sorry to ping. I wrote tests, but I was unable to actually verify them because *literally every single test in your test suite* seems to fail for some reason.

sindresorhus:
> I wrote tests, but I was unable to actually verify them because literally every single test in your test suite seems to fail for some reason.

Tests are passing on CI, so this sounds like a problem with your setup.

vadimdemedes:
I'm a little concerned how this will affect layout calculation, since `truncationCharacter` can be longer than one character. I thought the initial feature request you've had was to simply truncate without the three dots at the end?

PR #486, opened 2021-10-22
README - add a live demo for ink-big-text

hangtwenty:
Added a quick demo of react ink with ink-big-text. Forked the existing Counter demo ( https://replit.com/@vadimdemedes/ink-counter-demo ) - it displays the counter value in big text.


vadimdemedes:
We already have one counter example, but feel free to create something different and I'll happily add it to the readme ;)

PR #498, opened 2022-02-10
Move `react-devtools-core` to optional peer dependency

### LINKED ISSUE #491: react-devtools-core adds 14 MB to every installation using Ink
Hi @vadimdemedes,

Thank you so much for such an incredible package!

We've noticed that ink adds a lot of weight to the tool we want to use it for, and upon digging further into it, well over 80% of that weight (**14 MB**) comes exclusively from `react-devtools-core`, which is obviously completely unused by end users of tools which leverage ink at runtime.

Would you accept a PR which removes the direct dependency on `react-devtools-core` in favour of throwing a clear error if a developer tries to use the devtools without having first installed them? I think this trade off in favour of end users is well worth it.

Thanks again!

SimenB:
Fixes #491.

Should possibly wrap https://github.com/vadimdemedes/ink/blob/763db800f6b93c0807b016c6dc1ac8cccef0a67b/src/reconciler.ts#L24-L30 into a more specific error as well?

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,6 @@
 		"is-ci": "^2.0.0",
 		"lodash": "^4.17.20",
 		"patch-console": "^1.0.0",
-		"react-devtools-core": "^4.19.1",
 		"react-reconciler": "^0.26.2",
 		"scheduler": "^0.20.2",
 		"signal-exit": "^3.0.2",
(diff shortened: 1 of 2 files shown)


vadimdemedes:
> Should possibly wrap into a more specific error as well?

You mean something like this?

```js
try {
	require('react-devtools-core');
} catch {
	throw new Error('React Devtools requires `react-devtools-core` dependency to be installed.');
}
```

vadimdemedes:
Yep, `DEV` environment variable is only used for devtools. However, I like the idea of just having `console.warn`, because technically nothing is broken, so there's no reason to crash the whole app.

Let's go with `console.warn` 👍 Could you extend the error message to:

```
Debugging with React Devtools requires `react-devtools-core` dependency to be installed.

$ npm install --save-dev react-devtools-core

```

SimenB:
image

Seems fine?

vadimdemedes:
Yes, thank you!

PR #504, opened 2022-02-26
fix: disable `esModuleInterop`

SimenB:
Using it in `ink` forces me to use it downstream


vadimdemedes:
Disabled in https://github.com/vadimdemedes/ink/commit/71770f8f94c93bd129b25b1cedc84ca17fb0d49e.

PR #511, opened 2022-04-06
Import `throttle` lodash function directly

merceyz:
Updates the `lodash` import to only include the used function which improves startup time and reduces the size of Ink when bundled.

Ref https://github.com/yarnpkg/berry/pull/4315


vadimdemedes:
Per-method Lodash packages are deprecated → https://lodash.com/per-method-packages. I think we should use another npm package for this.

merceyz:
I'm aware of that, however this PR doesn't use those packages.

sindresorhus:
@vadimdemedes I think this pull request makes sense. Importing sub-exports of Lodash does improve loading time as Lodash is huge.

vadimdemedes:
@sindresorhus It wasn't possible to import `throttle` directly from `lodash-es` like so:

```js
import throttle from 'lodash-es/throttle.js`
```

But it was possible with `lodash`, which seems to still be a CJS module. Would that be a problem for ESM-only Ink?

sindresorhus:
Should be fine and also beneficial as `lodash-es` is also not maintained anymore.

PR #516, opened 2022-04-20
fix: set correct peer dependency range for react

SimenB:
Matches `react-reconciler`: https://www.runpkg.com/?react-reconciler@0.26.2/package.json#29


vadimdemedes:
Closing, since it got updated in #543.

PR #517, opened 2022-05-01
refactor: use lodash.throttle instead of lodash

privatenumber:
Use `lodash.throttle` instead of `lodash`, since only `throttle` is being imported.

In a package I'm using `ink` in, [this reduced the bundle size by 26%](https://github.com/privatenumber/tasuku/pull/19) (or ~70kB minified).


privatenumber:
Just noticed this is a duplicate of https://github.com/vadimdemedes/ink/pull/511

Feel free to pick whichever.

merceyz:
Those packages are deprecated and shouldn't be used. See https://lodash.com/per-method-packages.

PR #527, opened 2022-07-15
declare Item before Focus uses it

Silverium:


vadimdemedes:
Is there an issue without this change?

Silverium:
As far as I know, `const` is not hoisted and this code should blatantly fail. When using `var` or `function` this is not a problem, but if I am not mistaken, this is simply an error in the example, which is not run, therefore it's not failing any test

mAAdhaTTah:
@Silverium The example is valid. This runs as expected as a smaller example:

```js
const hello = () => console.log(world);
const world = 'Hello world!'
hello();
```

The reason is the reference to `world` in the above example isn't actually used until after it's declared & assigned. This would fail with a `ReferenceError` if `hello` was called before `world` was declared & assigned.

Silverium:
meh...
image
https://typescript-eslint.io/rules/no-use-before-define/

Silverium:
I, male person near 40 years old, could go into a friend's wedding dressed exactly as the bride. I would be in perfect consonance with dressing policies... but it's still morally and ethically wrong ;)

PR #532, opened 2022-09-12
Add sql-multi-db-query to "Who's using Ink?"

faridprogrammer:
Hi. 
Thanks for your great package.

I have added my package which is using `ink`.

Hope that helps


sindresorhus:
Please put some more effort into this pull request and also the presentation of your package. It's missing a readme, screenshots, etc. And this PR is not following the format of other entries.

PR #536, opened 2022-10-03
Fix broken layout when emojis are used

### LINKED ISSUE #505: 3.2.0 Emojis render with trailing spaces within Boxes
This again doesn't work. See https://github.com/vadimdemedes/ink/pull/444

Version: 3.2.0
Platform: Windows 10 (vs-code, powershell)

image

lucas-labs:
Fixes #505 

The problem is happening due to (I think) `yoga` and `slice-ansi` incompatibility on string indexes. `slice-ansi` counts emojis as 1 char and `yoga` counts it as 2 (again, I think).

Here I'm using `emoji-regex` (which was already pulled in by other dependencies so it doesn't add any real new package to the dependency tree) to count emojis on the current line. Then I'm substracting that amount to the `x` variable (starting offset on where to replace the string).

If you find a better way to fix this issue please let me know and I'll be happy to try it out and implement it here.

Before: 
image

After: 
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,6 +49,7 @@
 		"cli-cursor": "^4.0.0",
 		"cli-truncate": "^3.1.0",
 		"code-excerpt": "^4.0.0",
+		"emoji-regex": "^10.2.1",
 		"indent-string": "^5.0.0",
 		"is-ci": "^3.0.1",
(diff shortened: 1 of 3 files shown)


vadimdemedes:
@lucas-labs Thank you for a nice fix!

PR #537, opened 2022-10-17
Add Shopify CLI to "Who's Using Ink?"

matteodepalo:
Hi,
We've added `ink` to Shopify CLI in order to display output in the `shopify app dev` command.

Thank you for the great project!


vadimdemedes:
💛

PR #544, opened 2023-01-13
Replace `lodash` import with `lodash.throttle`

BeeeQueue:
The same thing as #511 but using the separated package instead.

Closes #511


sindresorhus:
`lodash.throttle` is not maintained. Last update 7 years ago.

Issue #543, opened 2023-01-06
Migrate Ink to ESM


pepicrft:
Hi 👋🏼 

It's Pedro and Ju (@Arkham) from the team building Shopify's [CLI](https://github.com/shopify/cli). First of all, congrats for the great work done on this library. Ink has become the foundation of our CLI and we couldn't be happier with the choice we made. It's so easy to implement UI components on it.

We've been working on some performance optimizations, and we noticed the loading of Ink blocks the runtime while it loads the CommonJS graph. We'd love for the loading of Ink to happen along with the loading of the CLI's ESM module graph. Unfortunately, we can't make that happen because Ink is implemented in CommonJS with support for ESM interoperability.

We'd be up for doing the migration ourselves (including Ink's dependencies) with guidance from the maintainers of the project if this transition aligns with the direction of the package.

matteodepalo:
@sindresorhus @vadimdemedes any thoughts on this?

vadimdemedes:
Thanks @pepicrft for the massive contribution in #546 ❤️

For context, we're trying to resolve the remaining issues in that PR that currently block it from being merged. Once CI is green, ESM support will be shipped.

PR #546, opened 2023-01-30
Migrate Ink to ESM and update React

pepicrft:
Resolves https://github.com/vadimdemedes/ink/issues/543

I'm migrating this package to ESM to nudge the ecosystem towards embracing the standards.

- [x]  Update Typescript
- [x]  Update `@sindresorhus/tsconfig`
  - [x] Fix some type-checking issues after the update
- [x] Set `exports` attribute in the `package.json`.
- [x] Set `moduleResolution` and `module` attributes in the `tsconfig.json` to `node16`
- [x] Update some dependencies to use their ESM version instead of relying on CJS/ESM interoperability.


package.json
```diff
@@ -112,12 +113,12 @@
 		}
 	},
 	"ava": {
-		"extensions": [
-			"ts",
-			"tsx"
-		],
-		"require": [
-			"ts-node/register/transpile-only"
+		"extensions": {
```
pepicrft, line 116: I'm working on updating the tests to ESM and fixing them.

pepicrft:
@sindresorhus I might need your help to get the tests passing. For some reason when I run the tests they time out without any useful output. My guess is that the `ts-node` node module loader fails or hangs internally and AVA ends up timing out. I'd appreciate if you could give me a hand or pointers with this one.

sindresorhus:
Try this config: https://github.com/sindresorhus/got/blob/0ca0b7f7134f41b45a51370154041cc97c28ca60/tsconfig.json#L17-L21

pepicrft:
That tricked worked 🚀. Thanks a lot @sindresorhus. I'll work on fixing the tests and I'll let you know when the PR is ready for a review.

src/colorize.ts
```diff
@@ -10,7 +10,7 @@ const getMethod = (name: string, type: ColorType): string => {
 		return name;
 	}
 
-	return 'bg' + name[0].toUpperCase() + name.slice(1);
+	return 'bg' + (name[0] as string).toUpperCase() + name.slice(1);
```
vadimdemedes, line 13: Is this a replacement for `name[0]!`?

src/components/App.tsx
```diff
@@ -329,7 +329,7 @@ export default class App extends PureComponent<Props, State> {
 			index++
 		) {
 			if (state.focusables[index]?.isActive) {
-				return state.focusables[index].id;
+				return (state.focusables[index] as Focusable).id;
```
vadimdemedes, line 332: Curious why is type casting needed now, but wasn't needed before?

vadimdemedes:
Thanks for this massive contribution, @pepicrft! Let me know when it's ready to review 👍

vadimdemedes:
You can change CI to only test Node.js 14, 16 and 18, since 10 and 12 aren't maintained anymore. This will allow tests to run in this PR.

benchmark/static/static.tsx
```diff
@@ -1,7 +1,9 @@
+#!/usr/bin/env npx ts-node-esm
```
pepicrft, line 1: I changed the approach here. Instead of using `import-jsx` in an `index.js` file I added a shebang at the top to run the script through `ts-node-esm`, which will use Node loaders to transform the module into standard JS.

test/hooks.tsx
```diff
@@ -1,216 +1,227 @@
...
-		reject(new Error(`Process exited with non-zero exit code: ${code}`));
-	});
-
-	return result;
-};
-
-test('useInput - handle lowercase character', async t => {
-	const ps = term('use-input', ['lowercase']);
-	ps.write('q');
-	await ps.waitForExit();
```
pepicrft, line 56: @sindresorhus / @vadimdemedes I'm having issues with `node-pty`. When this line runs, the Node process blows up with the following stacktrace:

```bash
FATAL ERROR: HandleScope::HandleScope Entering the V8 API without proper locking in place
 1: 0x100eda200 node::Abort() [/Users/pepicrft/.asdf/installs/nodejs/18.12.1/bin/node]
 2: 0x100eda2f8 node::OOMErrorHandler(char const*, bool) [/Users/pepicrft/.asdf/installs/nodejs/18.12.1/bin/node]
 3: 0x101022aa8 v8::HandleScope::HandleScope(v8::Isolate*) [/Users/pepicrft/.asdf/installs/nodejs/18.12.1/bin/node]
```

I did some searching but I couldn't find anything useful to fix the issue. Does anything come to your mind regarding what might be causing this?

test/errors.tsx
```diff
@@ -3,30 +3,30 @@ import React from 'react';
...
-	restore = patchConsole();
+	restore = patchConsole(() => {});
 });
 
 test.after(() => {
 	restore();
 });
 
-test('catch and display error', t => {
+test.skip('catch and display error', t => {
```
pepicrft, line 19: @vadimdemedes the output changed and it's causing the test to fail. Where is the formatting of the error happening?

pepicrft:
Quick update here @vadimdemedes & @sindresorhus:
- The CI pipeline has been adjusted to run with Node 14, 16, and 18
- I updated `xo` and tackled some linting that showed up due to changes in rules.
- Went through all the tests and did the necessary adjustments to make them pass.
- Updated React to 18

There are some remaining bits but I'm a bit blocked with them:

- I added [skip](https://github.com/vadimdemedes/ink/pull/546/files#r1097459717) to a test that's testing that the right contend is being logged through the standard output. I can't figure out where the formatting is happening and why it broke with the code changes.
- I disabled the test modules that use `node-pty`. As mentioned [here](https://github.com/vadimdemedes/ink/pull/546#discussion_r1097450916), Node crashes when the tests try to interact with the pseudo-terminal instance. 

I'd appreciate your help to get this one to the finish line 🙏🏼

vadimdemedes:
Not sure what's happening yet, but running `npx xo` fails immediately with this error:

```
❯ npx xo
node:internal/process/esm_loader:94
    internalBinding('errors').triggerUncaughtException(
                              ^

Error [ERR_MODULE_NOT_FOUND]: Cannot find package '/Users/vadimdemedes/Projects/ink/node_modules/xo/node_modules/eslint/' imported from /Users/vadimdemedes/Projects/ink/node_modules/xo/index.js
Did you mean to import eslint/lib/api.js?
    at new NodeError (node:internal/errors:372:5)
    at legacyMainResolve (node:internal/modules/esm/resolve:341:9)
    at packageResolve (node:internal/modules/esm/resolve:941:14)
    at moduleResolve (node:internal/modules/esm/resolve:1003:20)
    at defaultResolve (node:internal/modules/esm/resolve:1218:11)
    at ESMLoader.resolve (node:internal/modules/esm/loader:580:30)
    at ESMLoader.getModuleJob (node:internal/modules/esm/loader:294:18)
    at ModuleWrap.<anonymous> (node:internal/modules/esm/module_job:80:40)
    at link (node:internal/modules/esm/module_job:78:36) {
  code: 'ERR_MODULE_NOT_FOUND'
}
```

Have you encountered this before? I deleted `node_modules` and reinstalled everything.

test/errors.tsx
vadimdemedes: Errors are caught in `App` component (src/components/App.tsx) and formatted inside `ErrorOverview` component (src/components/ErrorOverview.tsx).

vadimdemedes:
Same error is actually happening on CI.

src/components/Static.tsx
```diff
@@ -58,11 +58,9 @@ const Static = <T,>(props: Props<T>) => {
 	);
 
 	return (
-		<ink-box
-			// @ts-ignore
-			internal_static
-			style={style}
-		>
+		// eslint-disable-next-line @typescript-eslint/ban-ts-comment, @typescript-eslint/prefer-ts-expect-error
```
sindresorhus, line 61: Don't disable things that can be fixed. Here, the linting is correct.

vadimdemedes:
Pushed a commit that fixes the test you skipped in `test/errors.tsx` for testing error boundaries.

src/reconciler.ts
```diff
@@ -1,3 +1,4 @@
+import * as process from 'node:process';
```
sindresorhus, line 1:
```suggestion
import process from 'node:process';
```

vadimdemedes:
Trying to figure out why `node-pty` crashes the tests in this PR, but not in `master`. I'm using the same Node.js 16 version when running both branches, so it's weird why it would fail here.

vadimdemedes:
Fixed XO crash on launch by removing extra dependencies and reinstalling `eslint-config-xo-react` and React-related plugins. Hopefully it runs fine on CI now and we'll see lint errors being reported.

pepicrft:
> Trying to figure out why node-pty crashes the tests in this PR, but not in master. I'm using the same Node.js 16 version when running both branches, so it's weird why it would fail here.

Something perhaps related to `type: "module"` in the `package.json`?

pepicrft:
Thanks @vadimdemedes and @sindresorhus for your review and commits. I've created an issue on the [node-pty](https://github.com/microsoft/node-pty/issues/579) project to see if they can help us diagnose and fix the issue. I'm a bit out of ideas around how to tackle the problem.

vadimdemedes:
Wow, good catch!

.eslintignore
```diff
@@ -0,0 +1,6 @@
+/build/**
+/node_modules/**
+/examples/**
+/test/**
+/benchmark/**
```
sindresorhus, line 5: Why are these ignored by linting?

benchmark/simple/simple.tsx
```diff
@@ -1,7 +1,12 @@
+#!/usr/bin/env npx ts-node-esm
+
 /* eslint-disable react/jsx-curly-brace-presence */
 'use strict';
-const React = require('react');
-const {render, Box, Text} = require('../..');
+import React from "react";
+import {createRequire} from "node:module";
```
sindresorhus, line 6: Single-quotes

package.json
```diff
@@ -9,10 +9,11 @@
 		"email": "vdemedes@gmail.com",
 		"url": "https://github.com/vadimdemedes"
 	},
-	"main": "build",
-	"types": "build/index.d.ts",
+	"type": "module",
+	"exports": "./build/index.js",
+	"types": "./build/index.d.ts",
```
sindresorhus, line 14:
```suggestion
	"exports": {
		"types": "./build/index.d.ts",
		"default": "./build/index.js"
	},
```

src/components/ErrorOverview.tsx
```diff
@@ -1,27 +1,35 @@
-import * as fs from 'fs';
-import React, {FC} from 'react';
+import * as fs from 'node:fs';
```
sindresorhus, line 1:
```suggestion
import fs from 'node:fs';
```

src/ink.tsx
```diff
@@ -74,18 +76,25 @@ export default class Ink {
...
 			false,
+			null,
+			'id',
+			_error => {},
 			null
 		);
 
 		// Unmount when process exits
+		// eslint-disable-next-line @typescript-eslint/ban-ts-comment, @typescript-eslint/prefer-ts-expect-error
+		// @ts-ignore
```
sindresorhus, line 94: Use `@ts-expect-error`

test/errors.tsx
```diff
@@ -3,30 +3,30 @@ import React from 'react';
...
-	restore = patchConsole();
+	restore = patchConsole(() => {});
 });
 
 test.after(() => {
 	restore();
 });
 
-test('catch and display error', t => {
+test.only('catch and display error', t => {
```
sindresorhus, line 19:
```suggestion
test('catch and display error', t => {
```

test/exit.tsx
```diff
@@ -1,96 +1,103 @@
-import {serial as test} from 'ava';
-import {spawn} from 'node-pty';
-import {run} from './helpers/run';
-
-test('exit normally without unmount() or exit()', async t => {
+import { createRequire } from 'module';
+import test from "ava";
+import {run} from './helpers/run.js';
+import * as url from 'url';
+import * as path from "path";
```
sindresorhus, line 5:
```suggestion
import path from 'path';
```

test/helpers/create-stdout.ts
```diff
@@ -11,9 +11,13 @@ interface Stream extends EventEmitter {
 
 export default (columns?: number): Stream => {
 	const stdout = new EventEmitter();
+	// @ts-ignore
```
sindresorhus, line 14: ts-expect-error ^

test/text.tsx
```diff
@@ -40,25 +40,25 @@ test('text with rgb color', t => {
...
 });
 
-test('text with hsl color', t => {
-	const output = renderToString(<Text color="hsl(32, 100, 50)">Test</Text>);
-	t.is(output, chalk.hsl(32, 100, 50)('Test'));
-});
+// test('text with hsl color', t => {
+// 	const output = renderToString(<Text color="hsl(32, 100, 50)">Test</Text>);
+// 	t.is(output, chalk.hsl(32, 100, 50)('Test'));
+// });
```
sindresorhus, line 46: ^

sindresorhus (changes requested):

pepicrft:
@sindresorhus / @vadimdemedes thank you very much for the review. I addressed your comments and there are still some open items for discussion:

1. I [adjusted ESLint](https://github.com/vadimdemedes/ink/pull/546#discussion_r1101841122) to ignore the Javascript modules under `templates`, `test`, and `benchmark` because they yield around 300 issues and that'd increase the scope of the PR quite significantly. I propose that we do it in a follow-up PR.
2. The test [`useStdout - write to stdout`](https://github.com/vadimdemedes/ink/actions/runs/4134427011/jobs/7151690042)  seems to be flaky. It passes locally but fails on CI. Is it possible that React 18 might have impacted this? 
3. I disabled the tests at [`test/text.tsx`](https://github.com/vadimdemedes/ink/pull/546#discussion_r1101850473) because the `chalk.hsl` no longer exists in the most recent version of Chalk. How do you recommend going about this @sindresorhus ?
4. The test `ensure wrap-ansi doesn’t trim leading whitespace` fails locally and on CI. `wrap-ansi` shouldn't be trailing the leading whitespace but it does. I can't figure out why.

test/text.tsx
pepicrft: `chalk.hsl` no longer exists. What should I be using instead?

vadimdemedes:
> I https://github.com/vadimdemedes/ink/pull/546#discussion_r1101841122 to ignore the Javascript modules under templates, test, and benchmark because they yield around 300 issues and that'd increase the scope of the PR quite significantly. I propose that we do it in a follow-up PR.

Since all of these errors are the result of updating XO in this PR, I think it makes more sense to bundle fixes for those errors here as well. I fixed all errors in `test` folder to start with.

> The test [useStdout - write to stdout](https://github.com/vadimdemedes/ink/actions/runs/4134427011/jobs/7151690042) seems to be flaky. It passes locally but fails on CI. Is it possible that React 18 might have impacted this?

I pushed a temporary fix for that test, but I'd love to know why it broke first. I'll look into it, but help would be appreciated if possible 👍 

> I disabled the tests at https://github.com/vadimdemedes/ink/pull/546#discussion_r1101850473 because the chalk.hsl no longer exists in the most recent version of Chalk. How do you recommend going about this @sindresorhus ?

I decided to follow what Chalk did and remove hsl, hsv, hwb and ansi color formats from Ink as well. Doubt anyone used those in terminal UIs anyway.

> The test ensure wrap-ansi doesn’t trim leading whitespace fails locally and on CI. wrap-ansi shouldn't be trailing the leading whitespace but it does. I can't figure out why.

That's interesting, because it doesn't fail for me and seems to pass on CI as well? Could you double check that one?

src/colorize.ts
```diff
@@ -22,49 +18,55 @@ export default (
...
-	if (color in chalk) {
-		const method = getMethod(color, type);
-		return (chalk as any)[method](str);
+	if (isNamedColor(color)) {
+		if (type === 'foreground') {
+			return chalk[color](str);
+		}
+
+		const methodName = `bg${
+			color[0]!.toUpperCase() + color.slice(1)
```
LitoMore, line 27:
```suggestion
			color[0]!.toUpperCase() + color.slice(1)
```

I believe our maintainers don't want `!.` anywhere.

pepicrft:
> Since all of these errors are the result of updating XO in this PR, I think it makes more sense to bundle fixes for those errors here as well. I fixed all errors in test folder to start with.

@vadimdemedes I continued with `examples` and `benchmark`.

> I pushed a temporary fix for that test, but I'd love to know why it broke first. I'll look into it, but help would be appreciated if possible 👍

Let me give it another thought to see if I can understand the failure.

> I decided to follow what Chalk did and remove hsl, hsv, hwb and ansi color formats from Ink as well. Doubt anyone used those in terminal UIs anyway.

Sounds good :+1:

> That's interesting, because it doesn't fail for me and seems to pass on CI as well? Could you double check that one?

It's working now for me. Odd.

pepicrft:
@vadimdemedes I fixed all the examples. They were using `import-jsx`, which only works with CJS. The approach that I took was to add the esbuild Node loader a  a `devDependency` and [add a script](https://github.com/vadimdemedes/ink/pull/546/commits/b1461928560083cc661d2189eb9c3f65483b423d#diff-5a831ea67cf5cf8703b0de46901ab25bd191f56b320053be9332d9a3b0d01d15R1831) to the `package.json` to load a given test.

I also [reverted](https://github.com/vadimdemedes/ink/pull/546/commits/b1461928560083cc661d2189eb9c3f65483b423d#diff-720a909ed188aa6f228da389f0fd2dad16b983d0528e0a1e9bc84560468a2dcaR230) your changes to make the test pass. For some reasons I still don't know they are no longer necessary. I still have the feeling is flakiness but I can't figure out why.

Please, let me know if there's anything else to be addressed. Otherwise I'd appreciate if we can get it merged and release a new major version. We'd like to use the changes in this PR to prevent multiple conflicting versions of React in the dependency graph.

Thanks a lot for your support in making this one happen 🙏🏼

vadimdemedes:
> I also [reverted](https://github.com/vadimdemedes/ink/pull/546/commits/b1461928560083cc661d2189eb9c3f65483b423d#diff-720a909ed188aa6f228da389f0fd2dad16b983d0528e0a1e9bc84560468a2dcaR230) your changes to make the test pass. For some reasons I still don't know they are no longer necessary. I still have the feeling is flakiness but I can't figure out why.

@pepicrft Looks like it's still failing on CI without this fix that got reverted.

pepicrft:
> @pepicrft Looks like it's still failing on CI without this fix that got reverted.

Ouch :/. It is possible then that the test's implementation depends on the environment in which it runs.

vadimdemedes:
Oh my, after several days of debugging, I finally fixed the [`useStdout` test](https://github.com/vadimdemedes/ink/actions/runs/4256542080/jobs/7405558134) in `test/hooks.tsx` and CI is green again.

For context, Ink has 3 rendering modes:

- Default mode - renders as soon as possible, throttled to 32ms.
- CI mode - collects all render frames (cycles) and writes them to stdout once after app is unmounted. This is necessary, because CI services normally don't support ANSI escapes that clear lines.
- Debug mode - renders immediately without throttling and without clearing previous output. Basically `console.log`.

When I ran this PR's tests locally, `useStdout` test has succeeded. On GitHub Actions it was always failing. Expected output from `test/fixtures/use-stdout.tsx` script is:

```
Hello world
Hello from Ink to stdout
Hello world
exited
```

But the first "Hello world" was never displayed on CI.

I started adding a bunch of `console.log`s to tests and Ink's codebase and I noticed something odd. [`isCi`](https://github.com/pepicrft/ink/blob/esm/src/ink.tsx#L17) was still `true` when running `test/fixtures/use-stdout.tsx`. Despite the fact that Ink's tests explicitly delete `CI` environment variable → https://github.com/pepicrft/ink/blob/2238385acff19b6b12159c40dd702da24ce264f0/test/hooks.tsx#L31.

Since it worked before this PR, there must've been some change in how CI enivronment is detected. I traced it down to this commit in `ci-info` package, which is a dependency of `is-ci` that Ink uses → https://github.com/watson/ci-info/commit/24cc450.

Looks like deleting `CI` environment isn't enough anymore and an explicit `CI=false` must be set to disable CI detection. For context, we want to make Ink's tests think they're not running in CI, so that we can replicate behavior of Ink in real-world usage.

So here's the fix → https://github.com/pepicrft/ink/commit/67d7a8ae2a35c5091ef43c73a4d3d944cee8b208.

vadimdemedes:
The last thing to figure out before shipping this PR is a replacement of [import-jsx](https://www.npmjs.com/package/import-jsx), since it's not supported in ESM anymore. It seems to be quite popular, so we need to have either a replacement or some recommendations in the migration guide.

vadimdemedes:
Submitted https://github.com/vadimdemedes/import-jsx/pull/18 to support ESM in `import-jsx`. I also realized that packages like [ink-text-input](https://github.com/vadimdemedes/ink-text-input) also need to be converted to ESM. Hopefully we can ship it all this week.

pepicrft:
Just WOW on that [thorough debugging](https://github.com/vadimdemedes/ink/pull/546#issuecomment-1445059387). I wouldn't have been able to figure that out myself.
Regarding the updating of `import-jsx`, how do you feel about using an existing Node loader and deprecating that package?  It'll be less maintenance work for you going forward and we can focus the efforts on Ink. Is there any Ink component other than [ink-text-input](https://github.com/vadimdemedes/ink-text-input) that we should also update? I'm happy to take a look at that one.

vadimdemedes:
Thanks for an amazing contribution, @pepicrft 🚢

vadimdemedes:
> Regarding the updating of import-jsx, how do you feel about using an existing Node loader and deprecating that package? It'll be less maintenance work for you going forward and we can focus the efforts on Ink.

Replied in https://github.com/vadimdemedes/import-jsx/pull/18#issuecomment-1448658760.

> Is there any Ink component other than [ink-text-input](https://github.com/vadimdemedes/ink-text-input) that we should also update? I'm happy to take a look at that one.

Yep, here are the packages I wanted to update too:
- https://github.com/vadimdemedes/create-ink-app
- https://github.com/vadimdemedes/ink-select-input
- https://github.com/vadimdemedes/ink-spinner
- https://github.com/vadimdemedes/ink-text-input
- https://github.com/vadimdemedes/ink-testing-library

Going to start working on these too and report back.

vadimdemedes:
Thanks for an amazing contribution, @pepicrft 🚢

pepicrft:
Thanks a lot @vadimdemedes and @sindresorhus for supporting through this effort. It's a pleasure contributing to open source projects like Ink.

pepicrft:
@vadimdemedes when is the next release going out?

Issue #552, opened 2023-03-02
TypeScript doesn't find type declarations in v4.0.0


AlCalzone:
I believe `package.json` must be changed in one of the following ways:

```diff
	"type": "module",
-	"exports": {
-		"types": "./build/index.d.ts",
-		"default": "./build/index.js"
-	},
+	"module": "./build/index.js",
+	"types": "./build/index.d.ts",
```
(tested, works) or
```diff
	"type": "module",
	"exports": {
-		"types": "./build/index.d.ts",
-		"default": "./build/index.js"
+		".": {
+			"types": "./build/index.d.ts",
+			"default": "./build/index.js"
+		}
	},
```
according to https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-7.html#packagejson-exports-imports-and-self-referencing

AlCalzone:
typescript 4.8 in VSCode?

AlCalzone:
Ok, so I did this
```
npm init -y
npm i typescript ink-repro
echo 'import {render} from "ink"' > index.mts
code index.mts
```

and got:
![grafik](https://user-images.githubusercontent.com/17641229/222525474-70874752-faec-496f-804c-160afd2d2571.png)

It does start working after I add a tsconfig with the corresponding `module` and `moduleResolution` flags (or change it in an existing project).
I'd argue though that adding two lines (first diff) is a small price to pay to allow consumers not to change their config, causing other issues down the line.

maxijonson:
I have the same issue. 

@sindresorhus Not sure what "issue with your build tool" means if all other packages work just fine. 

I'm new to Ink, is it ESM only? (similar to how the latest versions of Chalk are). @AlCalzone  seems to be using "mts" files. Changing my `module` would be more of a hassle and if Ink is now ESM only, I don't quite understand packages like Chalk that force ESM to their consumers 😅 But I'm not here to debate build practices, I'd just like a solution that doesn't break the rest of my development workflow.

Furthermore, on NPM, there seem to be no included type definitions, hence the "DT" logo, suggesting a @types package, but the package has a deprecation notice stating it's a stub. However, I can confirm that they are present in my node_modules, but I don't understand why TypeScript (and NPM) isn't seeing the declarations. Again, Ink is the only package in my project with this kind of issue.

sindresorhus:
> I'm new to Ink, is it ESM only?

[...]

PR #553, opened 2023-03-07
Add fallback for TS `moduleResolution: 'node'` users

amitdahan:
# Summary

Using the new major (which relies on `exports` field alone for `.d.ts`/`.js` resolution) with a TS project that's still using `"moduleResolution": "node"` (not `nodenext`) means TS doesn't leverage `exports` at all.

So using the new major like this results (during type-check / `tsc` build) in:
```
error TS2307: Cannot find module 'ink' or its corresponding type declarations.

  import { Box, render, Text } from "ink";
                                    ~~~~~
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,6 +10,7 @@
 		"url": "https://github.com/vadimdemedes"
 	},
 	"type": "module",
+	"types": "./build/index.d.ts",
 	"exports": {
 		"types": "./build/index.d.ts",
 		"default": "./build/index.js"


amitdahan:
If this & https://github.com/vadimdemedes/ink-spinner/pull/17 are accepted I'll do the rest

vadimdemedes:
@sindresorhus mentioned in https://github.com/vadimdemedes/ink/issues/552 that current setup is correct, so it looks like indeed an issue in your TS config. I've lost track of all the mess about exporting modules, so I'll just trust Sindre's opinion on this one, since he definitely knows the right answer to this one.

sindresorhus:
`"types": "./build/index.d.ts",` will just paper over the real problem and it may stop working at any time. Ink requires ESM, which requires changing your tsconfig to output ESM. There's no way around that.

I have a guide here on how to get it working: https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c#how-can-i-make-my-typescript-project-output-esm

fantua:
Can't we just inline `exports` field? Package still will be ESM and will be backward compatible with TS also.
```diff
	"type": "module",
-	"exports": {
-		"types": "./build/index.d.ts",
-		"default": "./build/index.js"
-	},
+	"exports": "./build/index.js",
+	"types": "./build/index.d.ts",
```
For example this way [chalk](https://github.com/chalk/chalk/blob/main/package.json#L10) and [got](https://github.com/sindresorhus/got/blob/702ed352ac5d8527deaf4a5821d0274afc56e9d8/package.json#L9-L10) are  defined.

sindresorhus:
Chalk and Got are only defined that way as it will require a major release to change them. They will be changed at some point too.

nicolewhite:
Has anyone been able to get this working? I've had to downgrade to ink version 3 after failing to get the types to work in VSCode. Tried following [this guide](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c) but making those changes breaks the types for other packages

dymk:
Yeah, that guide isn't a general solution. It's simply not tractable to take projects of any moderate complexity and convert them into ESM. If they have dependencies which aren't ESM, they can't be converted. I'm disappointed the maintainer has taken such a rigid approach to compatibility, excluding large number of projects that would benefit from ink.

PR #556, opened 2023-03-13
feat: add `<Line>` component to render horizontal and vertical lines

AlCalzone:
I previously tried to achieve this by measuring a `Box` and rendering text instead, but as mentioned in https://github.com/vadimdemedes/ink/pull/472#issuecomment-1466413182, the `measureElement` API often does not work in more complex layouts (returns `NaN`).
Since `ink` does draw boxes correctly, I figured I'd try this by implementing a `Line` component, so `ink` can handle drawing the lines.

### DIFF
--- a/src/styles.ts
+++ b/src/styles.ts
@@ -3,6 +3,7 @@ import Yoga, {type YogaNode} from 'yoga-layout-prebuilt';
 import {type Boxes} from 'cli-boxes';
 import {type LiteralUnion} from 'type-fest';
 import {type ForegroundColorName} from 'chalk';
+import {DOMNode} from './dom.js';
 
 export type Styles = {
 	readonly textWrap?:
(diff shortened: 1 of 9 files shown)


src/styles.ts
```diff
@@ -336,17 +362,32 @@ const applyDimensionStyles = (node: YogaNode, style: Styles): void => {
...
+const applyBorderStyles = (
+	domNode: DOMNode & {yogaNode: YogaNode},
+	style: Styles
+): void => {
+	const node = domNode.yogaNode;
+
 	if ('borderStyle' in style) {
-		const borderWidth = typeof style.borderStyle === 'string' ? 1 : 0;
+		const borderWidth =
+			domNode.nodeName === 'ink-line'
```
AlCalzone, line 386: I needed access to `nodeName` here, so I figured I'd update all the applyXYZ method to take the DOM node instead.

vadimdemedes:
I'd prefer to ship https://github.com/vadimdemedes/ink/pull/475 instead, which will let you render `Box` border on a specific side, which is essentially the same as a `Line` component introduced here.

AlCalzone:
Fine by me. Out of curiosity, what's blocking that PR?

vadimdemedes:
Nothing technical, only my availability.

vadimdemedes:
Closing this, since border sides can now be disabled individually → https://github.com/vadimdemedes/ink#bordertop.

PR #558, opened 2023-03-15
Update documentation and file names imports

hamzahsn:
This is a tiny change after going through the documentation and noticing that some links are not working properly as well as the file names are not having the correct extensions when importing them.

### DIFF
--- a/benchmark/static/index.ts
+++ b/benchmark/static/index.ts
@@ -1,1 +1,1 @@
-import './static.js';
+import './static.tsx';
--- a/readme.md
+++ b/readme.md
@@ -895,7 +895,7 @@ Ink uses border styles from [`cli-boxes`](https://github.com/sindresorhus/cli-bo
 
 <img src="media/box-borderStyle.jpg" width="521">
 
-See example in [examples/borders](examples/borders/borders.js).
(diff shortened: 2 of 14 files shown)


benchmark/static/index.ts
```diff
@@ -1 +1 @@
-import './static.js';
+import './static.tsx';
```
vadimdemedes, line 1: These were correct, it should have a `.js` extension, since after transpilation `static.tsx` is written to `static.js`. See last point in https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c#how-can-i-make-my-typescript-project-output-esm.

readme.md
```diff
@@ -895,7 +895,7 @@ Ink uses border styles from [`cli-boxes`](https://github.com/sindresorhus/cli-bo
 
 <img src="media/box-borderStyle.jpg" width="521">
 
-See example in [examples/borders](examples/borders/borders.js).
+See example in [examples/borders](examples/borders/borders.tsx).
```
vadimdemedes, line 898: Good job catching these.

vadimdemedes:
Fixed links in 2a67354.

Issue #542, opened 2022-12-14
waitUntilExit() not resolved with syntax errors


elmpp:
- Node 16.6.1
 - Jest 29.3.1
 - OSX Ventura 13.0.1 (22A400)

The following hangs:

```
import { render, Text, useApp } from 'ink'
import React, { ErrorInfo, useCallback } from 'react'

jest.setTimeout(2000)


class ErrorBoundary extends React.Component<{cb: (err: Error | null, errorInfo: ErrorInfo) => void}> {

  static getDerivedStateFromError(error: Error) {
    return null // https://tinyurl.com/y4k99pwr
  }
  componentDidCatch(error: Error, errorInfo: React.ErrorInfo) {
    this.props.cb(error, errorInfo)
    this.setState(null) // prevents - https://github.com/reactjs/reactjs.org/issues/3028
  }
  render() {
    return this.props.children
  }
}

const MalformedComponent = () => {

  const renderSomething = () => <Text>Broken</Text>
  return (<React.Fragment key="meh">
    {renderSomething()} */} // intentional syntax error
  </React.Fragment>)
}
MalformedComponent.displayName = 'MalformedComponent'


const RenderAndExit: React.FC<{}> = () => {
  const {exit} = useApp()
  const errorBoundaryCb = useCallback(
    (err, errInfo) => {
      console.log(`errorBoundaryCb called :>> `, err, errInfo)
      exit(err)
      console.log(`errorBoundaryCb called 2 :>> `, err, errInfo)
    }
    ,
    [],
  )

  return (
    <ErrorBoundary cb={errorBoundaryCb}>
      <React.Fragment>
        <MalformedComponent />
      </React.Fragment>
    </ErrorBoundary>
  )
}

const renderWithInk = async () => {

  try {
    const {waitUntilExit} = render(<RenderAndExit />, {})
    await waitUntilExit() // hangs here (despite exit() being called). App unmount waitUntilExit is neither resolved nor rejected as expected
    console.log(`:>> not reached`);
  }
  catch (err) {
    console.log(`err :>> `, err) // not called as expected!!
  }
}

describe('render-component', () => {
  it(`app not unmounting with malformed component despite exit() being called with error`, async () => {
    await renderWithInk()
  })
})

```

It's strange because waitUntilExit rejects as expected if the rendered component simply throws an error like:

```
const MalformedComponent = () => {
  useEffect(() => {
    setTimeout(() => {
      cb(new Error('cb supplied error'))
[...]

PR #563, opened 2023-03-18
Fix `waitUntilExit()` not resolved with sync errors

fantua:
Fixes https://github.com/vadimdemedes/ink/issues/542

### DIFF
--- a/test/errors.tsx
+++ b/test/errors.tsx
@@ -15,14 +15,14 @@ test.after(() => {
 	restore();
 });
 
-test('catch and display error', t => {
+test('catch and display error', async t => {
 	const stdout = createStdout();
 
 	const Test = () => {
 		throw new Error('Oh no');
 	};
 
-	render(<Test />, {stdout});
+	const {waitUntilExit} = render(<Test />, {stdout});
 
 	t.deepEqual(
 		stripAnsi((stdout.write as any).lastCall.args[0])
@@ -40,9 +40,11 @@ test('catch and display error', t => {
 			" 22:     throw new Error('Oh no');",
 			' 23:   };',
 			' 24:',
-			' 25:   render(<Test />, {stdout});',
+			' 25:   const {waitUntilExit} = render(<Test />, {stdout});',
 			'',
 			' - Test (test/errors.tsx:22:9)'
 		]
 	);
+
+	await t.throwsAsync(waitUntilExit, {message: 'Oh no'});
 });
--- a/test/errors.tsx
+++ b/test/errors.tsx
@@ -15,14 +15,14 @@ test.after(() => {
 	restore();
 });
 
-test('catch and display error', async t => {
+test('catch and display error', t => {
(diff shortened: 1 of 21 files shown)


test/errors.tsx
```diff
@@ -40,9 +40,11 @@ test('catch and display error', t => {
...
 			' 23:   };',
 			' 24:',
-			' 25:   render(<Test />, {stdout});',
+			' 25:   const {waitUntilExit} = render(<Test />, {stdout});',
 			'',
 			' - Test (test/errors.tsx:22:9)'
 		]
 	);
+
+	await t.throwsAsync(waitUntilExit, {message: 'Oh no'});
```
fantua, line 49: fix for `Unhandled rejection in test/errors.tsx`

test/exit.tsx
```diff
@@ -54,7 +54,7 @@ test.serial('exit on unmount() with raw mode', async t => {
 
 test.serial('exit with thrown error', async t => {
 	const output = await run('exit-with-thrown-error');
-	t.true(output.includes('errored'));
```
fantua, line 57: Currently test pass because we print error message in [ErrorOverview](https://github.com/vadimdemedes/ink/blob/master/src/components/ErrorOverview.tsx)

vadimdemedes: Could you undo all changes to existing tests and create new tests instead? I want to make sure existing functionality doesn't break.

test/errors.tsx
fantua: > To avoid it, we should add something like a isWaitingForExit boolean property to Ink and set it to true in waitUntilExit method.

With synchronous errors `waitUntilExit` is never called. I can add some prop to render method maybe. Wdyt?

vadimdemedes: Hm, wasn't that the purpose of this PR? To resolve/reject the promise returned by `waitUntilExit` for synchronous errors?

vadimdemedes: Hm, that's not good. Whatever solution we decide to go with, it needs to work with this use case, but without unhandled rejection promises.

test/exit.tsx
fantua: done

test/errors.tsx
fantua: Hey @vadimdemedes. I've pushed new implementation, please have a look 🙂

vadimdemedes: I'm not a fan of an option that makes `waitUntilExit` work. As a user of Ink, I'd expect it to just work when I call it.

It looks like this problem isn't solvable without breaking changes. Perhaps we can think what the better API would look like if we weren't constrained by the presence of `waitUntilExit` and ship this in a major release.

sindresorhus:
Closing for lack of activity and the fact that a better API is required: https://github.com/vadimdemedes/ink/issues/542#issuecomment-3194386543

Issue #560, opened 2023-03-16
Performance issue in `sliceAnsi` function


AlCalzone:
I noticed that when there are multiple components that frequently render, like the `ink-spinner` component, my application ends up with high CPU load. A somewhat complex screen goes from 0.5 to 25% CPU usage when there are 3 spinners rendered.

So far, I've traced it to the `sliceAnsi` function calls in `Output.get()`. Will investigate further if I can see which part of that function seems to cause this.

Here's a simple component to reproduce:
```tsx
import { Box, render, Text } from "ink";
import Spinner from "ink-spinner";

const Root: React.FC = () => {
  const numSpinners = 10;
  const loop = new Array(numSpinners).fill(0);
  return (
    <Box flexDirection="row" borderColor="blue" borderStyle="round">
      {loop.map((_, i) => (
        <Box key={i} borderColor="white" borderStyle="double">
          <Text color="green">
            <Spinner type="dots" />
          </Text>
        </Box>
      ))}
    </Box>
  );
};

render(<Root />);
```
With just 10 spinners, I get significant slowdown. 30 spinners bump the CPU usage up to almost 100%.

AlCalzone:
I understand this might better be opened in the `slice-ansi` repo, but maybe there's something on the consumer side to improve performance too, e.g. caching calls to `sliceAnsi` etc.

vadimdemedes:
Could you share some details how you traced to `slice-ansi`?

AlCalzone:
I executed the process in VSCode with debugger attached, then captured a performance profile for a couple of seconds. That showed that most of the time was spent in  sliceAnsi.
I can share some screenshots if you like.

There should also be a way to do it with command line flags to node.js, but I find the above easier.

vadimdemedes:
Yep, would be cool to see the details. Perhaps there are some other things that are standing out that we could optimize?

AlCalzone:
Ok, so here's a video:

https://user-images.githubusercontent.com/17641229/226054976-71488549-d27e-43eb-881f-c8d4d53cc326.mp4

At the end you see how much time is spent in each function, **self time** is the time the function itself was executing, **total time** includes the functions it calls. So a method with a high self-time is worth looking at.

You can also switch to a flamegraph to see the time slices over the entire execution time, or switch it to the left-heavy view to group the calls by callstack:
[...]

PR #564, opened 2023-03-19
Speed up generation of output

AlCalzone:
This is a proof of concept for a different rendering approach that does not rely on `sliceAnsi` as heavily.

I tried this with my test component from #560, but with 50 spinners. Without this change, 50 spinners put one core of my CPU at 100%, and rendering lags heavily. With this change, I'm looking at ~40% utilization of one core, which is still not ideal, but renders fluently.

The output for 10s of profiling now looks like this, is idle half the time and most of the work is done in `wasm`, which is probably from `yoga` internals:
grafik

fixes: https://github.com/vadimdemedes/ink/issues/560

This will need some more testing with emojis and clipping, but I wanted to get this out for some early feedback.

### DIFF
--- a/src/output.ts
+++ b/src/output.ts
@@ -198,12 +198,16 @@ export default class Output {
 					const chars = styledCharsFromTokens(tokenize(line));
 					let offsetX = x;
 					for (const char of chars) {
+						if (offsetX >= this.width) break;
+
 						currentLine[offsetX] = char;
 
 						// Some characters take up more than one column. In that case, the following
 						// pixels need to be cleared to avoid printing extra characters
 						const isWideChar = char.fullWidth || char.value.length > 1;
 						if (isWideChar) {
+							if (offsetX + 1 >= this.width) break;
+
 							currentLine[offsetX + 1] = {
 								type: 'char',
 								value: '',
--- a/src/output.ts
+++ b/src/output.ts
@@ -6,7 +6,7 @@ import {
 	styledCharsFromTokens,
 	styledCharsToString,
 	tokenize
-} from 'ansi-tokenize';
+} from '@alcalzone/ansi-tokenize';
 import {type OutputTransformer} from './render-node-to-output.js';
 
 /**
--- a/src/output.ts
+++ b/src/output.ts
@@ -98,7 +98,7 @@ export default class Output {
 
 	get(): {output: string; height: number} {
 		// Initialize output array with a specific set of rows, so that margin/padding at the bottom is preserved
-		const output: Array<StyledChar[]> = [];
+		const output: StyledChar[][] = [];
 
 		for (let y = 0; y < this.height; y++) {
 			const row: StyledChar[] = [];
--- a/src/output.ts
+++ b/src/output.ts
@@ -198,8 +198,6 @@ export default class Output {
 					const chars = styledCharsFromTokens(tokenize(line));
 					let offsetX = x;
 					for (const char of chars) {
-						if (offsetX >= this.width) break;
-
 						currentLine[offsetX] = char;
 
 						// Some characters take up more than one column. In that case, the following
@@ -225,7 +223,12 @@ export default class Output {
 		}
 
 		const generatedOutput = output
-			.map(line => styledCharsToString(line).trimEnd())
+			.map(line => {
+				// See https://github.com/vadimdemedes/ink/pull/564#issuecomment-1637022742
+				const lineWithoutEmptyItems = line.filter(item => item !== undefined);
+
+				return styledCharsToString(lineWithoutEmptyItems).trimEnd();
+			})
 			.join('\n');
 
 		return {
--- a/src/output.ts
+++ b/src/output.ts
@@ -1,13 +1,13 @@
 import sliceAnsi from 'slice-ansi';
 import stringWidth from 'string-width';
 import widestLine from 'widest-line';
-import {type OutputTransformer} from './render-node-to-output.js';
 import {
-	StyledChar,
+	type StyledChar,
 	styledCharsFromTokens,
 	styledCharsToString,
 	tokenize
 } from 'ansi-tokenize';
+import {type OutputTransformer} from './render-node-to-output.js';
 
 /**
  * "Virtual" output class
@@ -110,6 +110,7 @@ export default class Output {
 					styles: []
 				});
 			}
+
 			output.push(row);
 		}
 
--- a/src/output.ts
+++ b/src/output.ts
@@ -98,10 +98,11 @@ export default class Output {
 
 	get(): {output: string; height: number} {
 		// Initialize output array with a specific set of rows, so that margin/padding at the bottom is preserved
-		const output: StyledChar[][] = [];
+		const output: Array<StyledChar[]> = [];
 
 		for (let y = 0; y < this.height; y++) {
 			const row: StyledChar[] = [];
+
 			for (let x = 0; x < this.width; x++) {
 				row.push({
 					type: 'char',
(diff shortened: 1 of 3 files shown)


AlCalzone:
Ok, looks like emojis and some full-width characters are causing test failures, I'll look into that.

And there's probably an ANSI code in a different position than before - I don't see a visual difference here:
grafik

vadimdemedes:
This is really impressive! Would you be willing to push this PR to completion?

AlCalzone:
Yeah, will do soon.

AlCalzone:
Tests are passing too now. Just for fun, I ran `time npm run benchmark benchmark/simple` to compare.
**master**
```
real    0m11,412s
user    0m13,342s
sys     0m0,604s
```

**this PR**
```
real    0m8,208s
user    0m9,051s
sys     0m0,330s
```

So roughly 32% faster rendering for even simple outputs.

src/output.ts
```diff
@@ -92,10 +98,20 @@ export default class Output {
 
 	get(): {output: string; height: number} {
 		// Initialize output array with a specific set of rows, so that margin/padding at the bottom is preserved
-		const output: string[] = [];
+		const output: StyledChar[][] = [];
```
vadimdemedes, line 101:
```suggestion
		const output: Array<StyledChar[]> = [];
```

src/output.ts
```diff
@@ -172,29 +188,41 @@ export default class Output {
...
-						sliceAnsi(currentLine, x + width);
+					const chars = styledCharsFromTokens(tokenize(line));
+					let offsetX = x;
+					for (const char of chars) {
+						currentLine[offsetX] = char;
+
+						// Some characters take up more than one column. In that case, the following
+						// pixels need to be cleared to avoid printing extra characters
+						const charWidth = char.fullWidth ? 2 : char.value.length;
+						for (let dx = 1; dx < charWidth; dx++) {
```
vadimdemedes, line 206: Could you use a descriptive variable name for index instead of `dx`? Does `characterIndex` fit here?

src/output.ts
```diff
@@ -172,29 +188,41 @@ export default class Output {
...
 
 					for (const transformer of transformers) {
 						line = transformer(line);
 					}
 
-					output[y + offsetY] =
-						sliceAnsi(currentLine, 0, x) +
-						line +
-						sliceAnsi(currentLine, x + width);
+					const chars = styledCharsFromTokens(tokenize(line));
```
vadimdemedes, line 198:
```suggestion
					const characters = styledCharsFromTokens(tokenize(line));
```

src/output.ts
```diff
@@ -172,29 +188,41 @@ export default class Output {
...
+					let offsetX = x;
+					for (const char of chars) {
```
vadimdemedes, line 200:
```suggestion
					for (const character of characters) {
```

src/output.ts
vadimdemedes, line 206: Why does this skip the first character and starts from the second one?

vadimdemedes:
Superb work 🔥

src/output.ts
AlCalzone: Because the first char is the char that's actually written, so width 1 is already accounted for.

vadimdemedes: Not sure I understand, could you elaborate the "that's actually written" part?

AlCalzone: Line 201 writes the current character to the output array. Some characters are wider than a single column though, so we have to set the following columns (up to the character width) to "" so we don't end up with extra characters.
So if the char is a 2-column emoji, we write the emoji to column x and "" to column x+1. The character after the emoji will then correctly be at column x+2.

src/output.ts
```diff
@@ -172,29 +188,41 @@ export default class Output {
...
+
+						// Some characters take up more than one column. In that case, the following
+						// pixels need to be cleared to avoid printing extra characters
+						const charWidth = char.fullWidth ? 2 : char.value.length;
+						for (let dx = 1; dx < charWidth; dx++) {
+							currentLine[offsetX + dx] = {
+								type: 'char',
+								value: '',
+								fullWidth: false,
+								styles: char.styles
```
AlCalzone, line 211: maybe this should be
```suggestion
								styles: []
```
though.

src/output.ts
vadimdemedes: Do we need a loop here then? Could this be simplified to this?

```js
if (char.fullWidth) {
  currentLine[offsetX + 1] = /* empty char */
}

offsetX += character.fullWidth ? 2 : 1
```

AlCalzone: Aren't there some emojis which have length > 4?

AlCalzone: I've pushed a commit to eliminate the loop.

src/output.ts
```diff
@@ -198,12 +198,16 @@ export default class Output {
 					const chars = styledCharsFromTokens(tokenize(line));
 					let offsetX = x;
 					for (const char of chars) {
+						if (offsetX >= this.width) break;
```
vadimdemedes, line 201: How did you run into this issue where an out-of-bounds write would occur? Could you add a test for it to ensure it doesn't regress back?

AlCalzone: Done. In my case it was a pretty complicated nested layout, but I managed to reduce it to a `<Text wrap="end">` that's longer than the stdout width.

vadimdemedes: There's no `wrap="end"` value, but there's `wrap="truncate-end"`. I updated it in the test you added, then commented out "prevent out-of-bounds writes" commit changes and it didn't crash for me. Looks like we don't need this code then?

AlCalzone: Hmm maybe the test case isn't correct then. I ended up with gaps in the output array, which crashed the function that converts Tokens back to strong.

vadimdemedes: Do you still see the same issue when using `wrap="truncate-end"` instead of `wrap="end"` (which is an invalid prop)?

AlCalzone: Will try tomorrow or so.

vadimdemedes: Hey @AlCalzone, do you still have the capacity to finish this PR? Would really love to ship it. No worries if you're busy though.

vadimdemedes: No worries, take your time 👍

vadimdemedes:
Thanks @AlCalzone, that was really helpful! Here's what I've found.

I took your `Box` example and rendered it in the terminal.

```jsx
import React from 'react';
import {Box, render} from 'ink';

function Test() {
	return <Box width={12} height={10} borderStyle={'round'}></Box>;
}

process.stdout.columns = 10;
render(<Test />, {debug: true});
```

CleanShot 2023-07-16 at 11 13 06@2x

The right border is rendered in the wrong place, even though there are correct X and Y coordinates for all sides:

- Top border has X = 0, Y = 0
- Left border has X = 0, Y = 1
- Right border has X = 11, Y = 1
- Bottom border has X = 0, Y = 9

Here's where each line is written to output:

https://github.com/vadimdemedes/ink/blob/8a0476024cf997e0dbc02e493ad3972349b02474/src/output.ts#L187-L190

I added this `console.log` to see what exactly gets written and where:

```js
console.log({
	// X coordinate of the `line` to write to output
	x,
	// Existing output before X coordinate
	before: sliceAnsi(currentLine, 0, x),
	beforeWidth: sliceAnsi(currentLine, 0, x).length,
	// New text to write to output
	line,
	lineWidth: line.length,
	// Existing output after X coordinate + width of the `line`
	after: sliceAnsi(currentLine, x + width),
	afterWidth: sliceAnsi(currentLine, x + width).length
});
```

Here's the output for one of the lines for writing the right border:

```
{
  x: 11,
  before: '│         ',
  beforeWidth: 10,
  line: '│',
  lineWidth: 1,
  after: '',
  afterWidth: 0
}
```

Normally, `sliceAnsi(currentLine, 0, x)` should return a string where `length === x`, but here `beforeWidth` equals 10, not 11. The reason is `currentLine` comes from the `output` variable, which is initialized with strings that are as wide as the terminal (in our case terminal width is 10).

https://github.com/vadimdemedes/ink/blob/8a0476024cf997e0dbc02e493ad3972349b02474/src/output.ts#L174

https://github.com/vadimdemedes/ink/blob/8a0476024cf997e0dbc02e493ad3972349b02474/src/output.ts#L97-L99

This creates a problem where `output` contains lines with the length of 10, so the index of the last character in each line is 9. At the same time, we're trying to write a character at index 11, which results in invalid output where character at index 10 is missing. As a result, right border is placed at X = 10, not X = 11.

It doesn't cause a crash in the current version of Ink, because `sliceAnsi` is used to extract string ranges, which doesn't fail if you run `sliceAnsi(str, 0, 11)` where `str.length === 10`. However, your PR no longer uses `sliceAnsi` and is sensitive to `undefined` items.

To sum up, this is an existing issue of Ink with content larger than the terminal and should be fixed separately from this PR. To avoid regressions in this PR though, we need to skip `undefined` array items to keep the rendering behavior the same as it is now.

vadimdemedes:
Pushed a commit to keep existing behavior with out of bounds writes.

vadimdemedes:
CI is green, merging!

Thank you for this incredible performance boost, @AlCalzone!

PR #579, opened 2023-04-25
Add top-level `types` to `package.json`

draconisNoctis:
Typescript doesn't support `exports.types` in `package.json`, yet.
So, I added the supported top-level property.


sindresorhus:
Duplicate of https://github.com/vadimdemedes/ink/issues/552

Issue #585, opened 2023-05-03
Output terminal isn't cleared when second to last frame is taller than terminal


matteodepalo:
Hi,
As per description, I noticed that the terminal is correctly cleared if the last frame output is taller than the terminal window, however, if the last frame is shorter, the second to last frame isn't cleared. Here's a video to demonstrate this:

https://user-images.githubusercontent.com/151725/235908525-e94e1955-4663-425d-a2f6-4430200365be.mp4

I would expect ink to not show the list after pressing enter. Is this expected behavior?

For now I've littered my code with things like this, but I would expect Ink to handle such scenarios.

```ts
if (stdout && wrapperHeight >= stdout.rows - 1) {
  stdout.write(ansiEscapes.clearTerminal)
}

setSomething(false)
```

Where `wrapperHeight` is the height of the `Box` containing everything.

Code to reproduce (make your terminal window height less than 20 lines):

```ts
import React, { useState } from 'react';
import { render, Text, Box, useInput } from 'ink';

function ClearTerminalBug() {
  const lines = [];

  for (let i = 1; i <= 20; i++) {
    lines.push(`line ${i}`);
  }

  const [done, setDone] = useState(false)

  useInput((_input, key) => {
    if (key.return) {
      setDone(true)
    }
  });

  return done ? <Text>done</Text> : <Box flexDirection="column">
    {lines.map((line, index) => (
      <Text key={index}>{line}</Text>
    ))}
  </Box>
}

render(
  <ClearTerminalBug />
)
```

Reproduction repo https://github.com/matteodepalo/clear-terminal

matteodepalo:
After creating this issue I noticed that this doesn't happen just on exit, it happens even without it. I've updated the issue.

vadimdemedes:
Thanks for a detailed bug report! Submitted a fix over here → https://github.com/vadimdemedes/ink/pull/586.

PR #586, opened 2023-05-03
Fix rendering when terminal is erased on first render

vadimdemedes:
Fixes https://github.com/vadimdemedes/ink/issues/585.

@matteodepalo did a great job documenting the issue, so I'm not going to repeat it here.

## Context

Ink has two ways of rendering output, depending how tall it is:

1. If output is shorter than terminal height, Ink erases the lines from last output and writes lines from the new output. Ink tracks how many lines did the last output have for this to work. This is handled in https://github.com/vadimdemedes/ink/blob/master/src/log-update.ts, which is a fork of [`log-update`](https://github.com/sindresorhus/log-update) package.

2. If output is equal or taller than terminal height, Ink erases the entire terminal history (or scrollback as it's commonly referred to) and writes new output. Note that Ink doesn't need to track how many lines to erase here, because it just erases the entire contents of the current terminal.
 https://github.com/vadimdemedes/ink/blob/8a0476024cf997e0dbc02e493ad3972349b02474/src/ink.tsx#L176-L182

## Root cause

The key insight here is this:

> Note that Ink doesn't need to track how many lines to erase here, because it just erases the entire contents of the current terminal.

In https://github.com/vadimdemedes/ink/issues/585, the very first render starts with the second rendering method and erases the terminal completely. On next render, the first method kicks in, but there's no information how many lines to erase, since second method doesn't store that information.

That's why the output from the first render doesn't get erased, since for `logUpdate` this variable is zero.

https://github.com/vadimdemedes/ink/blob/8a0476024cf997e0dbc02e493ad3972349b02474/src/log-update.ts#L12

## Solution

This PR fixes this bug by separating the write operation that erases the terminal and writes the output. This way, even if the entire terminal is erased, `logUpdate` is still used to render the output and keep track of how many lines were written.

I've also added `force` option to `logUpdate` render function to force it to always render the string we give to it without bailing out if it equals to the last output. This can happen when [`Static`](https://github.com/vadimdemedes/ink#static) output changes, but normal output isn't.

https://github.com/vadimdemedes/ink/blob/8a0476024cf997e0dbc02e493ad3972349b02474/src/log-update.ts#L23-L25

I've also added `erase` option to `logUpdate` render function, so that we can tell it to not do erasing of its own, if we erased the entire terminal already.

### DIFF
--- a/src/ink.tsx
+++ b/src/ink.tsx
@@ -175,8 +175,10 @@ export default class Ink {
 
 		if (outputHeight >= this.options.stdout.rows) {
 			this.options.stdout.write(
-				ansiEscapes.clearTerminal + this.fullStaticOutput + output
+				ansiEscapes.clearTerminal + this.fullStaticOutput
 			);
+
+			this.log(output, {force: true, erase: false});
 			this.lastOutput = output;
 			return;
 		}
--- a/src/log-update.ts
+++ b/src/log-update.ts
@@ -5,7 +5,7 @@ import cliCursor from 'cli-cursor';
 export type LogUpdate = {
 	clear: () => void;
 	done: () => void;
-	(str: string): void;
+	(str: string, options?: {force?: boolean; erase?: boolean}): void;
 };
 
 const create = (stream: Writable, {showCursor = false} = {}): LogUpdate => {
@@ -13,19 +13,25 @@ const create = (stream: Writable, {showCursor = false} = {}): LogUpdate => {
 	let previousOutput = '';
 	let hasHiddenCursor = false;
 
-	const render = (str: string) => {
+	const render = (
+		str: string,
+		{force = false, erase = true}: {force?: boolean; erase?: boolean} = {}
+	) => {
 		if (!showCursor && !hasHiddenCursor) {
 			cliCursor.hide();
 			hasHiddenCursor = true;
 		}
 
 		const output = str + '\n';
-		if (output === previousOutput) {
+		if (output === previousOutput && !force) {
 			return;
 		}
 
 		previousOutput = output;
-		stream.write(ansiEscapes.eraseLines(previousLineCount) + output);
+
+		const eraser = erase ? ansiEscapes.eraseLines(previousLineCount) : '';
+		stream.write(eraser + output);
+
 		previousLineCount = output.split('\n').length;
 	};
 
--- a/test/fixtures/erase-once-with-static.tsx
+++ b/test/fixtures/erase-once-with-static.tsx
@@ -1,0 +1,33 @@
+import process from 'node:process';
+import React, {useState} from 'react';
+import {Box, Static, Text, render, useInput} from '../../src/index.js';
+
+function Test() {
+	const [fullHeight, setFullHeight] = useState(true);
+
+	useInput(
+		input => {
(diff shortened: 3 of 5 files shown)


vadimdemedes:
Not sure why tests are failing on CI, while passing locally consistently. Will need to look into it. Help also appreciated to ship this fix faster!

so1ve:
You can use https://github.com/nektos/act to debug the ci

amcaplan:
@vadimdemedes I tried this change locally which seems to CI. Bit of a hack, but it seems justifiable:

```
diff --git a/test/render.tsx b/test/render.tsx
index 7b57075..e27e537 100644
--- a/test/render.tsx
+++ b/test/render.tsx
@@ -31,8 +31,9 @@ const term = (
                reject = reject2;
        });
 
+       const {CI, ...ciAgnosticEnv} = process.env;
        const env = {
-               ...process.env,
+               ...ciAgnosticEnv,
                // eslint-disable-next-line @typescript-eslint/naming-convention
                NODE_NO_WARNINGS: '1'
        };
```

Because CI environments are so different, it probably makes sense to test the erasing behavior in a non-CI-like environment. Same reason we're already using `node-pty`.

matteodepalo:
@vadimdemedes I tried this change locally setting `CI=true` and it seems to do the trick https://github.com/vadimdemedes/ink/pull/609/commits/cae25d51b02932e840fd4d79d410d5a8d405870a

It makes sense because in CI we're not erasing anything so `ansiEscapes.clearTerminal` isn't included in the output.

For some reason though, on my branch `xo` throws an error. I had the same issue locally. Can it be something related to some permissions lacking in the fork and locally that you have? Could you try applying this change in this branch?

matteodepalo:
@vadimdemedes strangely with [this fix](https://github.com/vadimdemedes/ink/pull/609/commits/178a873f36c8fd0e038701b115e3f73318e62718), tests pass locally if I run `CI=true npm run test`, but they don't seem to pass on CI.

The reason for that fix is that in CI we don't clear the terminal, we return earlier in the code, so it makes sense that the check for `ansiEscapes.clearTerminal` fails.

DaniGuardiola:
Hey there, wondering if there are any updates on this work. I have a need on this being resolved. Is there anything I can do to help get this over the finish line? :)

matteodepalo:
@DaniGuardiola unfortunately I haven't had the time to continue on this bug, the issue as I left it was one with tests. The code was working fine. CI is failing while local tests are passing, but I couldn't find the root cause. If you'd like to pick it up you can also check [what I tried to do in a separate branch](https://github.com/vadimdemedes/ink/pull/609/files) to maybe get some ideas.

vadimdemedes:
@DaniGuardiola The weird issue is that I fixed the original issue in this PR and confirmed it works on my laptop, but when I submitted this PR, it consistently fails on CI. Huge help would be to figure out what's going on in CI here to unblock this PR.

DaniGuardiola:
I changed my approach and this stopped being a blocker for me. I don't think I'll have the time to look into it, sorry.

sindresorhus:
Closing for lack of activity and in favor of https://github.com/vadimdemedes/ink/pull/717

PR #597, opened 2023-06-01
Fix devtools import in production

eXhumer:
webpack bundler sets the `process.env.NODE_ENV` to `production` when `mode` is set to `production` in webpack config file. My aim with this commit is to skip devtools injection and import in production mode.


vadimdemedes:
This would be a breaking change. Can you set `DEV=false` environment variable in your webpack config, so it strips out this code the same way?

PR #599, opened 2023-06-06
remove tscrypt error when working within IDE

olegtaranenko:
eliminated annoying ts configuration mismatch, which produces red marks working with examples inside IDE
I'm not sure how it could affect on the other parts of project but to me look all good.
See https://stackoverflow.com/questions/62273153/this-module-is-declared-with-using-export-and-can-only-be-used-with-a-defau where I got a solution -)

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,6 +1,7 @@
 {
 	"extends": "@sindresorhus/tsconfig",
 	"compilerOptions": {
+		"esModuleInterop": true,
 		"moduleResolution": "node16",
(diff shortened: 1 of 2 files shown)


tsconfig.json
```diff
@@ -1,6 +1,7 @@
 {
 	"extends": "@sindresorhus/tsconfig",
 	"compilerOptions": {
+		"esModuleInterop": true,
```
sindresorhus, line 4: This is moot. It's already `true` by default.

PR #607, opened 2023-07-08
[Chore] Fix CI

dkempner:
Fixes #606

PR #609, opened 2023-07-14
Fix rendering when terminal is erased on first render

matteodepalo:
I've created a new branch from [your fix](https://github.com/vadimdemedes/ink/pull/586) and tried to fix CI.


dkempner:
[Fix CI](https://github.com/vadimdemedes/ink/pull/607)

PR #610, opened 2023-07-19
fix: add main to work around tsserver bug

isaacs:
Same issue as https://github.com/vadimdemedes/ink-testing-library/pull/21


isaacs:
Not planned. https://github.com/vadimdemedes/ink-testing-library/pull/21#issuecomment-1646985536

Issue #611, opened 2023-07-21
RFE: expose index of output line in Transform


isaacs:
I need to add a hanging indent to some text.  I'm able to do it, albeit with a few contortions:

```ts
import { Text, TextProps } from 'ink'
const sigil = '\x01\x02\x03\x03\x02\x01'
const HangingIndent: FC<{ indent?: number } & TextProps> = ({
  indent = 4,
  children,
  ...props
}) => (
  <Transform
    transform={o =>
      o.includes(sigil) ? o.replace(sigil, '') : ' '.repeat(indent) + o
    }>
    <Text {...props}>
      {sigil}
      {children}
    </Text>
  </Transform>
)

// usage: <HangingIndent dimColor bold color="red">Some long string of text blah blah blah</HangingIndent>
```

Of course, this will break if the string contains the sigil, but I'm reasonably sure no one is going to have that string of control characters in this situation. But it does feel like a weird hack.

If the `<Transform>` method knew what line it was working on, then this could be much cleaner.  I'm thinking of it somewhat like `Array.map`:

```ts
import React, { FC } from 'react'
import { Text, TextProps } from 'ink'
const HangingIndent: FC<{ indent?: number } & TextProps> = ({
  indent = 4,
  children,
  ...props
}) => (
  <Transform
    transform={(o, line) => line === 0 ? o : ' '.repeat(indent) + o
    }>
    <Text {...props}>
      {children}
    </Text>
  </Transform>
)
```

vadimdemedes:
Transforms are applied line by line, if I remember correctly, so this change makes sense to me. Feel free to open a PR.

PR #619, opened 2023-08-25
Add `index` property to Transform

isaacs:
This makes it possible to know which line is being transformed.

Fix: https://github.com/vadimdemedes/ink/issues/611

There are some linting errors, but I'm not sure how to go about fixing them. `for(;;)` (or maybe `.forEach()`) is the right thing to use here, since the index is relevant, but xo really wants to only ever see `for(of)`.

### DIFF
--- a/src/output.ts
+++ b/src/output.ts
@@ -187,9 +187,7 @@ export default class Output {
 
 				let offsetY = 0;
 
-				for (let index = 0; index < lines.length; index++) {
-					let line = lines[index];
-					if (line === undefined) continue;
+				for (let [index, line] of lines.entries()) {
 					const currentLine = output[y + offsetY];
 
 					// Line can be missing if `text` is taller than height of pre-initialized `this.output`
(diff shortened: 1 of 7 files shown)


src/output.ts
```diff
@@ -187,7 +187,9 @@ export default class Output {
 
 				let offsetY = 0;
 
-				for (let line of lines) {
+				for (let index = 0; index < lines.length; index++) {
```
sindresorhus, line 190:
```suggestion
				for (let [index, line] of lines.entries()) {
```

isaacs: Ah, yeah, that works, too, I guess.  I'll use that.

isaacs:
Pushed a commit that fixes all lint errors, and most of the warnings.  I didn't feel confident reducing parameter counts or indent levels, since that's a more involved refactoring than just running `prettier --format` and adding a few overrides.

vadimdemedes:
Thanks!

PR #622, opened 2023-09-05
Fix inability to read user input after app exit

matteodepalo:
There is a bug I introduced with https://github.com/vadimdemedes/ink/pull/616

If someone invokes `render` once with a component that uses `useInput`, then everything works fine (and that's why tests are passing).
If instead you invoke `render`, wait for Ink to exit and then later in the execution you invoke it again with the same component, then node just quits because `stdin` has been dereferenced in the previous call to `render`.

With this PR I add a call to `ref` so that when the second call happens node stays up.


vadimdemedes:
Thanks!

PR #623, opened 2023-09-07
Fix broken link escapes

matteodepalo:
A bug has been introduced with the work done [here](https://github.com/vadimdemedes/ink/pull/564) and it relates to the tokenization of ansi escape characters. We noticed it with `ansiEscapes.link` and I added a test case in this repo which now fails. 

@AlCalzone I don't know enough about your `ansi-tokenize` project, but could this be fixed there?

Here's a screenshot of the failing test:

Screenshot 2023-09-07 at 11 25 04

You can see from the screenshot that the link escape is not being closed properly so the link "leaks" in all the text after that.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,7 @@
 		"text"
 	],
 	"dependencies": {
-		"@alcalzone/ansi-tokenize": "^0.1.2",
+		"@alcalzone/ansi-tokenize": "^0.1.3",
 		"ansi-escapes": "^6.0.0",
 		"auto-bind": "^5.0.1",
 		"chalk": "^5.2.0",
--- a/test/components.tsx
+++ b/test/components.tsx
@@ -722,10 +722,10 @@ test('vertical spacer', t => {
 	t.is(output, 'Top\n\n\n\n\nBottom');
 });
 
-test('ansi escapes are closed properly', t => {
+test('link ansi escapes are closed properly', t => {
 	const output = renderToString(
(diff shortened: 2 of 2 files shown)


AlCalzone:
I agree this needs to be fixed in `ansi-tokenize`

AlCalzone:
Fixed in `v0.1.3`: https://github.com/AlCalzone/ansi-tokenize/releases/tag/v0.1.3

(except that your test has extra `"` around the string)

matteodepalo:
@AlCalzone we also saw a difference in the ansi escapes related to colors when upgrading. For example this component

```
<Box>
  <Text color="red">hello</Text>
  <Text color="green">world</Text>
</Box>
```

Before it would produce

`'[31mhello[39m[32mworld[39m'`

Now it produces

`'[31mhello[32mworld[39m'`

There seems to be a missing `[39m` after the first word which I think it's the code to reset the color? Not sure if this is a difference in the way `chalk` works, maybe it has been updated? The end result doesn't seem to change much in our outputs.

AlCalzone:
That's expected. ansi-tokenize uses the least amount of ansi codes necessary for a style change.

matteodepalo:
Got it! Make sense, I'll update our tests.

vadimdemedes:
Thanks @matteodepalo for the PR and thanks @AlCalzone for fixing it so quickly 💛

PR #631, opened 2023-09-19
fix: only focus active elements

acdibble:
When wrapping around at the beginning or end of the list of components, we should search respectively for the last or first _active_ component.

I didn't use `Array.prototype.findLast` because the `tsconfig` doesn't support it and it was simple enough to write a loop.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -1858,7 +1858,7 @@ const Example = () => {
 
 Switch focus to the next focusable component.
 If there's no active component right now, focus will be given to the first focusable component.
-If active component is the last in the list of focusable components, focus will be switched to the first component.
+If active component is the last in the list of focusable components, focus will be switched to the first focusable component.
 
 **Note:** Ink calls this method when user presses <kbd>Tab</kbd>.
 
--- a/readme.md
+++ b/readme.md
@@ -1858,7 +1858,7 @@ const Example = () => {
 
 Switch focus to the next focusable component.
 If there's no active component right now, focus will be given to the first focusable component.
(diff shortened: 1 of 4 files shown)


sindresorhus:
> I didn't use Array.prototype.findLast because the tsconfig doesn't support it and it was simple enough to write a loop.

We now target Node.js 18, so it should be fine to use `.findLast`.

acdibble:
> > I didn't use Array.prototype.findLast because the tsconfig doesn't support it and it was simple enough to write a loop.
> 
> We now target Node.js 18, so it should be fine to use `.findLast`.

Cool, I'll fix it up soon

sindresorhus (approved):

vadimdemedes:
@acdibble Thanks for this PR! Could you add a test for this behavior?

acdibble:
> @acdibble Thanks for this PR! Could you add a test for this behavior?

@vadimdemedes , I added three tests.

I also had to update the tsconfig to include the `ES2023` lib in order to get the compilation to work and the tests to run. This seems safe because Node 18 supports all of ES2023 (https://node.green/#ES2023).

readme.md
```diff
@@ -1858,7 +1858,7 @@ const Example = () => {
 
 Switch focus to the next focusable component.
 If there's no active component right now, focus will be given to the first focusable component.
-If active component is the last in the list of focusable components, focus will be switched to the first component.
+If active component is the last in the list of focusable components, focus will be switched to the first focusable component.
```
acdibble, line 1861:
```suggestion
If active component is the last in the list of focusable components, focus will be switched to the first active component.
```

PR #638, opened 2023-11-09
Use simpler CI detection

sindresorhus:
> The `is-ci` package attempts to detect every CI service, which is unsustainable. It also has a higher risk of false-positives. For example, it detects the environment variable RUN_ID as CI-specific, although other services could use it. Constant updates for new CIs create version fragmentation, resulting in inconsistent behavior across dependent packages. Pushing for CI services to use a standardized CI environment variable is a more robust solution.
> - https://github.com/sindresorhus/is-in-ci#how-is-this-different-from-is-ci

---

Feel free to close if you disagree. Just a suggestion.


src/ink.tsx
```diff
@@ -16,7 +16,6 @@ import logUpdate, {type LogUpdate} from './log-update.js';
 import instances from './instances.js';
 import App from './components/App.js';
 
-const isCi = process.env['CI'] === 'false' ? false : originalIsCi;
```
sindresorhus, line 19: `is-in-ci` already has this check

vadimdemedes:
👍

PR #673, opened 2024-09-14
Replace lodash with es-toolkit

wojtekmaj:
es-toolkit is a modern lodash alternative that claims to be much faster and lighter, potentially improving overall install and runtime performance: https://es-toolkit.slash.page/intro.html


src/ink.tsx
```diff
@@ -53,19 +54,13 @@ export default class Ink {
...
 
 		this.rootNode.onImmediateRender = this.onRender;
 		this.log = logUpdate.create(options.stdout);
 		this.throttledLog = options.debug
 			? this.log
-			: throttle(this.log, undefined, {
-					leading: true,
-					trailing: true,
-				});
+			: throttle(this.log, undefined);
```
sindresorhus, line 63: I would just do this to make it simple:

```suggestion
			: (throttle(this.log, undefined) as LogUpdate);
```

PR #674, opened 2024-09-26
Update useInput() to include `name` in `key`

depeele:
Some key values (e.g. 'insert', 'home', 'end') were basically lost. In these cases, the `input` value would be an empty string and the generated `key` data would contain no additional information about the value.

To resolve this, directly include `keypress.name` in the generated `key` data so this information is not lost.

### DIFF
--- a/src/hooks/use-input.ts
+++ b/src/hooks/use-input.ts
@@ -140,6 +140,7 @@ const useInput = (inputHandler: Handler, options: Options = {}) => {
 			const keypress = parseKeypress(data);
 
 			const key = {
+				name: keypress.name,
 				upArrow: keypress.name === 'up',
 				downArrow: keypress.name === 'down',
 				leftArrow: keypress.name === 'left',


sindresorhus:
Check if any types / docs needs to be updated.

depeele:
> Check if any types / docs needs to be updated.

OK. I've update:
- the `Key` type in src/hooks/use-input.ts to include the new `name` field;
- documentation for `inputHandler()` callback for `useInput()` to include a description of the new `key.name`;

readme.md
```diff
@@ -1459,6 +1459,27 @@ Type: `object`
 
 Handy information about a key that was pressed.
 
+###### key.name
```
sindresorhus, line 1462: The TS docs and readme should be in sync.

readme.md
```diff
@@ -1459,6 +1459,27 @@ Type: `object`
...
+Currently identified meta keys:
+- f1 - f12
+- escape
+- number (for all number keys 0-9)
+- return
+- up, down, left, right (arrow keys)
+- insert, delete
+- home, end
+- pageup, pagedown
+- space
```
sindresorhus, line 1480:
```suggestion
- `space`
```

And I would list one per line.

sindresorhus (changes requested):

readme.md
depeele: Done

readme.md
depeele: Pull request updated with changes to sync readme and TS docs

sindresorhus:
https://github.com/vadimdemedes/ink/pull/674#issuecomment-2478559767

PR #682, opened 2024-11-27
add support for iterm natural editing mode

bcherny:
Quick improvement to enable opt+left and opt+right

Screenshot 2024-11-26 at 4 19 16 PM


sindresorhus:
Are you sure this will not conflict with other terminals that may use these escapes for something else?

PR #685, opened 2024-12-02
Fix border collapsing when `borderStyle` changes

osaton:
Instead of this `Content`, you'll get


araguma:
@sindresorhus Would it be possible to merge this pull request? I'm looking to implement a recursive tree component using the borderStyle feature, but this bug leaves unwanted empty spaces after the second render.

sindresorhus:
If this is still relevant, would you be able to fix the merge conflict?

PR #689, opened 2024-12-06
Add option to render using full available height (rows)

DanielSRS:
Currently if one wants to create an application that fills all available columns and rows, they need to create a component that listens to stdout size changes, like this:

```tsx
import React from 'react';
import { Box, render, Text } from 'ink';
import useStdoutDimensions from 'ink-use-stdout-dimensions';

function Application() {
  const [columns, rows] = useStdoutDimensions();
  return (
    <Box width={columns} height={rows}>
      <Text>Content</Text>
    </Box>
  );
}

render(<Application />);
```

But ink already listens and recalculate layout when stdout resizes, but only for width (columns)

#### this PR adds an render option to also use height (rows) size when recalculating layout.

so, one only needs to pass an optional render option, like this:

```tsx
import React from 'react';
import { render, Text } from 'ink';

function Application() {
  return (
    <Text>Content</Text>
  );
}

render(<Application />, {fullHeight: true});
```

Added:
- fullHeight render option
- an example that uses full height


sindresorhus:
This needs tests and readme updates.

PR #695, opened 2025-01-23
Update styles.ts: add support to top,left,bottom and right

YPetremann:
This add support for absolute and relative positionning with top,left,bottom and right

    <Box position="absolute" bottom={0} right={0} width={5} height={5} borderStyle="single">
      <Text>Hello</Text>
    </Box>

This directly use yoga node.setPosition(edge, position) :
- If the node is static then insets have no effect.
- If the node is relative then insets will move the node the designated amount from where the node would normally be in the container. Each inset property is relative to the node's corresponding physical edge. So YGNodeStyleSetPosition(node, YGEdgeRight, 10.0f) will offset the node so that the right edge is 10 units away from where the right edge would be originally.
- If the node is absolute then insets will move the node the designated amount away from the node's [containing block](https://www.yogalayout.dev/docs/advanced/containing-block). So YGNodeStyleSetPosition(node, YGEdgeRight, 10.0f) will offset the node so that the right edge is 10 units away from the [containing block's](https://www.yogalayout.dev/docs/advanced/containing-block) right edge.


sindresorhus:
This will need tests, types, and docs updates.

PR #708, opened 2025-04-07
incremental rendering

bolinfest:
This PR tries to pick up where https://github.com/vadimdemedes/ink/pull/413 left off. Specifically, instead of writing the full UI state back to `stdout` as part of `render()`, do some basic line-by-line comparisons so that the only things that we write to stdout are:

- content of lines that changed (plus `ansiEscapes.eraseEndLine`)
- ANSI escape codes to put the cursor in the right place

For cases where you have a virtual spinner or text input where only one line of the screen is changing at a time, this results in considerably less output written to stdout for each call to `render()`.

### DIFF
--- a/src/log-update.ts
+++ b/src/log-update.ts
@@ -13,10 +13,21 @@ const create = (stream: Writable, {showCursor = false} = {}): LogUpdate => {
 	let previousOutput = '';
 	let hasHiddenCursor = false;
 
+	function clear(): void {
+		stream.write(ansiEscapes.eraseLines(previousLineCount));
+		previousOutput = '';
+		previousLineCount = 0;
+	}
+
 	const render = (str: string) => {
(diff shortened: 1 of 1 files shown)


ducan-ne:
have you tested this? I checked but the experience is not good
image
Not sure if it's because of my code though

pkkim:
Hi, unfortunately I was able to reproduce the problems @ducan-ne had with this, plus the flickering was still present. @bolinfest if you are looking to continue working on this, I would be happy to assist. If not, I can take over the PR. Grateful to you for getting started on this.


https://github.com/user-attachments/assets/e4d27022-1c87-4204-b212-c1f57000a511

sindresorhus:
Still interested in finishing this?

No worries if not. Then I will close so something else has a chance.

This will also need tests.

sindresorhus:
Closing in favor of #781

PR #717, opened 2025-05-07
Fix stale output when outputHeight changes from above stdout.rows to below stdout.rows

### LINKED ISSUE #585: Output terminal isn't cleared when second to last frame is taller than terminal

thkim1011:
When the outputHeight goes from a value ` >= stdout.rows` to a value `< stdout.rows` in between re-render, it is possible that the old content is not completely removed from the terminal. 

This happens for three reasons
1. We actually want to consider the previous output height when deciding to entirely clear the terminal or only delete several lines. If the current output has a `height >= stdout.rows`, but we only need to clear the previous render's output, there's no point in writing `ansiEscapes.clearTerminal`. Similarly, if the current output has a `height < stdout.rows` but the previous output exceeded `stdout.rows`, then writing `ansiEscapes.eraseLines(previousLineCount)` won't clear all of the old output.
2. `this.log.clear()` doesn't actually track the height of the output if we rendered via the `outputHeight >= this.options.stdout.rows` conditional. I've added a helper `LogUpdate.sync` to keep it in sync.
3. The `outputHeight >= this.options.stdout.rows` conditional differs slightly in behavior to `this.log` in that it does not append a newline at the end.

Addressing all three of these points fixes the bug.

To test, I verified locally using a small example, and I've also added a test (see "erase screen where state changes in small viewport") which fails on master but passes in this branch.

Fixes #585

### DIFF
--- a/src/devtools-window-polyfill.ts
+++ b/src/devtools-window-polyfill.ts
@@ -5,6 +5,7 @@ import ws from 'ws';
 const customGlobal = global as any;
 
 // These things must exist before importing `react-devtools-core`
+// eslint-disable-next-line n/no-unsupported-features/node-builtins
 customGlobal.WebSocket ||= ws;
 
 customGlobal.window ||= global;
--- a/src/ink.tsx
+++ b/src/ink.tsx
@@ -175,7 +175,7 @@ export default class Ink {
 
 		if (this.lastOutputHeight >= this.options.stdout.rows) {
 			this.options.stdout.write(
-				ansiEscapes.clearTerminal + this.fullStaticOutput + output + "\n",
+				ansiEscapes.clearTerminal + this.fullStaticOutput + output + '\n',
 			);
 			this.lastOutput = output;
 			this.lastOutputHeight = outputHeight;
--- a/src/ink.tsx
+++ b/src/ink.tsx
@@ -34,6 +34,7 @@ export default class Ink {
 	// Ignore last render after unmounting a tree to prevent empty output before exit
 	private isUnmounted: boolean;
 	private lastOutput: string;
+	private lastOutputHeight: number;
 	private readonly container: FiberRoot;
 	private readonly rootNode: dom.DOMElement;
 	// This variable is used only in debug mode to store full static output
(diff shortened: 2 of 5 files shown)


src/devtools-window-polyfill.ts
```diff
@@ -5,6 +5,7 @@ import ws from 'ws';
 const customGlobal = global as any;
 
 // These things must exist before importing `react-devtools-core`
+// eslint-disable-next-line n/no-unsupported-features/node-builtins
```
thkim1011, line 8: Unclear why this lint error just started to appear, but should be safe to ignore.

thkim1011:
Correct me if I'm wrong but I think my PR fixes https://github.com/vadimdemedes/ink/issues/585.

thkim1011:
@sindresorhus can I get a re-run of the ci?

thkim1011:
@sindresorhus just a heads up that this is ready for review.

Issue #688, opened 2024-12-06
React 19 support


dumbmatter:
It seems that the current version of ink does not work with React 19. For example with some minimal code like this:

```js
import React from 'react';
import {render, Text} from 'ink';

render(React.createElement(Text, null, "Hello world"));
```

I get this output with react 18.3.1:

```
$ node index.js
Hello world
```

And this with react 19.0.0:

```
$ node index.js
/home/user/ink-react-19/node_modules/.pnpm/react-reconciler@0.29.2_react@19.0.0/node_modules/react-reconciler/cjs/react-reconciler.development.js:491
var ReactCurrentOwner = ReactSharedInternals.ReactCurrentOwner;
                                             ^

TypeError: Cannot read properties of undefined (reading 'ReactCurrentOwner')
    at $$$reconciler (/home/user/ink-react-19/node_modules/.pnpm/react-reconciler@0.29.2_react@19.0.0/node_modules/react-reconciler/cjs/react-reconciler.development.js:491:46)
    at file:///home/user/ink-react-19/node_modules/.pnpm/ink@5.1.0_react@19.0.0/node_modules/ink/build/reconciler.js:61:16

Node.js v22.11.0
```

marviobezerra:
Same here

MattyBalaam:
Is someone using React internals? Hope they don't get fired.

KrisWis:
So the only thing left is to either downgrade React to version 18 or just wait?

KrisWis:
But, the thing is, I don't use ink directly, I use Loki (for UI tests in conjunction with Storybook), and it in turn uses react-reconciler, which already uses ink. It's like this /:

KrisWis:
> @KrisWis There's a temporary fork with a viable React 19 upgrade: https://github.com/QuentinFontenay/ink/tree/react-v19.0.0

But, the thing is, I don't use ink directly, I use Loki (for UI tests in conjunction with Storybook), and it in turn uses react-reconciler, which already uses ink. It's like this /:

jakubmazanec:
Based on the forked and updated Ink linked above (thank you @QuentinFontenay), I created this patch: https://gist.github.com/jakubmazanec/3b2b0c86bb7f34fe9ae0937d608b590b

Use package `patch-package` to apply it, and also add following overrides to your `package.json`:

```
"overrides": {
  "react": "^19.0.0",
  "react-dom": "^19.0.0",
  "react-reconciler": "^0.31.0"
}
```

[...]

PR #719, opened 2025-05-27
React 19 Support

msmps:
I spent some time over the weekend working on adding support for React 19. I’ve successfully got it working, but opted to stick with the **legacy rendering mode** rather than enabling concurrent mode for now.

This means that React will continue to render updates synchronously, maintaining the existing behaviour.

At this point, **all but two tests are passing**, and I'm currently stumped on why these specific tests are failing. I’d appreciate any suggestions or help debugging them:

- `borders › render border after update`
- `focus › switch focus to the last component if currently focused component is the first one on Shift+Tab`

Additionally, the types provided by DefinitelyTyped are not yet up to date with the latest version of the React reconciler, so I’ve added a few `@ts-expect-error` comments to work around that for the time being.

I’ve published a testing version of this fork as:

```
msmps-ink@6.0.0-beta.1
```

## Concurrent Mode

I did experiment with enabling concurrent mode and got it working, but the test suite revealed several issues:

- Many tests fail unless wrapped in `act()` and converted to async.
- There are frequent warnings about state updates occurring outside of `act()`.
- It looked like `unmount()` is/was firing multiple times, potentially causing these warnings.

It felt like concurrent mode would require a more involved refactor. I’d suggest we consider this as a **fast follow-up** once React 19 legacy support is merged.

I’ve pushed a [separate branch](https://github.com/msmps/ink/tree/feat/react-v19-support-concurrent) with the concurrent mode changes if anyone wants to explore or contribute to that effort.

Addresses: #688 #656

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
 		"indent-string": "^5.0.0",
 		"is-in-ci": "^1.0.0",
 		"patch-console": "^2.0.0",
-		"react-reconciler": "^0.29.0",
+		"react-reconciler": "^0.32.0",
 		"scheduler": "^0.23.0",
 		"signal-exit": "^3.0.7",
 		"slice-ansi": "^7.1.0",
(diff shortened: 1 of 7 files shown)


sindresorhus:
@msmps Thanks for doing the work on this 🙏

sindresorhus:
Moved concurrent discussion here: https://github.com/vadimdemedes/ink/issues/720

PR #727, opened 2025-06-27
Add Cursor component

### LINKED ISSUE #251: cursor support?
Hi!

First, `ink` is an amazing awesome project and I'm so glad it exists.

Have you considered support for mutating the cursor position? I'm working on the repl for [rickbutton/turbo](https://github.com/rickbutton/turbo) and need a way to represent the actual cursor.

Ideally, the cursor could exist as a special component, that when drawn on the screen moves the cursor to its actual position in the output. As a usage suggestion, does this make sense?

```jsx
import { Cursor } from "ink";
function Component(props) {
    return <span>Some text <Cursor/> </span>;
}
```

I would love to contribute an implementation of this if you think this is acceptable.

safu9:
Fixes #251

This PR adds a new component `<Cursor>`, which controls the position of an actual cursor in a terminal by ANSI escape codes.

Previously, there was no way to control the position and visibility of an actual cursor in Ink applications.
We could show a fake cursor, as [ink-text-input](https://github.com/vadimdemedes/ink-text-input/blob/4ac429ca11da17c2321f7dec1076e38c06d738d5/source/index.tsx#L91) does, but that only works for visual purposes and still causes problems with IME. Without an actual cursor, the IME always displays a text being entered on the last line.
The`<Cursor>` component works as an anchor for IME to display at the intended position.

## Example

```jsx
<Box borderStyle="single" margin={10} paddingX={1}>
	<Text>Label: </Text>
	<Cursor />
	<Text>Type here</Text>
</Box>
```

| Without Cursor | With Cursor |
| -- | -- |
| <video src="https://github.com/user-attachments/assets/f96889f6-dcc9-4ee5-9ac2-35042dc07dc8" /> | <video src="https://github.com/user-attachments/assets/0f829042-77f9-4f44-af30-8d91ad622b4a" /> |

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -131,6 +131,7 @@ Feel free to play around with the code and fork this repl at [https://repl.it/@v
   - [`<Spacer>`](#spacer)
   - [`<Static>`](#static)
   - [`<Transform>`](#transform)
+  - [`<Cursor>`](#cursor)
 - [Hooks](#hooks)
   - [`useInput`](#useinputinputhandler-options)
   - [`useApp`](#useapp)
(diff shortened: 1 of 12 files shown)


safu9:
Sorry, I just realized that it can't support multiline texts in this way...
Closing for now

PR #736, opened 2025-07-25
feat: add ci option to override CI environment detection

### LINKED ISSUE #718: isInCi is overly aggressive, and produces confusing hangs
If you happen to have an env var set on your machine that starts with `CI_`, all interactive ink apps hang indefinitely.

I find this behaviour unappealing, and have spent an unsatisfactory amount of time diagnosing it.

For example:

```sh
export CI_PRIVATE_INTERNAL_BEST_FOOD=potatoes
npm run -s example examples/use-input
```

This prints nothing and hangs. If you quit (`q`, or `ctrl+c`), you see the buffered prompt, and it cleanly exits.

Tested on linux (ubuntu 25.04 / gnome-terminal / konsole) and macos (Sequoa / 15 / iterm2).

allenhutchison:
## Summary
This PR adds an optional `ci` parameter to render options that allows explicitly controlling CI behavior, overriding the automatic detection from the `is-ci` package.

## Motivation
Some users have environments where environment variables incorrectly trigger CI mode detection, causing their Ink applications to only render the final frame instead of showing animations. This provides a way to force the desired behavior.

## Changes
- Added optional `ci?: boolean` parameter to `RenderOptions` and `Options` types
- Modified `Ink` constructor to use the `ci` option when provided, falling back to `isInCi` detection
- Added tests to verify both `ci: true` and `ci: false` behaviors work correctly

## Usage
```typescript
// Force CI behavior (only render last frame)
render(<App />, {ci: true});

// Force non-CI behavior (render all frames)
render(<App />, {ci: false});

// Use default is-ci detection (current behavior)
render(<App />);
```

## Testing
- Added test cases for both `ci: true` and `ci: false` scenarios
- All existing tests continue to pass
- The feature is backwards compatible - when `ci` is not specified, behavior remains unchanged

Fixes #718
Fixes https://github.com/google-gemini/gemini-cli/issues/1563

### DIFF
--- a/src/ink.tsx
+++ b/src/ink.tsx
@@ -26,6 +26,7 @@ export type Options = {
 	exitOnCtrlC: boolean;
 	patchConsole: boolean;
 	waitUntilExit?: () => Promise<void>;
+	ci?: boolean;
 };
 
 export default class Ink {
@@ -44,11 +45,13 @@ export default class Ink {
 	private exitPromise?: Promise<void>;
 	private restoreConsole?: () => void;
 	private readonly unsubscribeResize?: () => void;
+	private readonly isCi: boolean;
 
 	constructor(options: Options) {
 		autoBind(this);
 
 		this.options = options;
+		this.isCi = options.ci ?? isInCi;
 		this.rootNode = dom.createNode('ink-root');
 		this.rootNode.onComputeLayout = this.calculateLayout;
 
@@ -113,7 +116,7 @@ export default class Ink {
 			this.patchConsole();
 		}
 
-		if (!isInCi) {
+		if (!this.isCi) {
 			options.stdout.on('resize', this.resized);
 
 			this.unsubscribeResize = () => {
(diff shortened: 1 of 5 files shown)


sindresorhus:
I wonder if it would be better to make the fix more general and instead add a `renderMode?: 'auto' | 'live' | 'static'`.

allenhutchison:
I'm open to working on that, but can you tell me more about what you have in mind? I got started here because we were getting bug reports on Gemini CLI for users who have CI* env variables set, and the TUI just wouldn't load.

sindresorhus:
It's the same as yours, just better naming. The decision should not be whether to enable CI detection or not, but rather the effect the CI detection has.

```
render(<App />, {ci: false}); // Force interactive mode
render(<App />, {ci: true});  // Force static mode
render(<App />);  // Auto (default)
```

```
render(<App />, {renderMode: 'live'});   // Force interactive mode
render(<App />, {renderMode: 'static'}); // Force static mode  
render(<App />, {renderMode: 'auto'});   // Auto (default)
```

`live` could also be called `interactive`.

sindresorhus:
The actual issue though will be resolved in the next update:

- https://github.com/sindresorhus/is-in-ci/releases/tag/v2.0.0
- https://github.com/vadimdemedes/ink/commit/7276f28630956af6b197acb351bf4ee257232270

PR #744, opened 2025-08-13
Fix measure text bug reading text "constructor"

### LINKED ISSUE #743: Ink incorrectly measures the length of the word 'constructor' and other text strings that exactly match members of Object
Ink incorrectly measures the length of the word 'constructor' and other text strings that exactly match members of Object

This lead to the following issue reported on the Gemini CLI Ink application. https://github.com/google-gemini/gemini-cli/issues/4953

The underlying issue is that measure text in the Ink repo uses an Object as a cache for measuring text rather than using Map and didn't use the appropriate JavaScript helpers to gracefully handle collisions with names such as 'constructor'.

jacob314:
This fixes #743 by using a `Map` instead of an `Object` avoiding the bug reading the cached value for strings such as "constructor" that are properties of `Object`.

I've added a regression test that fails without the change.
As described in #743, fixing this is important for Gemini CLI.

### DIFF
--- a/src/measure-text.ts
+++ b/src/measure-text.ts
@@ -1,6 +1,6 @@
 import widestLine from 'widest-line';
 
-const cache: Record<string, Output> = {};
+const cache = new Map<string, Output>();
 
 type Output = {
 	width: number;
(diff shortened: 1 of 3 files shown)


jacob314:
Thanks for the quick approval! Really appreciate that.

PR #751, opened 2025-08-21
Use incremental rendering for output

amateurhuman:
## Summary

- Implemented incremental rendering in log-update.ts to eliminate terminal flickering
- Improved performance by only updating changed lines instead of full screen redraws

## Problem

The previous implementation would erase all lines and redraw the entire output on every update, causing visible flickering in terminals (specifically iTerm2). This was particularly noticeable with frequent updates or large outputs.

Reported a number of times and a few fixes have been attempted https://github.com/vadimdemedes/ink/issues/359, https://github.com/vadimdemedes/ink/pull/413, https://github.com/vadimdemedes/ink/pull/708, https://github.com/anthropics/claude-code/issues/1913

## Solution

Implemented a line-by-line diffing algorithm that:

  1. Compares new output with previous output to find changed lines
  2. Uses ANSI cursor movement to navigate directly to changed lines
  3. Updates only the lines that have actually changed
  4. Added tests for the new functionality

## Impact

  - Performance: Reduces terminal I/O operations, especially for large outputs
  - User Experience: Eliminates visible flickering during updates
  - Maintains full backward compatibility with existing API

### Flickering

https://github.com/user-attachments/assets/2b50d0a4-59c5-408f-a579-5406524f4f3d

### With patch

https://github.com/user-attachments/assets/06421fd2-2c04-4a0a-b3a2-1d17813ea692


sindresorhus:
CI is failing

sindresorhus:
If possible, it would be nice to separate the diffing logic into a separate function, for readability.

amateurhuman:
> If possible, it would be nice to separate the diffing logic into a separate function, for readability

Sure thing.

> CI is failing

I'm getting some flaky tests running `npm test` on `master`, sometimes 3-4 tests fail, other times 12-14 fail. Got the tests for `log-update.tsx` passing locally. Any insight into the flaky tests in master @sindresorhus ?

ali-aljufairi:
@amateurhuman is there any update from your side ?

amateurhuman:
@ali-aljufairi still planning on getting this in, but ran into a new issue when rebasing on the latest so need to hunt down that problem and wrap up updating the tests

amateurhuman:
@sindresorhus this is one of the tests that fails with the new incremental rendering because its pretty tightly coupled with the implementation 

```
test.serial('erase screen where state changes', async t => {
	const ps = term('erase-with-state-change', ['4']);
	await ps.waitForExit();

	const secondFrame = ps.output.split(ansiEscapes.eraseLines(3))[1];

	for (const letter of ['A', 'B', 'C']) {
		t.false(secondFrame?.includes(letter));
	}
});
```

I was thinking that the test should really test the output of the render rather than how it renders, something like

```
test.serial('erase screen where state changes', async t => {
	const stdout = createStdout();
	stdout.rows = 4; // Match the fixture's viewport size

	render(<Erase />, {stdout});
	await delay(100); // Wait for the state change

	// After state change, the final render calls should not contain A, B, C
	// Check the last few write calls (after the erase operations)
	// eslint-disable-next-line @typescript-eslint/no-unsafe-call
	const recentWrites = (stdout.write as any)
		.getCalls()
		.slice(-5) // Get last 5 calls
		.map((call: any) => call.args[0] as string)
		.join('') as string;

	const recentContent = stripAnsi(recentWrites);

	for (const letter of ['A', 'B', 'C']) {
		t.false(recentContent.includes(letter));
	}
});
```

I can certainly update the test to reflect the new`cursorUp` movement and `eraselines` behavior if you'd rather keep the tests in that style. Any thoughts?

ali-aljufairi:
@amateurhuman Any updates from your side

sindresorhus:
Closing in favor of #781

PR #761, opened 2025-09-01
Update readme.md

CoryMConway:
Adding a link in the "Who's using ink?" to a Shell, Python, and JS script manager that dyanmically builds an interactice cli off of directory structure


sindresorhus:
Thanks for contributing, but for this list we are looking for more complete uses of Ink, not just a list picker, that also has more traction.

PR #769, opened 2025-09-14
Fix extra newline in full-screen mode

### LINKED ISSUE #752: Ink 6 adds a blank row at the bottom in full-screen mode (regression vs 5.x)
### Summary
When rendering a full-terminal-height layout, Ink 6.x appends a newline in the full-screen render path, causing the terminal to scroll by one line when the cursor is on the last row. This makes the last row unusable for dynamic content and makes it hard (or impossible) to build a true full-screen app that uses the bottom row in Ink v6. Ink 5.2.1 did not append this newline and allowed using the last row.

### Environment
- OS: Linux (tested on linux 6.12.43-dist)
- Node: v22.13.1
- Terminal: standard TTY
- Packages tested:
  - ink 5.2.1 (works as expected)
  - ink 6.0.0 / 6.2.2 (reproducible)
  - react 19.1.1 (also reproduced with React 18.x)

### Expected behavior
- In full-screen rendering (when output height >= terminal rows), Ink should not append a newline per frame, so that applications can render dynamic content on the last screen row without causing a scroll.

### Actual behavior
- In Ink 6, full-screen frames append a newline. If the prompt is already at the bottom row, each frame scrolls the terminal by one line, leaving a blank row below the UI.

### Minimal repro
This file is runnable without JSX transpilation and requires only `ink` and `react` dependencies.

```bash
npm install ink@6 react@19
node index-fullscreen.js
```

Repro source (index-fullscreen.js):

```js
#!/usr/bin/env node
import React from 'react';
import {render, Box, Text, useStdout, useApp} from 'ink';

const FullScreen = () => {
  const {exit} = useApp();
  const {stdout} = useStdout();
  const [rows, setRows] = React.useState(stdout.rows || 24);
  const [count, setCount] = React.useState(0);

  React.useEffect(() => {
    const onResize = () => setRows(stdout.rows || 24);
    stdout.on?.('resize', onResize);
    return () => stdout.off?.('resize', onResize);
  }, [stdout]);

  React.useEffect(() => {
    if (count > 5) {
      exit();
    }
  }, [exit, count]);

  React.useEffect(() => {
    const id = setInterval(() => setCount(c => c + 1), 1000);
    return () => clearInterval(id);
  }, []);

  // Force the root to occupy exactly terminal rows
  return (
    React.createElement(Box, {height: rows, flexDirection: 'column'},
      React.createElement(Box, {flexGrow: 1},
        React.createElement(Text, null, 'Full-screen path: main area above, bottom line below')
      ),
      React.createElement(Text, null, `Bottom line (should be usable). Count: ${count}`)
    )
  );
};

async function main() {
  const useAltScreen = process.argv.slice(2).includes('-a');

  if (useAltScreen) {
    process.stdout.write("\x1b[?1049h");
  }
  const {waitUntilExit} = render(React.createElement(FullScreen));
  await waitUntilExit();
  if (useAltScreen) {
    process.stdout.write("\x1b[?1049l");
  }
}

main();
```

How to observe the issue:
1. Run `node index-fullscreen.js`.
2. Observe that a blank line appears beneath the bottom `<Text>`, i.e. output scrolls by one line.
3. Repeat with `ink@5.2.1` — the last row is usable and no extra blank line is added.

### Notes
- Using an alternate screen (`\x1b[?1049h`/`\x1b[?1049l`) avoids leaving artifacts on the main screen, but doesn’t change newline semantics inside the alt screen.
- The minimal repro avoids JSX so it runs with just Ink and React and no transpilation.

### Suspected culprit
- Commit 8f2ad8f6e4091a8faa5ea90b5b05274d21e4e8f9 appears related.

yukukotani:
## Summary
Fixed an issue where an extra newline was added at the bottom of the screen in full-screen mode.

Fix #752, Related https://github.com/vadimdemedes/ink/pull/717

## Changes
- Removed trailing newline from full-screen output
- Allow the bottom line of the terminal to be usable

## Test
- Added test to verify that the bottom line is correctly usable in full-screen mode

### DIFF
--- a/src/ink.tsx
+++ b/src/ink.tsx
@@ -238,7 +238,7 @@ export default class Ink {
 
 		if (this.lastOutputHeight >= this.options.stdout.rows) {
 			this.options.stdout.write(
-				ansiEscapes.clearTerminal + this.fullStaticOutput + output + '\n',
+				ansiEscapes.clearTerminal + this.fullStaticOutput + output,
 			);
 			this.lastOutput = output;
 			this.lastOutputHeight = outputHeight;
--- a/test/fixtures/fullscreen-no-extra-newline.tsx
+++ b/test/fixtures/fullscreen-no-extra-newline.tsx
@@ -1,0 +1,33 @@
+import process from 'node:process';
+import React, {useEffect} from 'react';
+import {Box, Text, render, useApp} from '../../src/index.js';
+
+function FullScreen() {
+	const {exit} = useApp();
+
+	useEffect(() => {
+		// Exit after first render to check the output
+		const timer = setTimeout(() => {
+			exit();
+		}, 100);
+
+		return () => clearTimeout(timer);
+	}, [exit]);
+
+	// Force the root to occupy exactly terminal rows
+	const rows = Number(process.argv[2]) || 5;
+
+	return (
+		<Box height={rows} flexDirection="column">
+			<Box flexGrow={1}>
+				<Text>Full-screen: top</Text>
+			</Box>
+			<Text>Bottom line (should be usable)</Text>
+		</Box>
+	);
+}
+
+// Set terminal size from argument
+process.stdout.rows = Number(process.argv[2]) || 5;
+
+render(<FullScreen />);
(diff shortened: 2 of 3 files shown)


sindresorhus:
CI is failing

yukukotani:
@sindresorhus sorry, fixed

PR #771, opened 2025-09-14
chore: pin deps and add lockfile

re-taro:
## Summary

Updated dependency versions to fixed releases and generated a lockfile to ensure reproducibility across environments.

## Changes

- Migrated dependencies and devDependencies from caret (^) ranges to exact versions
- Bumped multiple packages to their latest compatible releases
- Generated and committed lockfile to guarantee consistent installs

Follow-up Question 
---

to @sindresorhus

Why are we not using a lockfile or fixed versions in your repositories, and instead relying on caret (^) ranges for dependencies?
Given that Ink is widely used across many projects, I assume this approach provides flexibility for downstream consumers, but I’d like to confirm if that’s the primary reason.


sindresorhus:
https://github.com/sindresorhus/ama/issues/479#issuecomment-310661514

sindresorhus:
No need to lock versions. If consumers want that, they can use a lockfile.

re-taro:
@sindresorhus 

I saw your comment.

> If you use a lockfile for packages, your local dependency tree will not match the dependency tree of users having your package as a dependency.

This is correct, but the reverse is not necessarily true. The absence of a lock file does not guarantee that the local tree matches the user's tree. Therefore, deleting the lock file does not resolve the mismatch issue.

At the very least, the lock file ensures that co-developers have the same tree.

Generally, there is no way to guarantee that users have an identical dependency tree (except when specifying exact versions in package.json). However, keeping the lock file up to date allows you to match the configuration during a new installation.

PR #778, opened 2025-09-20
✨ feat: add border background colors for Box component

miunau:
Added background colors and dimming options for Box borders. Also added tests and an example.

### DIFF
--- a/examples/border-backgrounds/border-backgrounds.tsx
+++ b/examples/border-backgrounds/border-backgrounds.tsx
@@ -38,13 +38,13 @@ function BorderBackgrounds() {
 			</Box>
 
 			<Box
-				borderBackgroundDimColor
+				borderDimColor
 				borderBackgroundColor="rgb(128, 0, 128)"
 				borderStyle="classic"
 				borderColor="white"
 				padding={1}
 			>
-				<Text>Box with RGB purple dimmed background on border</Text>
+				<Text>Box with dimmed RGB purple background on border</Text>
 			</Box>
 
 			<Box
--- a/examples/border-backgrounds/border-backgrounds.tsx
+++ b/examples/border-backgrounds/border-backgrounds.tsx
@@ -1,0 +1,62 @@
+import React from 'react';
+import {render, Box, Text} from '../../src/index.js';
+
+function BorderBackgrounds() {
+	return (
+		<Box flexDirection="column" gap={1}>
+			<Box
+				borderStyle="round"
+				borderColor="white"
+				borderBackgroundColor="blue"
+				padding={1}
+			>
+				<Text>Box with blue background on white border</Text>
+			</Box>
+
+			<Box
+				borderStyle="single"
+				borderColor="black"
+				borderBackgroundColor="yellow"
+				padding={1}
+			>
+				<Text>Box with yellow background on black border</Text>
+			</Box>
+
+			<Box
+				borderStyle="double"
+				borderTopColor="red"
+				borderTopBackgroundColor="green"
+				borderBottomColor="blue"
+				borderBottomBackgroundColor="yellow"
+				borderLeftColor="cyan"
+				borderLeftBackgroundColor="magenta"
+				borderRightColor="white"
+				borderRightBackgroundColor="red"
+				padding={1}
+			>
+				<Text>Box with different colors per side</Text>
+			</Box>
+
+			<Box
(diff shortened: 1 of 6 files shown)


miunau:
need this for the full msdos edit experience :) lmk if anything needs tuning

sindresorhus:
AI review:

---

### Critical correctness

* **Vertical border coloring bleeds across lines.**
  You build the multi-line `leftBorder`/`rightBorder` strings first and then wrap them once with foreground/background styles. ANSI SGR persists across `\n`, so the style will leak into the content lines until a reset—unlike the previous per-line coloring. Colorize **each line** (char + reset + `\n`), not the whole multi-line string. Example fix:

  ```ts
  const mkSide = (char: string, fg?: Color, bg?: Color) => {
  	const one = colorize(colorize(char, fg, 'foreground'), bg, 'background');
  	return (one + '\n').repeat(verticalBorderHeight);
  };
  const leftBorder = showLeftBorder ? mkSide(box.left, leftBorderColor, leftBorderBackgroundColor) : '';
  const rightBorder = showRightBorder ? mkSide(box.right, rightBorderColor, rightBorderBackgroundColor) : '';
  ```

  Right now you do:

  ```ts
  let left = `\n${box.left}`.repeat(verticalBorderHeight);
  left = colorize(left, leftBorderColor, 'foreground');
  left = colorize(left, leftBorderBackgroundColor, 'background');
  ```

  which leaks. Same for `rightBorder`.

### Type issues

* **`LiteralUnion` without generics.**
  `borderBackgroundColor?: LiteralUnion;` is invalid; `LiteralUnion<T, U>` requires both params. Use the same type as `<Text backgroundColor>`, i.e. `LiteralUnion<BackgroundColorName, string>` (or `ForegroundColorName` if that’s what `Text` uses). Do the same for the four per-side background props if any are missing generics. 

### Tests—gaps that miss real regressions

* **No test for bleed on vertical borders.**
  Add a test where a `Box` with `borderLeftBackgroundColor` wraps `<Text>content</Text>` and assert the inner text is **not** wrapped in background codes on those lines. Today you only check that background SGR exists, not that resets occur per line. 
* **No test combining foreground + background + dim.**
  Add a case asserting both FG (e.g. `\x1b[31m`) and BG (e.g. `\x1b[46m`) codes are present and that `dim` doesn’t cancel them (i.e. proper nesting). 

### Rendering details

* **Dim logic name vs effect.**
  `border*BackgroundDimColor` suggests only dimming the background, but ANSI “dim” affects the whole sequence. Either document that or rename to `border*Dim` for backgrounds to match behavior. Also document the four per-side “background dim” props if you’re keeping them—README currently lacks them. 
* **Redundant conditionals.**
  `if (showLeftBorder && leftBorder)` and `if (showRightBorder && rightBorder)` are redundant since `leftBorder/rightBorder` are already empty when hidden. Drop the boolean flags in those `if`s to reduce branching noise. 

### Docs & examples

* **Anchor casing.**
  README links to `#backgroundColor`. GitHub anchors are lowercase; use `#backgroundcolor` to be safe/consistent with the rest of the file. 
* **Document shorthand precedence.**
  You implemented per-side props with fallback to `borderBackgroundColor`. Make that explicit in the README under each new prop block. 

### Implementation polish

* **Duplicate colorize passes.**
  You call `colorize` twice for top/bottom (FG then BG). That’s fine if `colorize` uses granular resets (39/49). If it uses full reset (`0`), nesting will drop the outer style. Confirm `colorize` behavior or switch to a single combined pass in a helper that composes open/close codes deterministically. 
* **Factor common “apply side styles” helper.**
  Top/bottom/left/right share the same “fg→bg→dim” pipeline. Extract it to avoid future divergence and mistakes like the vertical bleed above.

miunau:
My local tests are passing, not sure why the workflow didn't pass. I removed the background dim property as I was confused as to how it's applied.

sindresorhus:
I few remaining things:

- A few of the new tests are currently invalid because they render raw strings inside `<Box>` instead of `<Text>`. Ink throws before those assertions run, so those tests do not actually cover border background behavior yet.
- The vertical-bleed test is not testing a real multiline case right now. The current layout renders both text nodes on one line, so the second-line assertion fails and the bleed scenario is not actually verified.

sindresorhus:
https://github.com/vadimdemedes/ink/pull/906

PR #784, opened 2025-10-06
Fix string-width miscalculation for single-cell UI symbols

okets:
## Summary

Fixes an issue where common UI symbols (▶, ◀, ⚠, ℹ) were being incorrectly measured as width 2 by `string-width`, causing layout breaks in fixed-width containers.

## Problem

The `string-width` library uses the East Asian Width property, which marks certain common UI symbols as "Ambiguous". This causes them to be treated as wide characters (width 2) even though they display as single-width (width 1) in modern terminals.

When these characters are used in Ink components, they cause:
- Incorrect text alignment
- Layout overflow in fixed-width containers
- Visual artifacts from spacing calculations

## Solution

Introduced a `SINGLE_WIDTH_OVERRIDES` set that explicitly maps characters incorrectly measured by `string-width` to their correct width (1). The width calculation now checks this override map before falling back to `string-width`.

## Changes

- Added `SINGLE_WIDTH_OVERRIDES` constant with verified single-width characters
- Modified character width calculation in `src/output.ts:227` to check overrides first
- Added comprehensive documentation explaining the issue and verification criteria

## Test Plan

- ✅ Verified each character displays as single-width in modern terminals
- ✅ Confirmed `string-width` incorrectly measures these as width 2
- ✅ Tested that the fix prevents layout breaks in fixed-width containers

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
 		"signal-exit": "^3.0.7",
 		"slice-ansi": "^7.1.0",
 		"stack-utils": "^2.0.6",
-		"string-width": "^7.2.0",
+		"string-width": "^8.1.0",
 		"type-fest": "^4.27.0",
 		"widest-line": "^5.0.0",
 		"wrap-ansi": "^9.0.0",


okets:
I just went over to fix the problem at the source (string-width library)
I figured out that in string-width version 7.2.0 (ink uses this), the bug exists
But in string-width version 8.1.0 (latest) , the bug is fixed.

solving this bug could be as easy as updating string-width to version 8.1.0.

wu-json:
> I just went over to fix the problem at the source (string-width library) I figured out that in string-width version 7.2.0 (ink uses this), the bug exists But in string-width version 8.1.0 (latest) , the bug is fixed.
> 
> solving this bug could be as easy as updating string-width to version 8.1.0.

Also experiencing this right now. @okets Is that upgrade something you want to take on? If not I can also help out.

okets:
@wu-json I just pushed my changes. waiting for someone to appove the PR.

LilaRest:
I tried patching Ink locally to use `"string-width": "^8.1.0"`, but it didn't seem to fix the issue for me.
I still observe layout shifts with those symbol: `⬆` and `✘`.

sindresorhus:
https://github.com/vadimdemedes/ink/blob/9b21b24794230a574083c3f477c4518318805786/package.json#L64

PR #795, opened 2025-10-16
Add onRender hook

shishu314:
**Problem**
---
We are trying to see how long frames take to render in Gemini CLI and would like to make changes upstream in ink to allow access to this information. 

**Changes**
---
Add an option to listen for rendering time, this should not break any existing code.

**Testing Plans**
---
1. Run the test to make sure that the callback function is called on either `render` or `re-render`.

### DIFF
--- a/src/render.ts
+++ b/src/render.ts
@@ -11,23 +11,27 @@ export type RenderOptions = {
 	@default process.stdout
 	*/
 	stdout?: NodeJS.WriteStream;
+
 	/**
 	Input stream where app will listen for input.
 
 	@default process.stdin
 	*/
 	stdin?: NodeJS.ReadStream;
+
 	/**
 	Error stream.
 	@default process.stderr
 	*/
 	stderr?: NodeJS.WriteStream;
+
 	/**
 	If true, each update will be rendered as separate output, without replacing the previous one.
 
 	@default false
 	*/
 	debug?: boolean;
+
 	/**
 	Configure whether Ink should listen for Ctrl+C keyboard input and exit the app. This is needed in case `process.stdin` is in raw mode, because then Ctrl+C is ignored by default and the process is expected to handle it manually.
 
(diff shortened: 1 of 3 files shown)


sindresorhus:
Your `onRender` callback runs only for manual `render()`/`rerender()` calls. Internal React updates call `rootNode.onRender` → `Ink.onRender()`, which never invokes `options.onRender`.

src/render.ts
```diff
@@ -42,6 +42,11 @@ export type RenderOptions = {
 	*/
 	patchConsole?: boolean;
 
+	/**
+	Function to call after each render.
+	*/
+	onRender?: (renderTime: number) => void;
```
sindresorhus: I think it should be an object with renderTime property in case we want to add more things in the future.

src/ink.tsx
```diff
@@ -280,12 +279,17 @@ export default class Ink {
 			</AccessibilityContext.Provider>
 		);
 
+		const start = Date.now();
```
sindresorhus: Use `performance.now();`

shishu314:
> Your `onRender` callback runs only for manual `render()`/`rerender()` calls. Internal React updates call `rootNode.onRender` → `Ink.onRender()`, which never invokes `options.onRender`.

Addressed this and added the callback into `Ink.onRender()`

test/render.tsx
```diff
@@ -273,3 +273,48 @@ test.serial('throttle renders to maxFps', t => {
 		clock.uninstall();
 	}
 });
+
+test.serial('outputs renderTime when onRender is passed', t => {
```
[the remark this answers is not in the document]
shishu314: I added a case to check for internal state updates when stdin emits. However it seems like `onRender` only triggers once and does not advance the clock (see assertion). I am not sure why this is the case (not too familiar with react lifecycles but it seems like it should re-render). Any pointers on why this is the case?

PR #803, opened 2025-10-26
feat: add IME cursor positioning support

takeru:
## Summary

Implements IME (Input Method Editor) cursor positioning to fix the issue where IME candidate windows appear at the end of output instead of at the actual input position when using CJK input methods.

## Problem

When using Ink applications with CJK input methods (Japanese, Chinese, Korean), the IME candidate window appears at the wrong location because the terminal cursor stays at the end of the output instead of the actual input position.

## Solution

This PR adds an `enableImeCursor` option that positions the terminal cursor at the actual input location using:

1. **Cursor marker detection** (`src/cursor-marker.ts`)
   - Detect cursor position marker in rendered output
   - Calculate cursor position in multi-line text with proper CJK character width handling

2. **Cursor positioning** (`src/log-update.ts`)
   - Use relative cursor movement (ANSI A/B/C/D sequences)
   - Cursor save/restore for stable rendering

3. **API addition** (`src/ink.tsx`, `src/render.ts`, `src/components/App.tsx`)
   - Add `enableImeCursor` option (default: `false` for backward compatibility)
   - Conditionally control cursor visibility

## Usage

To enable IME cursor positioning in your Ink application:

### 1. Enable the option when rendering

```tsx
import {render} from 'ink';

render(<YourApp />, {
  enableImeCursor: true
});
```

### 2. Insert cursor marker at the input position

```tsx
import {Text} from 'ink';
import {CURSOR_MARKER} from 'ink/cursor-marker';

function TextInput({value, cursorPosition}) {
  const before = value.slice(0, cursorPosition);
  const after = value.slice(cursorPosition);
  
  return (
    <Text>
      {before}
      {CURSOR_MARKER}
      <Text inverse>{value[cursorPosition] || ' '}</Text>
      {after}
    </Text>
  );
}
```

The `CURSOR_MARKER` should be placed immediately before the visual cursor position. The terminal cursor will be positioned at this location.

## Demo Application

Interactive demo showing IME cursor positioning with multi-line text editing:
```bash
npx tsx examples/multiline-edit-demo/multiline-edit-demo.tsx
```

## Key Features

- ✅ Opt-in feature (backward compatible, default: disabled)
- ✅ Proper multi-line text handling
- ✅ Full-width CJK character support (2 cells per character)
- ✅ Relative cursor movement only (no terminal queries needed)

## Test Plan

Run the demo application and verify:
- Terminal cursor appears at the input position (white █ character)
- IME candidate window appears at the cursor location
- Text wrapping and cursor movement work correctly
- Multi-line editing works as expected

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/log-update.ts
+++ b/src/log-update.ts
@@ -41,16 +41,8 @@ const create = (stream: Writable, {showCursor = false} = {}): LogUpdate => {
 		// Detect and remove cursor marker
 		const {cleaned, position} = findAndRemoveMarker(str);
 
-		// When cursor is visible, remove any trailing newlines to keep cursor on input line
-		// When cursor is hidden, ensure there's a trailing newline
-		let output: string;
-		if (showCursor) {
-			// Remove trailing newlines when cursor is visible
-			output = cleaned.replace(/\n+$/, '');
-		} else {
-			// Ensure trailing newline when cursor is hidden
-			output = cleaned.endsWith('\n') ? cleaned : cleaned + '\n';
-		}
+		// Remove trailing newlines when cursor is visible
+		const output = cleaned.replace(/\n+$/, '');
 
 		// Check if both output AND cursor position are unchanged
 		const markerPositionChanged =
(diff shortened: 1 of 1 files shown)


sindresorhus:
You will have to find some humans to review this code. I don't have time to review vibe coded PRs. Claude Code is not known for producing bug-free code, especially not for complex things like this.

I also recommend getting ChatGPT 5 Thinking (with extended thinking) to review this. It's much smarter.

takeru:
Thank you for your reply.
I’ve reviewed it carefully. I don’t fully understand ink, so I can’t say for sure whether this code is perfect, but I’ve gone through all of the main code (though I only skimmed the demo test code).

This is quite a complex change, but it’s an important one for people who need to use an IME. I realize it’s difficult for those who don’t rely on IME to fully understand the importance, so I’ve been thinking for a while about how best to explain it. In the end, I decided to go ahead and open the PR so you could take a look.

Do you think there’s a possibility that this kind of approach could be accepted?
If so, I’d like to continue refining it further.

sindresorhus:
> Do you think there’s a possibility that this kind of approach could be accepted?

I don't want to say never, but it would have to be high-quality human-reviewed code and there would have to be someone that commits to maintaining it.

PR #811, opened 2025-11-02
Fix VSCode extension compatibility by eliminating top-level awaits

asinghvi17:
## Summary

This PR fixes the "Cannot read properties of null (reading 'useState')" error that occurs when using Ink in VSCode extensions.

## Problem

VSCode extensions and certain bundlers don't support top-level awaits, which causes:
1. Module loading failures
2. React dispatcher becoming null, breaking hooks

## Solution

- ✅ Eliminated all top-level awaits in production code
- ✅ Migrated from `yoga-layout` to `yoga-layout/load` (which doesn't use top-level await)
- ✅ Made Yoga initialization graceful and non-blocking
- ✅ Added fallback handling when Yoga isn't loaded yet

## Key Changes

1. **Replaced devtools top-level await** in `reconciler.ts` with `.then()/.catch()` pattern
2. **Created `yoga-init.ts`** module for managing Yoga initialization asynchronously
3. **Modified `createNode`** to handle uninitialized Yoga without throwing
4. **Updated `calculateLayout`** to skip when yogaNode is unavailable
5. **Added automatic re-render** when Yoga becomes available

## Testing

- Added comprehensive tests for Yoga initialization (`test/yoga-init.tsx`)
- Added CommonJS compatibility tests (`test/commonjs-compatibility.tsx`)
- Verified the fix works with React hooks in VSCode extensions
- All existing tests pass

## Compatibility

- ✅ Backward compatible
- ✅ Works in VSCode extensions
- ✅ Works with CommonJS
- ✅ No breaking changes

Fixes issues with VSCode extension integration where top-level awaits are not supported.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,574 @@
+# Ink Codebase Guide
+
+**Project Name:** Ink  
+**Version:** 6.4.0  
+**Type:** React renderer/library for CLI applications  
+**Language:** TypeScript (with some JavaScript utility files)  
+**Total Lines of Code:** ~4,600 (src)  
+**Last Commit:** 6.4.0 (recent updates include onRender hook addition)
+
+---
+
+## Quick Summary
+
+Ink is a React-compatible UI framework for building interactive command-line interfaces. It provides React developers with familiar component-based and hooks-based APIs to create rich terminal UIs with support for flexbox layouts, styling, focus management, and accessibility.
+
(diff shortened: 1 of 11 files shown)


sindresorhus:
I appreciate the effort, but top-level await is a standard JS feature. We are not going to do a lot of churn and complication just to support tools not supporting TLA. I recommend spending your effort trying to get VSCode and bundlers to support TLA.

PR #814, opened 2025-11-06
Fix text wrapping.

jacob314:
Fixes:
https://github.com/google-gemini/gemini-cli/issues/12653

To verify run 
`npm run example examples/text-wrap/index.ts`
and view the output at various widths to make sure the option is good.


sindresorhus:
@jacob314 Can you please stop submitting to the wrong repo. This is the 7th time you do this.

PR #825, opened 2025-11-14
perf: reuse `lastOutput` to reduce memory pressure

HomyeeKing:
there're two lastOutput stored which may cause OOM in some case.


src/log-update.ts
```diff
@@ -10,37 +10,37 @@ export type LogUpdate = {
...
-		previousOutput = output;
+		inkInstance.lastOutput = output;
 		stream.write(ansiEscapes.eraseLines(previousLineCount) + output);
 		previousLineCount = output.split('\n').length;
 	};
 
 	render.clear = () => {
 		stream.write(ansiEscapes.eraseLines(previousLineCount));
-		previousOutput = '';
 		previousLineCount = 0;
```
HomyeeKing: Won't clear inkInstance.lastOutput since it will trigger component rerender. Same for the incremental render part

PR #831, opened 2025-11-23
Add activeId property to useFocusManager hook

SubZtep:
## Summary

  - Exposes `activeId` property from `FocusContext` in `useFocusManager` hook to get the `id` of the currently focused component
  - Returns `undefined` when no component has focus
  - Added test coverage for the new property
  - Updated documentation in root readme

## Use case

This allows components to programmatically check which component currently has focus without needing to track focus state manually. Useful for displaying focus state in UI or conditionally rendering based on what's focused.

## Test plan
  - Added new test _"the focus helper reports whichever id currently holds focus"_ in test/focus.tsx
  - Test verifies that `activeId` is defined when a component is focused
  - Test verifies that `activeId` becomes `undefined` when focus is cleared with Esc key

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -1962,6 +1962,22 @@ const Example = () => {
 };
 ```
 
+#### activeId
+
+Type: `string | undefined`
+
+[`id`](#id) of the currently focused component. If there's no active focus on component right now, the value is `undefined`.
+
+```js
+import {useFocusManager} from 'ink';
+
+const Example = () => {
+	const {activeId} = useFocusManager();
+
+	return <Text>Focused component ID: {activeId ?? "-"}</Text>;
+};
+```
+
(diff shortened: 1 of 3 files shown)


sindresorhus:
I don't review AI generated PRs, so you will have to find someone to review this.

SubZtep:
> I don't review AI generated PRs, so you will have to find someone to review this.

The main change is only [one line insert](https://github.com/SubZtep/ink/blob/710b6458f769960746c3b8d05f196588e14c701e/src/hooks/use-focus-manager.ts#L48). The PR text is really based on an AI suggestion, but I think it make sense tho. Not sure who should I find. :finnadie:

SubZtep:
It's possible to achieve the same result without exposing the context variable.

PR #835, opened 2025-12-05
perf: use caches to speed up output generation

andyburris:
Really only noticeable for huge terminals so hard to use the current benchmarks, but anecdotally on my 150x200 terminal it's sped renders up from ~40ms to ~9ms, so almost 5x faster. Presumably it'd be more of a speedup the larger the terminal gets.

All tests (that were passing on current `master`) are passing.

### DIFF
--- a/src/output.ts
+++ b/src/output.ts
@@ -54,7 +54,7 @@ class OutputCaches {
 
 	getStyledChars(line: string): StyledChar[] {
 		let cached = this.styledChars.get(line);
-		if (!cached) {
+		if (cached === undefined) {
 			cached = styledCharsFromTokens(tokenize(line));
 			this.styledChars.set(line, cached);
 		}
@@ -64,7 +64,7 @@ class OutputCaches {
 
 	getStringWidth(text: string): number {
 		let cached = this.widths.get(text);
-		if (!cached) {
+		if (cached === undefined) {
 			cached = stringWidth(text);
 			this.widths.set(text, cached);
 		}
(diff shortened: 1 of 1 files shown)


sindresorhus:
`Output` is recreated on every render, so it's unclear to me why this actually improves performance, unless the output has duplicated lines.

andyburris:
I think because there's a lot of redundant processing even within a single output frame. For example:
- [`widestLine`](https://github.com/vadimdemedes/ink/blob/45ed972ddfd3b508ed6397f9f844d076454a891c/src/output.ts#L145) is internally calling `stringWidth` on every line, just to do that again [manually later on
](https://github.com/vadimdemedes/ink/blob/45ed972ddfd3b508ed6397f9f844d076454a891c/src/output.ts#L163)
- more importantly, `stringWidth` is getting re-run on [every individual character](https://github.com/vadimdemedes/ink/blob/45ed972ddfd3b508ed6397f9f844d076454a891c/src/output.ts#L208)

sindresorhus:
Some things I noticed:

- The memoization uses `if (!cached)` on numeric widths. A valid width of `0` is falsy, so cache hits for empty/ANSI-only strings and all-empty blocks become misses and get recomputed every time.
- That means the caceh does not help for those cases and can hurt perf in exactly the workloads this aims to speed up.

Could we switch those checks to `cached === undefined` (or `map.has(key)`) so `0` is treated as a real cached value?

andyburris:
Yes!

PR #837, opened 2025-12-08
Removed indent-string as dependency and replaced usages with built-in function

AnakinM:
The denepdency `indent-string` is extremaly minimal one, MIT licensed. I don't believe there is a point in keeping in as dependency in `node_modules` when it can easily be a function built in the codebase itself. There is not much more to maintain, it does one thing and does it well. In this case I propose to merge this PR, reducing the amount of dependencies for this project.


sindresorhus:
No thanks :)

PR #850, opened 2026-01-18
Concurrent rendering support

### LINKED ISSUE #720: Support React concurrent mode
- https://github.com/vadimdemedes/ink/pull/719
- https://github.com/vadimdemedes/ink/pull/656

It will require rewriting all the tests.

re-taro:
## Summary

Fixes #720 

   Add optional React Concurrent Rendering mode to Ink, enabling React 19's concurrent features like Suspense with async data and useTransition.

   - Add `concurrent?: boolean` option to render API
   - Use `ConcurrentRoot` instead of `LegacyRoot` when enabled
   - Integrate Scheduler for proper async scheduling
   - Add comprehensive test helpers and concurrent mode tests

   ## Changes

   ### Core Implementation
   - **src/ink.tsx**: Add concurrent option, switch between ConcurrentRoot/LegacyRoot, use updateContainer for concurrent updates
   - **src/reconciler.ts**: Add Scheduler integration (scheduleCallback, cancelCallback, shouldYield, now)
   - **src/render.ts**: Add concurrent option to RenderOptions with documentation
   - **package.json**: Add scheduler dependency

   ### Test Infrastructure
   - **test/helpers/render-to-string.ts**: Add `renderToStringAsync` for concurrent mode
   - **test/helpers/test-renderer.ts**: New file with `renderAsync`, `renderSync`, `withAct`, `waitForSuspense`

   ### Concurrent Mode Tests
   Added concurrent mode test variants for:
   - Text, Display, Gap, Borders, Background
   - Flex-direction, Margin, Padding, Width-height
   - Components, Overflow, Focus, Reconciler

   ### Examples
   - **examples/concurrent-suspense/**: Demonstrates Suspense with async data fetching
   - **examples/use-transition/**: Demonstrates useTransition for deferred updates

   ## Usage

   ```tsx
   import {render} from 'ink';

   // Enable concurrent mode
   render(<App />, { concurrent: true });

   // Now Suspense and useTransition work correctly
   ```

   ## Test Plan

   - [x] All existing tests pass (5 failures are pre-existing from master)
   - [x] New concurrent mode tests pass
   - [x] Examples run correctly
   - [x] TypeScript types are correct

   ## Related

   Continues the work from PR #656.

### DIFF
--- a/src/render.ts
+++ b/src/render.ts
@@ -83,7 +83,7 @@ export type RenderOptions = {
 	- `useTransition` and `useDeferredValue` are fully functional
 	- Updates can be interrupted for higher priority work
 
-	Note: Concurrent mode changes the timing of renders. Some tests may need to use `act()` to properly await updates.
+	Note: Concurrent mode changes the timing of renders. Some tests may need to use `act()` to properly await updates. The `concurrent` option only takes effect on the first render for a given stdout. If you need to change the rendering mode, call `unmount()` first.
 
 	@default false
 	*/


re-taro:
@sindresorhus 

The two test cases that are currently failing are failing on the latest master. I would like to ask for your judgment on whether they need to be fixed.

sindresorhus:
AI review:

- Concurrent unmount resolves `waitUntilExit()` before React has committed the unmount, so effect cleanups (stdin raw mode, input listeners) can run after `unmount()` returns. In concurrent mode updates are asynchronous and interruptible, and effect cleanups run on unmount; if you don’t flush, teardown may happen after callers assume it’s done. Proposed fix: for concurrent unmount either use a sync flush (`updateContainerSync` + `flushSyncWork`) or move `instances.delete(...)` and `resolveExitPromise()` into the `updateContainer` callback so it resolves only after commit; `flushSync` is designed to force pending work and effects to flush before continuing. `src/ink.tsx:413`
- Host config adds `scheduleMicrotask` but does not set `supportsMicrotasks: true`. React’s host config contract says `supportsMicrotasks` must be true when `scheduleMicrotask` is implemented; otherwise React assumes microtasks are unsupported. That can change scheduling semantics under concurrent rendering. `src/reconciler.ts:233-244`
- Input updates are not marked as discrete or given event priority. `useInput` wraps keypress handlers in `batchedUpdates`, but concurrent mode expects user interactions (keypresses) to be processed via `discreteUpdates` and/or `getCurrentEventPriority` so they aren’t deferred behind lower‑priority work. This can cause input lag under concurrent rendering or during transitions. `src/hooks/use-input.ts:145-196`, `src/reconciler.ts:307-317`
- `concurrent` option is sticky per `stdout`. `render()` reuses the existing instance from `instances` without reconciling option changes. If the first render used legacy mode, later calls with `concurrent: true` silently keep `LegacyRoot` and the legacy render path. That makes the option unreliable and can lead to mixed expectations. `src/render.ts:116-158`, `src/ink.tsx:115-129`

sindresorhus:
> The two test cases that are currently failing are failing on the latest master. I would like to ask for your judgment on whether they need to be fixed.

This was fixed in `master`.

re-taro:
@sindresorhus 

> Concurrent unmount resolves waitUntilExit() before React has committed the unmount, so effect cleanups (stdin raw mode, input listeners) can run after unmount() returns. In concurrent mode updates are asynchronous and interruptible, and effect cleanups run on unmount; if you don’t flush, teardown may happen after callers assume it’s done. Proposed fix: for concurrent unmount either use a sync flush (updateContainerSync + flushSyncWork) or move instances.delete(...) and resolveExitPromise() into the updateContainer callback so it resolves only after commit; flushSync is designed to force pending work and effects to flush before continuing. src/ink.tsx:413

After investigation, the current async unmount implementation for concurrent mode is correct and intentional.

  React DOM uses sync unmount (updateContainerSync + flushSyncWork) for a DOM-specific reason: React stores internal metadata on DOM nodes that
  must be cleared immediately upon unmount (see https://github.com/facebook/react/pull/22444).

  However, ink has no DOM nodes—it renders to a terminal. The synchronization concern that applies to React DOM does not apply here.
  Additionally, using sync unmount during concurrent rendering can trigger the warning: "Attempted to synchronously unmount a root while React
  was already rendering" (https://github.com/facebook/react/issues/25675).

  The async approach for concurrent mode is valid and avoids potential conflicts with in-progress renders.

> Input updates are not marked as discrete or given event priority. useInput wraps keypress handlers in batchedUpdates, but concurrent mode expects user interactions (keypresses) to be processed via discreteUpdates and/or getCurrentEventPriority so they aren’t deferred behind lower‑priority work. This can cause input lag under concurrent rendering or during transitions. src/hooks/use-input.ts:145-196, src/reconciler.ts:307-317

After investigation, switching from batchedUpdates to discreteUpdates is not necessary for ink.

  The key difference is that discreteUpdates sets DiscreteEventPriority for React's scheduler. However, this priority system is designed for
  React DOM's integration with browser events, where React can infer priority from window.event.type.

  Ink handles stdin input through its own event system, not React's synthetic event system. The getCurrentEventPriority / resolveUpdatePriority
  hooks in ink's reconciler config already return DefaultEventPriority, which is appropriate for terminal applications where all input should be
   treated uniformly.

  Furthermore, with React 18+'s automatic batching (https://github.com/reactwg/react-18/discussions/21), explicit batchedUpdates calls are
  mainly for backward compatibility. The current implementation is functionally correct.

> concurrent option is sticky per stdout. render() reuses the existing instance from instances without reconciling option changes. If the first render used legacy mode, later calls with concurrent: true silently keep LegacyRoot and the legacy render path. That makes the option unreliable and can lead to mixed expectations. src/render.ts:116-158, src/ink.tsx:115-129

This is a valid observation, but I believe throwing an error on mismatch would be overly aggressive for the typical use case.

  The render() function reusing existing instances per stdout is an intentional design for supporting re-renders (calling render() multiple
  times with different components). Changing the concurrent option mid-lifecycle is not a supported use case—the root type (LegacyRoot vs
  ConcurrentRoot) is determined at creation time and cannot be changed.

  Options to consider:

  1. Document the behavior — Add a note in the concurrent option's JSDoc that it only takes effect on the first render for a given stdout.
  2. Warn on mismatch — Log a warning (not throw) when the requested concurrent option differs from the existing instance, informing the user to
   call unmount() first.
  3. Throw an error — Strictly enforce that the option must match. This is the safest but most disruptive approach.

  I lean toward option 2 (warn on mismatch) as it provides visibility without breaking existing code. What are your thoughts?

sindresorhus:
2

sindresorhus:
Readme updates are missing.

re-taro:
@sindresorhus 

Sorry for the delay! And thank you for the review! I've addressed it, so I'd appreciate it if you could take a look!

sindresorhus:
Looks good. Thanks :)

re-taro:
Thank you so much!!!!!!

PR #851, opened 2026-01-19
feat: add IME cursor position support for CJK input

koshikawa-masato:
## Summary

This PR adds `enableImeCursor` option to support proper IME (Input Method Editor) cursor positioning for CJK (Japanese/Chinese/Korean) input in terminals.

## Problem

When using Ink-based TUI applications in terminals like Ghostty, the IME candidate window appears at the bottom-left corner of the screen instead of at the cursor position. This happens because Ink hides the terminal's real cursor and renders a visual cursor using ANSI styles, but IME systems use the terminal's real cursor position to determine where to display candidates.

## Solution

This PR introduces:

1. **`CURSOR_MARKER`** (`\u001B[999m`) - An invisible ANSI SGR sequence that can be placed in the render tree to mark the desired cursor position

2. **`enableImeCursor` render option** - When enabled, the renderer:
   - Replaces `CURSOR_MARKER` with `\u001B[s` (Save Cursor Position) during output
   - After writing output, sends `\u001B[u\u001B[?25h` (Restore + Show Cursor)
   - The terminal's real cursor moves to the marked position, allowing IME to display candidates correctly

## Usage

```tsx
import { render, Box, Text, CURSOR_MARKER } from 'ink';

function TextInput({ value, cursorPosition }) {
  const before = value.slice(0, cursorPosition);
  const after = value.slice(cursorPosition);
  
  return (
    <Text>
      {before}
      {CURSOR_MARKER}
      <Text inverse>{value[cursorPosition] || ' '}</Text>
      {after}
    </Text>
  );
}

render(<App />, { enableImeCursor: true });
```

## Changes

- `src/cursor-marker.ts` - New file with marker constant and ANSI sequence utilities
- `src/log-update.ts` - Save/restore cursor position logic
- `src/render.ts` - `enableImeCursor` option
- `src/ink.tsx` - Option propagation  
- `src/index.ts` - Export `CURSOR_MARKER`
- `test/cursor-marker.tsx` - 18 test cases

## Testing

- Tested in Ghostty terminal on macOS with Japanese input
- All existing tests pass
- 18 new tests for cursor marker functionality

## Backward Compatibility

- Feature is opt-in via `enableImeCursor: true`
- No changes to default behavior
- `CURSOR_MARKER` is invisible (SGR 999) and has no visual effect if not processed

### DIFF
--- a/examples/ime-cursor.tsx
+++ b/examples/ime-cursor.tsx
@@ -4,6 +4,7 @@
  * Demonstrates the enableImeCursor feature for proper CJK input support.
  * Run with: npm run example examples/ime-cursor.tsx
  */
+import process from 'node:process';
 import React, {useState} from 'react';
 import {render, Box, Text, useInput, CURSOR_MARKER} from '../src/index.js';
 
@@ -12,9 +13,9 @@ function TextInput({
 	onChange,
 	placeholder,
 }: {
-	value: string;
-	onChange: (value: string) => void;
-	placeholder?: string;
+	readonly value: string;
+	readonly onChange: (value: string) => void;
+	readonly placeholder?: string;
 }) {
 	const [cursorPosition, setCursorPosition] = useState(value.length);
 
@@ -36,6 +37,7 @@ function TextInput({
 				onChange(newValue);
 				setCursorPosition(cursorPosition - 1);
 			}
+
 			return;
 		}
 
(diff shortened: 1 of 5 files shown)


sindresorhus:
Going to close this, but I'm happy to consider follow up improvements to https://github.com/vadimdemedes/ink/pull/866

PR #852, opened 2026-01-19
feat: Add Kitty keyboard protocol support

### LINKED ISSUE #824: Support `kitty keyboard protocol` to enable more complex keyboard interaction
### Problem to solve

Currently I have not found a way to support some 'advanced' keyboard interactions with `useInput`.

For example, in a text input, we want to differentiate between `shift+enter` (enter a line break) and `enter` (submit). Pressing either key / key combo gives the same values:

```
[DEBUG] raw node process.stdin: { bytes: '0x0\r', length: 1, str: '\r', normalizedStr: '\\r' }
[DEBUG] Ink useInput received keypress: {
  charInput: '\r',
  charCode: 13,
  'key.return': true,
  'key.shift': false,
  'key.ctrl': false
}
```

While I can change settings in my terminal emulator to get around this, I don't want my end-users to have that burden.

### Proposed solution

Add support for `kitty keyboard protocol`: https://sw.kovidgoyal.net/kitty/keyboard-protocol/

> There are various problems with the current state of keyboard handling in terminals. They include:
> 
> * No way to use modifiers other than ctrl and alt
> * No way to reliably use multiple modifier keys, other than, shift+alt and ctrl+alt.
> * Many of the existing escape codes used to encode these events are ambiguous with different key presses mapping to the same escape code.
> * No way to handle different types of keyboard events, such as press, release or repeat
> * No reliable way to distinguish single Esc key presses from the start of a escape sequence. Currently, client programs use fragile timing related hacks for this, leading to bugs
> 
> To solve these issues and others, kitty has created a new keyboard protocol, that is backward compatible but allows applications to opt-in to support more advanced usages. The protocol is based on initial work in [fixterms](http://www.leonerd.org.uk/hacks/fixterms/), however, it corrects various issues in that proposal, listed at the [bottom of this document](https://sw.kovidgoyal.net/kitty/keyboard-protocol/#fixterms-bugs).
>
> In addition to kitty, this protocol is also implemented in:
> * The alacritty terminal
> * The ghostty terminal
> * The foot terminal
> * The iTerm2 terminal
> * The rio terminal
> * The WezTerm terminal
> * The TuiOS terminal (multiplexer)

It would be convenient if ink/`useInput` abstraction automatically handled kitty-protocol opt-in and parsing,

mattheworiordan:
Implements support for the Kitty keyboard protocol, enabling enhanced key handling in terminals that support it (Kitty, WezTerm, Ghostty, etc.) and closes #824. 

Features:
- Runtime detection of Kitty protocol support via terminal query
- Protocol enable/disable lifecycle management in App component
- Parsing of Kitty escape sequences with modifier key support
- Integration with useInput hook for seamless usage
- Falls back to standard input handling for unsupported terminals

New files:
- src/kitty-detection.ts: Protocol detection and control functions
- src/kitty-protocol.ts: Sequence parsing and type definitions
- test/kitty-protocol.tsx: Unit tests for protocol parsing
- test/fixtures/use-input-kitty.tsx: Integration test fixtures

> [!WARNING]  
> This PR has been created with [Colony](https://github.com/mattheworiordan/colony), as part of an experiment to compare Ralph vs Colony as tools to deliver code, with the latter focussed on quality landable PR-ready code. I've read the code and run the tests and I genuinely hope this PR is good enough quality to accept/land. If it is not, I apologise, but would appreciate feedback to make Colony better and make it possible for people to contribute higher quality PRs to open source proejcts in the future. Thanks for any input/feedback you have.

### DIFF
--- a/src/ink.tsx
+++ b/src/ink.tsx
@@ -473,10 +473,10 @@ export default class Ink {
 	}
 
 	/**
-	 * Detect if the terminal supports the Kitty keyboard protocol.
-	 * This sends a query sequence and waits for a response with a timeout.
-	 * If supported, enables the protocol.
-	 */
+	Detect if the terminal supports the Kitty keyboard protocol.
+	This sends a query sequence and waits for a response with a timeout.
+	If supported, enables the protocol.
+	*/
 	async detectKittyProtocol(): Promise<void> {
 		const {stdin, stdout} = this.options;
 
@@ -535,8 +535,8 @@ export default class Ink {
 	}
 
 	/**
-	 * Enable the Kitty keyboard protocol by sending the opt-in sequence.
-	 */
+	Enable the Kitty keyboard protocol by sending the opt-in sequence.
+	*/
 	enableKittyProtocol(): void {
 		if (this.isKittyProtocolEnabled || this.isUnmounted) {
 			return;
@@ -547,9 +547,9 @@ export default class Ink {
 	}
 
 	/**
-	 * Disable the Kitty keyboard protocol by sending the opt-out sequence.
-	 * This restores the terminal to its original mode.
-	 */
+	Disable the Kitty keyboard protocol by sending the opt-out sequence.
+	This restores the terminal to its original mode.
+	*/
 	disableKittyProtocol(): void {
 		if (!this.isKittyProtocolEnabled) {
 			return;


mattheworiordan:
Note:

- I could not run this locally because NodePTY does not work with the latest version of MacOS. I've bumped the dev only version so that others can run tests on Mac.
- I saw that there was a flakey test with dim+bold test, so fixed that so that we this PR reliably passes.

sindresorhus:
AI review:

- `readable` listener can close over stale `exitOnCtrlC`/`onExit` if those props ever change while raw mode stays enabled. The listener is attached once and never updated, unlike the previous class method that always read latest props via `this`. If those props are not guaranteed stable, this is a behavior regression. Consider refs for latest callbacks or reattaching when they change. `src/components/App.tsx:92` `src/components/App.tsx:115` `src/components/App.tsx:124`
- The raw-mode cleanup test mutates `process.stdin.setRawMode` and restores it only after `t.throwsAsync`. If the assertion fails, the restore never runs, potentially contaminating later tests. Use `try/finally` or `t.teardown` for restoration. `test/errors.tsx:77`

mattheworiordan:
Thanks for the review feedback! I've pushed two commits addressing the inline comments:

1. Option naming (https://github.com/mattheworiordan/ink/commit/64158d0)
  - Renamed disableKittyProtocol → useKittyProtocol with inverted logic
  - Defaults to false (opt-in for now)
  - Added TODO comments to enable by default in next major version

2. Doc comment style (https://github.com/mattheworiordan/ink/commit/7807544)
  - Removed asterisk prefixes from JSDoc continuation lines to match codebase conventions

Regarding the AI review comments:

Stale closure concern - This is a false positive. App is a class component, not a functional component. The handleReadable method accesses this.props.exitOnCtrlC at call time (line 197), which always reads the current prop values. Class methods don't create closures over stale values like functional components with hooks do.

Test cleanup at test/errors.tsx:77 - This line doesn't exist; the file only has 49 lines. This appears to be a hallucination or reference to a different version of the code.

sindresorhus:
Closing in favor of https://github.com/vadimdemedes/ink/commit/c183c53a8174fa1693e04685072973ba519d72d5 Happy to receive follow improvements to that if you notice anything that could be improved.

PR #859, opened 2026-01-23
Fix borderDimColor dimming child Text components

### LINKED ISSUE #840: `borderDimColor` dims all child `Text` components at the left edge of the `Box`
On a parent box component with a `borderStyle`, setting `borderDimColor` to true dims all `bold` child `Text` components (even ones nested in more boxes).

Interestingly it seems to only happen to `Text` components who are "touching" the left edge of the `Box`. E.g. setting `paddingLeft` to 1 or changing the alignment to center of the root or any nested `Box` makes the `Text` act normally.

Touching left edge:
Image

`paddingLeft` = 1
Image

`alignItems` = center
Image

costajohnt:
## Summary
- Updates `@alcalzone/ansi-tokenize` from `^0.2.1` to `^0.2.3`
- This fixes the issue where `borderDimColor` would incorrectly dim child `Text` components that are touching the left edge of the `Box`

## Details
Fixes #840

The bug was in the upstream `ansi-tokenize` library and has been fixed in v0.2.3:
- Issue: https://github.com/AlCalzone/ansi-tokenize/issues/42
- Fix: https://github.com/AlCalzone/ansi-tokenize/releases/tag/v0.2.3

## Test plan
- [x] Existing border tests pass (including dim border color tests)
- [x] The reproduction from #840 should now work correctly:
  ```jsx
  import React from "react";
  import { render, Box, Text } from "ink";

  render(
    <Box borderStyle={"round"} borderDimColor>
      <Text color={"blue"} bold>
        this shouldn't be dimmed
      </Text>
    </Box>
  );
  ```

---
Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/test/borders.tsx
+++ b/test/borders.tsx
@@ -65,6 +65,17 @@ test('single node - fit-content box with emojis', t => {
 	t.is(output, boxen('🌊🌊', {borderStyle: 'round'}));
 });
 
+// Issue #733: Emojis with variation selectors (FE0F) should align properly
+test('single node - fit-content box with variation selector emojis', t => {
+	const output = renderToString(
+		<Box borderStyle="round" alignSelf="flex-start">
+			<Text>🌡️⚠️✅</Text>
+		</Box>,
+	);
+
+	t.is(output, boxen('🌡️⚠️✅', {borderStyle: 'round'}));
+});
+
 test('single node - fixed width box', t => {
 	const output = renderToString(
 		<Box borderStyle="round" width={20}>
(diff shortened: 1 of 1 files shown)


sindresorhus:
Would be good to have a test for this. Otherwise, this PR has not much point as the semver specifier already matches the fixed version.

costajohnt:
Added a regression test for issue #840 that verifies:
1. Child `Text` components with styling (color + bold) retain their styles when `borderDimColor` is used
2. The border itself is correctly dimmed

The test specifically covers the edge case where styled text touches the left edge of the box, which was the problematic scenario.

PR #863, opened 2026-01-26
fix: make unmount flush pending renders and await stdout drain

### LINKED ISSUE #796: Wait for Ink to finish any pending renders before process exit
I have a custom logger class which uses Ink.

If I hit an exception anywhere in my application, the parent process immediately exits and stops Ink from going through any final render cycles that would result in logs being printed.

If I put a timeout for 10ms directly before throwing an error, Ink finishes as expected. But I don't always know where errors are going to occur.

This is the best I've come up with in my logger class:

```typescript
    private setupExitHandler() {
        const flush = async () => {
            // give Ink a bit of time to get one or two additional renders in
            await new Promise(resolve => setTimeout(resolve, 100))
        }
        process.on('uncaughtException', async error => {
            await flush()
            console.error(`${error.stack} ${error.message}`)
        })

        process.on('unhandledRejection', async reason => {
            await flush()
            const error = reason instanceof Error ? reason : new Error(String(reason))
            console.error(`${error.stack} ${error.message}`)
        })
    }
```

Is there a canonical/better way to do this?

costajohnt:
## Summary
- Makes `unmount()` flush pending throttled renders and log writes before the final render, so the last frame is always displayed
- Makes `waitUntilExit()` await the stdout drain callback, so the promise only resolves once buffered writes are actually flushed to the terminal
- Users can use `process.exitCode` + `await waitUntilExit()` instead of `process.exit()` for guaranteed final output
- No new public API surface area

## Changes
- `unmount()` now calls `this.throttledOnRender.flush()` and `this.throttledLog.flush()` before the final render
- Exit promise resolution is deferred until `stdout.writableNeedDrain` is false (or the `drain` event fires)
- Removed `flush()` public method, `src/flush.ts`, and `flush` export
- Added test verifying unmount forces pending throttled render

## Test plan
- [x] New test: "unmount forces pending throttled render" 
- [x] All existing tests pass
- [x] Linting passes

Closes #796

### DIFF
--- a/src/ink.tsx
+++ b/src/ink.tsx
@@ -454,6 +454,10 @@ export default class Ink {
 		// exit promise. For real writable streams, queue an empty write as a
 		// barrier — its callback fires only after all prior writes complete.
 		// For non-stream objects (e.g. test spies), resolve on next tick.
+		//
+		// When called from signal-exit during process shutdown (error is a
+		// number or null rather than undefined/Error), resolve synchronously
+		// because the event loop is draining and async callbacks won't fire.
 		const resolveOrReject = () => {
 			if (error instanceof Error) {
 				this.rejectExitPromise(error);
@@ -462,7 +466,11 @@ export default class Ink {
 			}
 		};
 
-		if (
+		const isProcessExiting = error !== undefined && !(error instanceof Error);
+
+		if (isProcessExiting) {
+			resolveOrReject();
+		} else if (
 			(this.options.stdout as any)._writableState !== undefined ||
 			this.options.stdout.writableLength !== undefined
 		) {


sindresorhus:
I do not think we need a public flush API. The real fix can live inside existing behavior: make unmount force the final render without throttling, and have `waitUntilExit` await the last stdout write callback so buffered writes are actually flushed. Then recommend using `process.exitCode` and `await waitUntilExit` instead of `process.exit`. This keeps the surface area unchanged and still guarantees the last frame is written.

costajohnt:
Great feedback, agreed! I've revised this PR to avoid adding a public `flush()` API. Instead:

1. `unmount()` now flushes pending throttled renders and log writes before the final render, so the last frame is always displayed
2. `waitUntilExit()` now awaits the stdout drain callback, so the promise only resolves once buffered writes are actually flushed to the terminal

Users can use `process.exitCode` + `await waitUntilExit()` instead of `process.exit()` and get guaranteed final output. No new surface area.

sindresorhus:
1. **[P1] `waitUntilExit()` can resolve before the last write callback runs**  
	File: `ink/src/ink.tsx:463`  
	Current code only waits when `stdout.writableNeedDrain` is `true`. That misses streams where writes are async but never exceed `highWaterMark` (`writableNeedDrain` stays `false`).  
	I reproduced this with a delayed custom `Writable`: `waitUntilExit` resolved in ~1ms while the stream callback fired ~150-200ms later.

	What should change:
	- Don’t gate on `writableNeedDrain`.
	- Resolve/reject via a write callback barrier (for example `stdout.write('', callback)`), so resolution happens after queued writes are processed.

2. **[P2] Missing regression test for that path**  
	File: `ink/test/render.tsx:406` (new test area)  
	The new test covers throttled render flushing, but not the new `waitUntilExit` flush guarantee. Add a test with a delayed custom `Writable` where `writableNeedDrain` remains `false`, and assert `waitUntilExit()` does not resolve early.

costajohnt:
Thanks @sindresorhus, great catch on both points.

**P1 – write callback barrier:** You're right, gating on `writableNeedDrain` is insufficient since it only reflects backpressure, not whether queued write callbacks have actually fired. Replaced the `writableNeedDrain` check with a `stdout.write('', callback)` barrier so resolution is always sequenced after all prior writes are processed.

**P2 – regression test:** Added a test using a custom `Writable` with a delayed `_write()` that never exceeds `highWaterMark`, asserting that `waitUntilExit()` only resolves after the write callback fires.

costajohnt:
Follow-up fix: the empty write barrier (`stdout.write('', callback)`) was recording an observable call on sinon spy-based test stdouts, breaking tests that check `lastCall`. Now detects real writable streams via `_writableState`/`writableLength` and only uses the write barrier for those. For non-stream objects (test spies), resolves via `setImmediate` instead.

costajohnt:
Pushed an additional fix (6d537d2): when `unmount()` is called from `signal-exit` during process shutdown, the exit promise now resolves synchronously instead of using the write barrier. During process shutdown the event loop is draining, so async callbacks (`stdout.write` barrier or `setImmediate`) may never fire.

The detection works by checking the `error` parameter — `signal-exit` passes `(exitCode: number, signal: string | null)` while manual calls pass `undefined` or an `Error`.

Regarding the 2 remaining CI failures (`exit > exit normally` and `hooks > useStdout`): both fail identically on `master` with exit code 13 — they're pre-existing failures unrelated to this PR (likely a node-pty compatibility issue with the CI Node.js version).

PR #868, opened 2026-02-10
Add renderToString() for synchronous string output

### LINKED ISSUE #459: How to render to string
Is there a way to render to string? I am wanting to render to a file for documentation, etc.

costajohnt:
Closes #459.

## Summary

Add a `renderToString()` function that renders React elements to a string synchronously, without writing to stdout or setting up terminal event listeners.

```jsx
import {renderToString, Text, Box} from 'ink';

  <Box padding={1}>
    <Text color="green">Hello World</Text>
  </Box>,
  {columns: 40}
);
```

### How it works

The implementation uses the reconciler directly in **legacy (synchronous) mode** with a standalone root node and Yoga layout, bypassing the `Ink` class's terminal management entirely:

1. Creates a standalone `ink-root` DOM node (no stdout, stdin, or signal handlers)
2. Wires up `onComputeLayout` for Yoga layout and `onImmediateRender` to capture `<Static>` output during intermediate commits
3. Uses `updateContainerSync` + `flushSyncWork` to synchronously build and commit the React tree
4. Calls the existing `renderer()` to produce the string output
5. Tears down the reconciler container and frees Yoga nodes

### What's included

- **`src/render-to-string.ts`** — The implementation (~100 lines)
- **`test/render-to-string.tsx`** — 30 tests covering text, layout, styling, `<Static>`, `<Transform>`, borders, wrapping, columns option, effect behavior, error propagation, cleanup, and independent multiple calls
- **`src/index.ts`** — Exports `renderToString` and `RenderToStringOptions`
- **`readme.md`** — API documentation in the existing format

### Effect behavior

With React 19's reconciler, `flushSyncWork()` eagerly flushes effects synchronously. The two effect types behave differently:

- **`useEffect`** — Callbacks execute, but state updates they trigger do **not** affect the returned output. The output always reflects the initial render.
- **`useLayoutEffect`** — Fires synchronously during commit, so state updates **are** processed immediately and reflected in the output.

Effect cleanup functions run during teardown, preventing resource leaks.

### Error handling

Component errors are propagated to the caller rather than silently swallowed. The `onUncaughtError` reconciler callback captures the first error, and `renderToString` re-throws it after completing Yoga memory cleanup. This matches the semantics of React DOM's `renderToString()` and ensures callers can catch and handle rendering failures.

```jsx
// Errors throw instead of silently returning ''
try {
  renderToString(<BrokenComponent />);
} catch (error) {
  console.error('Render failed:', error.message);
}
```

Subsequent calls after an error work correctly — each invocation creates an independent reconciler container.

## Test plan

- [x] All 30 new tests pass
- [x] All 597 total tests pass (no regressions)
- [x] TypeScript compiles cleanly (`tsc --noEmit`)
- [x] Linter passes (`xo`)
- [x] Full `npm test` passes (tsc + xo + ava)

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -2254,6 +2254,14 @@ const output = renderToString(
 console.log(output);
 ```
 
+**Notes:**
+
+- Terminal-specific hooks (`useInput`, `useStdin`, `useStdout`, `useStderr`, `useApp`, `useFocus`, `useFocusManager`) return default no-op values since there is no terminal session. They will not throw, but they will not function as in a live terminal.
+- `useEffect` callbacks will execute during rendering (due to synchronous rendering mode), but state updates they trigger will not affect the returned output, which reflects the initial render.
+- `useLayoutEffect` callbacks fire synchronously during commit, so state updates they trigger **will** be reflected in the output.
+- The `<Static>` component is supported — its output is prepended to the dynamic output.
+- If a component throws during rendering, the error is propagated to the caller after cleanup.
+
 ##### tree
 
 Type: `ReactNode`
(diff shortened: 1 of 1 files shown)


sindresorhus:
I found one issue while testing this:

`renderToString()` currently returns a trailing newline for static-only trees. Example: rendering only `<Static>` items returns `"A\nB\n"` instead of `"A\nB"`.

Suggested fix:
- Normalize captured static output before returning by removing a single trailing newline that is added by the renderer for terminal rendering.
- Build the final return value like this:
  - static only: return normalized static output
  - static + dynamic: return normalized static output + `"
"` + dynamic output (keeps correct seperation)
  - dynamic only: return dynamic output
- Add tests for:
  1) static-only output has no trailing newline
  2) static + dynamic output still has exactly one newline between parts

costajohnt:
Thanks for the thorough review! All three points addressed:

**Trailing newline fix:** The renderer appends `\n` to static output for terminal rendering, so `renderToString()` now strips it before returning. Uses 3-way return logic:
- Static only → normalized static output (no trailing newline)
- Static + dynamic → normalized static + `\n` + dynamic
- Dynamic only → dynamic output as-is

**TSDoc:** Un-wrapped and synced word-for-word with the readme's Notes section.

**Tests added:**
1. Static-only output has no trailing newline
2. Static + dynamic output has exactly one newline between parts

PR #872, opened 2026-02-11
Add declarative `Cursor` component

### LINKED ISSUE #251: cursor support?

### LINKED ISSUE #870: useCursor requires component to know the position it's displayed at
The new `useCursor` hook's `setCursorPosition` method requires you to specify the absolute (ink-mount-relative) coordinate to render the cursor, which is very awkward if you want to use it from a component that's not your root component or if you make a lot of use of `<Box>` for layouts.

For example, if we wanted to update a component like [TextInput from ink-ui](https://github.com/vadimdemedes/ink-ui/?tab=readme-ov-file#text-input) to use `useCursor`, we would need to make it take props specifying the coordinate it's rendered at (because it might be rendered after a heading or nested inside a box with borders), and depending on what the parent components render, calculating this may require the user to use `measureElement` on one or more `<Box>` elements. Needing `measureElement` means that the cursor can't be positioned correctly on the initial render, and it means that components may need to be built to call `measureElement` again on every render that may affect layout to update the coordinates for the cursor. This all really breaks the clean encapsulation of React components.

It would be much more natural if Ink provided a `<Cursor />` element. Then you never have to pass a component's own coordinates to it as props, and you would never need to combine `useEffect` and `measureElement` to measure layout positioning. Updating components like ink-ui's TextInput would just take adding a `<Cursor />` element in the correct position.

Alternatively, if `measureElement` returned `x` and `y` properties in addition to `width` and `height` (making it very similar to the web's [Element.getBoundingClientRect](https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect) method), then users could easily implement a `<Cursor />` element on their own using `useEffect` and `measureElement`.

sindresorhus:
Fixes #870
Fixes #251

### DIFF
--- a/examples/cursor-input/cursor-input.tsx
+++ b/examples/cursor-input/cursor-input.tsx
@@ -1,0 +1,54 @@
+import React, {useRef, useState} from 'react';
+import {
+	render,
+	Box,
+	Text,
+	Cursor,
+	useInput,
+	type DOMElement,
+} from '../../src/index.js';
+
+function CursorTextInputExample() {
+	const [value, setValue] = useState('');
+	const cursorAnchorReference = useRef<DOMElement>(null);
+	const prompt = '> ';
+
+	useInput((input, key) => {
+		if (key.escape) {
+			setValue('');
+			return;
+		}
+
+		if (key.backspace || key.delete) {
+			setValue(previousValue => previousValue.slice(0, -1));
+			return;
+		}
+
+		if (
+			!key.ctrl &&
+			!key.meta &&
+			!key.return &&
+			input &&
+			!input.includes('\n') &&
+			!input.includes('\r')
+		) {
+			setValue(previousValue => previousValue + input);
+		}
+	});
+
+	return (
+		<Box flexDirection="column">
+			<Text>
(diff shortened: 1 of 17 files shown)


LoganDark:
I would love this, otherwise text inputs are very difficult since there is seemingly not even a way to get an absolute position to use for the cursor in the first place. On top of that there is no way to get the position of text after reflow either.

oliveryasuna:
Text inputs can be done with https://github.com/vadimdemedes/ink-ui.

PR #885, opened 2026-02-20
Handle missing package.json in dev mode + bundler regression

### LINKED ISSUE #886: Bundlers can't tree-shake `react-devtools-core` import after v6.8.0 refactor of DEV check
After upgrading from ink 6.7.0 to 6.8.0, production builds with Rspack (and likely other bundlers) fail with:

```
ERROR in ../../node_modules/ink/build/devtools.js 5:1-43
  × Module not found: Can't resolve 'react-devtools-core' in '/Users/user/Documents/release-checker/node_modules/ink/build'
```

This is because `react-devtools-core` is an optional peer dependency that is not installed in production (and not needed there), and the devtools import is no longer being tree-shaken by the bundler.

## Root Cause

In v6.7.0, the devtools guard in `reconciler.ts` used an inline check:

```js
if (process.env['DEV'] === 'true') {
    await import('./devtools.js');
}
```

Bundlers could statically evaluate `process.env['DEV']` (typically replaced with `undefined` or `"false"`), determine the condition is always `false`, and eliminate the entire branch - including the dynamic `import('./devtools.js')` which contains `import devtools from 'react-devtools-core'`.

In v6.8.0 (introduced by #877), this was refactored to:

```js
import {isDev} from './utils.js';
if (isDev()) {
    await import('./devtools.js');
}
```

Where `isDev` is:

```js
const isDev = () => process.env['DEV'] === 'true';
```

Even though Rspack inlines the function body and simplifies it to `return false === 'true'`, the bundler does **not** eliminate the `if (isDev())` branch and the dynamic import is still resolved, causing the build to fail because `react-devtools-core` isn't installed.

## Suggested Fix

Revert to the inline `process.env['DEV'] === 'true'` check in `reconciler.ts` (and `ink.ts`):

```js
if (process.env['DEV'] === 'true') {
    await import('./devtools.js');
}
```

## Environment

- ink: 6.8.0
- Bundler: Rspack (likely affects webpack and other bundlers too)
- Node.js: 24.11.0
- Works correctly in: ink 6.7.0

clicktodev:
Gracefully handle the absence of package.json by returning default values when loading it fails in development mode.

it seems undefined can still cause crashes in some instances

related. to #877

closes #886

### DIFF
--- a/src/reconciler.ts
+++ b/src/reconciler.ts
@@ -6,7 +6,7 @@ import {
 } from 'react-reconciler/constants.js';
 import * as Scheduler from 'scheduler';
 import Yoga, {type Node as YogaNode} from 'yoga-layout';
-import {createContext, version as ReactVersion} from 'react';
+import {createContext, version as reactVersion} from 'react';
 import {
 	createTextNode,
 	appendChildNode,
@@ -115,7 +115,7 @@ async function loadPackageJson() {
 
 let packageInfo = {
 	name: 'ink',
-	version: ReactVersion,
+	version: reactVersion,
 };
 
 if (process.env['DEV'] === 'true') {


sindresorhus:
CI is failing

clicktodev:
fixed

PR #669, opened 2024-08-29
Render last frame only when there is no output TTY

nickwesselman:
Fixes #667

Mirror existing CI behavior when there is no output TTY. This will enable the output of ink-based CLIs to gracefully pipe into other processes, such as pagers like `more` and `less`.

Also provides control over this behavior for CLIs which wish to support piping into processes that support ANSI deletion escapes.

## Testing
Try some various uses of the `Static` example --

```
npm run example examples/static
npm run example examples/static | more
CI=1 npm run example examples/static 
```

PR #888, opened 2026-02-25
Add interactive option for non-TTY and CI environments

### LINKED ISSUE #667: Expected behavior when there is no output TTY
We have a command with long-running output in a `<Static>` which we would like our users to be able to pipe into a command like `more` or `less`. However ink does not seem to handle piped output / lack of TTY as there are ANSI escapes in the output. You also see re-renders for non-Static elements.

```
> npm run example examples/static | more

> ink@5.0.1 example
> NODE_NO_WARNINGS=1 node --loader ts-node/esm examples/static


Completed tests: 0
ESC[2KESC[1AESC[2KESC[1AESC[2KESC[G✔ Test #1

Completed tests: 1
ESC[2KESC[1AESC[2KESC[1AESC[2KESC[G✔ Test #2

Completed tests: 2
ESC[2KESC[1AESC[2KESC[1AESC[2KESC[G✔ Test #3

Completed tests: 3
ESC[2KESC[1AESC[2KESC[1AESC[2KESC[G✔ Test #4

Completed tests: 4
ESC[2KESC[1AESC[2KESC[1AESC[2KESC[G✔ Test #5

Completed tests: 5
ESC[2KESC[1AESC[2KESC[1AESC[2KESC[G✔ Test #6

Completed tests: 6
ESC[2KESC[1AESC[2KESC[1AESC[2KESC[G✔ Test #7

Completed tests: 7
ESC[2KESC[1AESC[2KESC[1AESC[2KESC[G✔ Test #8

Completed tests: 8
ESC[2KESC[1AESC[2KESC[1AESC[2KESC[G✔ Test #9

Completed tests: 9
ESC[2KESC[1AESC[2KESC[1AESC[2KESC[G✔ Test #10

Completed tests: 10
```

If you simulate a CI run via environment variable, it actually behaves exactly as I would expect:

```
CI=1 npm run example examples/static | more

> ink@5.0.1 example
> NODE_NO_WARNINGS=1 node --loader ts-node/esm examples/static

✔ Test #1
✔ Test #2
✔ Test #3
✔ Test #4
✔ Test #5
✔ Test #6
✔ Test #7
✔ Test #8
✔ Test #9
✔ Test #10

Completed tests: 10
```

Shouldn't ink behave the same way for `!process.stdout.isTTY` as it would for `isInCi`? Is support for piping output not a common requirement in ink-based CLIs?

costajohnt:
## Summary

Closes #667

Adds a configurable `interactive` option that defaults to `false` when running in CI or when `stdout` is not a TTY. This allows Ink to gracefully handle piped output (e.g. `node app.js | cat`) by skipping terminal-specific features.

When `interactive` is disabled:
- Only the final frame of non-static output is written at unmount (no intermediate ANSI erase/rewrite cycles)
- Synchronized output (BSU/ESU) is skipped
- Cursor manipulation is skipped
- Resize handling is disabled
- Kitty keyboard auto-detection is disabled

The option can be explicitly set to override the auto-detected default:
```jsx
// Force interactive mode even when piping
render(<MyApp />, { interactive: true });

// Force non-interactive mode even in a TTY
render(<MyApp />, { interactive: false });
```

### Prior art
This builds on the approach from #669 (which was closed for inactivity) but is a fresh implementation against the current codebase.

## Files changed

- **src/ink.tsx** — Added `interactive` field with auto-detection default; gates terminal features on `this.interactive`
- **src/render.ts** — Added `interactive` to `RenderOptions` type with JSDoc
- **src/write-synchronized.ts** — `shouldSynchronize` accepts `interactive` parameter
- **src/components/App.tsx** — Added `interactive` prop; gated cursor show on unmount
- **test/write-synchronized.tsx** — Updated tests for new signature; added non-interactive TTY test
- **test/components.tsx** — Integration tests for non-TTY rendering, explicit override, TTY + interactive=false
- **test/helpers/create-stdout.ts** — Added optional `isTTY` parameter
- **readme.md** — Documented `interactive` option

## Test plan

- [x] All 878 tests pass
- [x] TypeScript type-check passes
- [x] Only pre-existing lint warnings (unrelated to this change)

### DIFF
--- a/src/ink.tsx
+++ b/src/ink.tsx
@@ -1012,19 +1012,28 @@ export default class Ink {
 		const opts = this.options.kittyKeyboard;
 		const mode = opts.mode ?? 'auto';
 
-		if (
-			mode === 'disabled' ||
-			!this.interactive ||
-			!this.options.stdin.isTTY ||
-			!this.options.stdout.isTTY
-		) {
+		if (mode === 'disabled') {
 			return;
 		}
 
 		const flags: KittyFlagName[] = opts.flags ?? ['disambiguateEscapeCodes'];
 
+		// 'enabled' force-enables the protocol as long as both streams are TTYs,
+		// regardless of the interactive setting (e.g. even in CI).
 		if (mode === 'enabled') {
-			this.enableKittyProtocol(flags);
+			if (this.options.stdin.isTTY && this.options.stdout.isTTY) {
+				this.enableKittyProtocol(flags);
+			}
+
+			return;
+		}
+
+		// Auto mode: require interactive + TTY
+		if (
+			!this.interactive ||
+			!this.options.stdin.isTTY ||
+			!this.options.stdout.isTTY
+		) {
 			return;
 		}
 
--- a/test/write-synchronized.tsx
+++ b/test/write-synchronized.tsx
@@ -1,5 +1,6 @@
 import EventEmitter from 'node:events';
 import test from 'ava';
+import isInCi from 'is-in-ci';
 import {bsu, esu, shouldSynchronize} from '../src/write-synchronized.js';
 
 const createStream = ({tty = false} = {}) => {
@@ -38,8 +39,12 @@ test('shouldSynchronize returns false for non-TTY stream', t => {
 test('shouldSynchronize uses CI detection when interactive is not specified', t => {
 	const ttyStream = createStream({tty: true});
 	// When interactive is omitted, shouldSynchronize falls back to is-in-ci.
-	// In a non-CI environment with a TTY stream, this should return true.
-	t.true(shouldSynchronize(ttyStream));
+	// In CI the result is false (non-interactive by design); outside CI it's true.
+	if (isInCi) {
+		t.false(shouldSynchronize(ttyStream));
+	} else {
+		t.true(shouldSynchronize(ttyStream));
+	}
 });
 
 test('shouldSynchronize returns false for non-TTY stream when interactive is not specified', t => {


sindresorhus:
- The default non-interactive detection only flips when `stdout` explicitly has `isTTY === false`. In a real piped run (`node app.js | cat`), `process.stdout` usually has no `isTTY` property at all, so it stays in interactive mode and still emits erase/cursor ANSI sequences.
- The new non-TTY test only asserts that the final write contains the last count. That can still pass even if intermediate frames are written, so the core behvaior is not fully protected.

costajohnt:
Good catches, both fixed:

1. **TTY detection** -- changed the guard from `('isTTY' in options.stdout && !options.stdout.isTTY)` to just `!options.stdout.isTTY`. Now piped streams where `isTTY` is absent (`undefined`) correctly trigger `nonInteractive` mode, since `!undefined` is truthy.

2. **Stronger non-TTY test** -- the test now asserts that no intermediate frames (`Count: 0` through `Count: 2`) are written, that no erase sequences appear, and that only the final frame (`Count: 3`) is present. This protects the core behavior rather than just checking the last write.

The detection change means `createStdout()` needed to default `isTTY` to `true` so test fakes behave like real TTY streams. That cascaded into ~50 test assertion updates across the suite to account for `cli-cursor` and synchronized-output escape sequences that now get emitted. All 880 tests pass (the only failure is the pre-existing `wrap-ansi` trim issue).

Let me know if the approach looks right or if you'd prefer a different direction.

sindresorhus:
I still see three test-quality gaps worth tightening before merge:

- `nonInteractive option overrides TTY detection`: this assertion is too weak for the claimed behvaior. It only checks for absence of one erase sequence, which can also happen in interactive paths, so it may pass even if the override regresses.
- `render all frames when nonInteractive is explicitly false`: the test name says "all frames", but it only verifies `Count: 0` and `Count: 2`. A regression dropping `Count: 1` would still pass.
- `static output is written immediately in non-interactive mode`: current assertions only prove static output appears at some point by the end, not that it was written immediately (before unmount/final teardown write).

costajohnt:
Tightened all three tests:

1. **Override test** — swapped the static `Hello` for a Counter that goes through 4 state changes, then asserts no intermediate frames were written and only the final frame appears. Can't pass unless nonInteractive actually engaged.
2. **Explicit false test** — added the missing `Count: 1` check and widened timers (50ms between renders, 500ms wait) so React can't batch away the intermediate frame.

EDIT:
3. Static output test: captures stdout writes before unmount() and asserts A and B are already present, proving they were flushed during render. Also verifies Dynamic was not written until unmount

sindresorhus:
The option name should be `interactive`.

src/write-synchronized.ts
```diff
@@ -1,9 +1,15 @@
 import {type Writable} from 'node:stream';
-import isInCi from 'is-in-ci';
```
sindresorhus: It should continue using this when `interactive` is not specified.

sindresorhus:
A few minor things:

1. The test named "shouldSynchronize uses CI detection when interactive is not specified" is incorrect in CI.
It currently expects `shouldSynchronize(ttyStream)` to be true when `interactive` is omitted, but with `CI=true` it should be false by design. I reproduced this by running AVA with `CI=1`.

2. `kittyKeyboard.mode: "enabled"` no longer acts like a force-enable when interactive resolves to false (for example, the default in CI).
Right now there is an early non-interactive return before the `"enabled"` branch, while the docs still describe `"enabled"` as force-enabling (with TTYs).

PR #893, opened 2026-03-01
Update XO to support react 19 rules

clicktodev:
ReactNode can be a promise however Ink's code assumes it's not a promise so i undid my async changes but they can be found [here](https://github.com/vadimdemedes/ink/pull/893/changes/2b40fcd6081ea730d7d4095c6e283404bf004318) if you wanna handle them in the future


sindresorhus:
Thanks :)

PR #894, opened 2026-03-06
Fix fullscreen flicker when incrementalRendering is enabled

jamiechapman:
The fullscreen code path in `renderInteractiveFrame` unconditionally uses `clearTerminal`, even when `incrementalRendering` is turned on. This means any app whose output fills the terminal gets a full screen wipe on every re-render, making incrementalRendering effectively broken for real-world fullscreen apps.

The fix routes fullscreen frames through the incremental line-diff renderer when the option is set, so only changed lines are rewritten. The `clearTerminal` path is preserved as-is for standard mode.

Adds a test + fixture for the incremental fullscreen case.

### DIFF
--- a/src/ink.tsx
+++ b/src/ink.tsx
@@ -959,6 +959,18 @@ export default class Ink {
 		});
 
 		if (shouldClearTerminal) {
+			// skip clearTerminal in incremental mode to avoid erase-then-redraw flicker
+			if (this.options.incrementalRendering) {
+				if (output !== this.lastOutput || this.log.isCursorDirty()) {
+					this.throttledLog(outputToRender);
+				}
+
+				this.lastOutput = output;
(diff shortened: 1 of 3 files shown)


sindresorhus:
I think this still regresses one supported case: when the new incremental early-return runs, it returns before the `staticOutput` handling, so any newly emitted `<Static>` content in the same render never gets written.

I reproed it with `incrementalRendering: true`, first frame at full height, second frame at `rows - 1`, and a `<Static>` line appearing on that second frame. On this branch the static line never shows up, while the same setup on `master` does show it.

So this fixes the fullscreen flicker case, but it also drops `<Static>` output in that shrink path. The new test does not catch that because it never adds `<Static>` content while going through this branch.

PR #925, opened 2026-04-02
Add wrap-anywhere option for Text component

### LINKED ISSUE #741: [Feature Request] Support `word-break` like props for `<Text />` component
Currently, the `<Text />` component supports the `wrap` prop, which controls the overflow behavior.

```js
	if (wrapType === 'wrap') {
		wrappedText = wrapAnsi(text, maxWidth, {
			trim: false,
			hard: true,
		});
	}
```

Can we add a prop like `word-break` to control the wrapping behavior, provided by the `wrap-ansi` package?  
[wrap-ansi: wordwrap](https://github.com/chalk/wrap-ansi?tab=readme-ov-file#wordwrap)

``` js
	if (wrapType === 'wrap') {
		wrappedText = wrapAnsi(text, maxWidth, {
			trim: false,
			hard: true,
                        wordwrap: style['word-break'] ?? true
		});
	}
```

virskyline:
Add 'wrap-anywhere' as a new value for the wrap prop on <Text />.

Unlike the default 'wrap' which tries to break at word boundaries, 'wrap-anywhere' fills each line to the full column width, breaking mid-word when needed. Uses wrap-ansi with wordWrap: false.

Closes #741

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -366,12 +366,12 @@ Invert background and foreground colors.
 #### wrap
 
 Type: `string`\
-Allowed values: `wrap` `wrap-anywhere` `truncate` `truncate-start` `truncate-middle` `truncate-end`\
+Allowed values: `wrap` `hard` `truncate` `truncate-start` `truncate-middle` `truncate-end`\
 Default: `wrap`
 
 This property tells Ink to wrap or truncate text if its width is larger than the container.
 If `wrap` is passed (the default), Ink will wrap text and split it into multiple lines.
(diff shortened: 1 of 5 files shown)


sindresorhus:
The `wrap` value should really have been `normal`, not `wrap`, since `wrap="wrap"` is really weird.

So I don't think we should continue that bad API choice and instead name this explicitly: `wrap="hard"`.

virskyline (commented):
Good Point

sindresorhus:
You have merge conflict

virskyline:
Merge Conflict solved

PR #936, opened 2026-04-10
Replace clearTerminal with eraseScreen to preserve scrollback

### LINKED ISSUE #935: Render loop calls clearTerminal (CSI 3 J) — wipes terminal scrollback when output exceeds viewport
## Description

Ink's render path calls `clearTerminal` from `ansi-escapes` when `outputHeight > viewportRows`. This sequence includes `\e[3J` (CSI 3 J, "Erase Saved Lines"), which wipes the host terminal's scrollback buffer on every re-render that overflows.

Short-lived apps are not affected since they exit before accumulating scrollback. Long-running apps hit this on every re-render cycle. The most visible downstream case is `@anthropic-ai/claude-code`, where this has generated 25+ interactions across two open issues (anthropics/claude-code#16310, #2479).

Related: this was partially identified in #359 (2020), where a contributor pointed out the `clearTerminal` sequence wipes scrollback and proposed `\e[H\e[0J` instead. That issue was closed without changing the render path.

## Root cause

```javascript
ansiEscapes.clearTerminal =
    process.platform === 'win32'
        ? `${ansiEscapes.eraseScreen}${ESC}0f`
        : `${ansiEscapes.eraseScreen}${ESC}3J${ESC}H`;
```

`clearTerminal` is correct per VT spec — it clears everything including saved lines. The problem is that Ink uses it for *render repaints*, where only the visible screen needs clearing.

## Impact

Every terminal that implements VT correctly honors CSI 3 J. There is no terminal-side opt-out except iTerm2 on macOS. Affected: tmux, GNU screen, xterm.js (and everything built on it: VS Code, code-server, ttyd), Microsoft Terminal, WezTerm, Alacritty, Kitty, Ghostty.

Downstream issues:
- anthropics/claude-code#16310, #2479
- openai/codex#14277 (Codex CLI, same xterm.js root cause)
- microsoft/terminal#8736 (closed without toggle)

## Proposed fix

Replace `clearTerminal` in the render path with `eraseScreen + cursor-home`:

```javascript
// Instead of clearTerminal (which includes \e[3J):
process.stdout.write(ansiEscapes.eraseScreen + ansiEscapes.cursorTo(0, 0));
```

This clears the visible viewport for redrawing without touching the scrollback buffer. Reserve `clearTerminal` for explicit user-initiated clears.

Alternative: add a `clearScrollbackOnRender` option (default `false`) so apps can opt in to the current behavior if they need it.

## Minimal reproducer

```jsx
import {render, Text} from 'ink';

const App = () => {
    const [tick, setTick] = useState(0);

    useEffect(() => {
        const timer = setInterval(() => setTick(t => t + 1), 500);
        return () => clearInterval(timer);
    }, []);

    // Force re-renders with output exceeding viewport
    return (
        <>
            {Array.from({length: 80}, (_, i) => (
                <Text key={i}>Line {i} (tick {tick})</Text>
            ))}
        </>
    );
};

render(<App />);
```

Run under tmux. Wait a few seconds for re-renders to fire, then check:
```bash
tmux display-message -p '#{history_size}'
# Returns 0 — scrollback was wiped by the render loop
```

Without Ink (plain `console.log` producing the same output), `history_size` grows as expected.

moeritze:
## Summary

Replace `clearTerminal` with `eraseScreen + cursorHome` in the render path to preserve terminal scrollback buffer.

- `clearTerminal` includes `\e[3J` (CSI 3 J / "Erase Saved Lines") which wipes the host terminal's scrollback on every re-render that overflows the viewport
- The render path only needs to clear the visible screen for repainting — scrollback should be preserved
- This change uses `eraseScreen` (`\e[2J`) + `cursorTo(0, 0)` instead, which clears the viewport without touching scrollback

Fixes #935

## Test plan

- [x] All 61 render tests pass with updated assertions
- [x] Build succeeds (`tsc`)
- [x] Linter passes (`xo`)
- [ ] Verify manually in a terminal that scrollback is preserved during fullscreen re-renders

### DIFF
--- a/src/ink.tsx
+++ b/src/ink.tsx
@@ -1056,7 +1056,10 @@ export default class Ink {
 			}
 
 			this.options.stdout.write(
-				ansiEscapes.clearTerminal + this.fullStaticOutput + output,
+				ansiEscapes.eraseScreen +
+					ansiEscapes.cursorTo(0, 0) +
+					this.fullStaticOutput +
+					output,
 			);
 			this.lastOutput = output;
 			this.lastOutputToRender = outputToRender;
--- a/test/render.tsx
+++ b/test/render.tsx
@@ -29,6 +29,9 @@ import createStdout from './helpers/create-stdout.js';
 const eraseScreenSequence =
 	ansiEscapes.eraseScreen + ansiEscapes.cursorTo(0, 0);
 
+// The scrollback-wiping escape sequence that should never appear in output (#935)
+const scrollbackWipeSequence = '\u001B[3J';
+
 const textDecoder = new TextDecoder();
 
 const require = createRequire(import.meta.url);
@@ -363,6 +366,10 @@ test.serial('erase screen', async t => {
 	const ps = term('erase', ['3']);
 	await ps.waitForExit();
 	t.true(ps.output.includes(eraseScreenSequence));
+	t.false(
+		ps.output.includes(scrollbackWipeSequence),
+		'Erase screen must not emit \\e[3J (scrollback wipe)',
+	);
 
 	for (const letter of ['A', 'B', 'C']) {
 		t.true(ps.output.includes(letter));
@@ -375,6 +382,10 @@ test.serial(
 		const ps = term('erase', ['3']);
 		await ps.waitForExit();
 		t.true(ps.output.includes(eraseScreenSequence));
+		t.false(
+			ps.output.includes(scrollbackWipeSequence),
+			'Erase screen must not emit \\e[3J (scrollback wipe)',
+		);
 
 		for (const letter of ['A', 'B', 'C']) {
 			t.true(ps.output.includes(letter));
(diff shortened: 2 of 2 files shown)


sindresorhus:
Right now the assertions only check for the new viewport-clear sequence. That proves the new sequence is emitted, but it does not prove the scrollback-wiping escape is gone. `clearTerminal` is a different sequence (`2J + 3J + H`), so a future change could reintroduce it and these tests would still pass.

I'd tighten the tests to explicitly assert that `clearTerminal` / `\u001B[3J` is absent in the fullscreen and overflow cases. That's the acutal user-visible bug this PR fixes, so the coverage should pin that down too.

sindresorhus:
I would like to see a couple of more tests:

---

1. “fullscreen rerender does not emit `\u001B[3J`”

This should be a direct raw-output regression test for `#935`.

- Render a fixture at exactly terminal height or taller.
- Trigger a few rerenders.
- Assert the captured output does not contain `\u001B[3J`.
- Keep a basic sanity check that later frames still render.

Why: the current tests prove the new clear sequence is used, but not that the old scrollback-wiping escape is gone.

---

2. “shrink from overflow to non-fullscreen clears stale content without emitting `\u001B[3J`”

This should protect the old reason the clear path existed.

- First frame is taller than the terminal.
- Next frame shrinks below terminal height.
- Assert removed lines from the tall frame are not still visible in the final frame.
- Also assert the raw output does not contain `\u001B[3J`.

Why: otherwise a future change could preserve scrollback but reintroduce the old stale-lines bug.

sindresorhus:
I double-checked this manually. Removing `CSI 3J` is a real improvement, but I don't think `eraseScreen + cursorHome` is the best replacement.

VS Code enables xterm.js's `scrollOnEraseInDisplay`, and Windows Terminal handles `CSI 2J` similarly: the current viewport is moved into scrollback before it is cleared. Repeated redraws therefore churn the bounded history and eventually evict older content.

I reproduced this with `@xterm/headless`. After five redraws, `CSI 2J + home` moved the scrollback base from 2 to 17, while `home + eraseDown` kept it at 2. I also tested the actual overflow-to-short fixture; both produced the correct final viewport, but `home + eraseDown` caused much less scrollback movement.

Could we use:

```js
ansiEscapes.cursorTo(0, 0) + ansiEscapes.eraseDown
```

This clears from the top-left through the end of the display without ED2's terminal-specific scrollback behavior.

There are two test gaps:

1. The clear count searches for the compound `eraseScreen + cursorHome` sequence. Emitting `eraseScreen` alone, or reintroducing `clearTerminal`, would incorrectly count as zero. I would count `ansiEscapes.eraseScreen` seperately.

2. The stale-content test does not verify the final terminal state. Give overflow-only rows a unique marker and assert that it is absent from the final emulated viewport.

Oversized dynamic frames will still consume scrollback simply by being written, so normal-buffer preservation can only be best-effort. Fullscreen apps should use the alternate screen or keep their UI within the viewport.

costajohnt:
Since this has been quiet since April and the review left a concrete ask, I opened #994 building on this: same idea, but with `cursorTo(0, 0) + eraseDown` per the review, the count assertions tightened so `clearTerminal` / `eraseScreen` regressions fail, and a fixture that replays the PTY output through Ink's terminal reducer to check pre-existing scrollback survives and the final viewport has no stale rows. Credited this PR in the description. Happy to close mine if you'd rather pick this one back up.

PR #940, opened 2026-04-23
Refactor: Split App component into smaller modules

j1mr10rd4n:
This PR splits the top-level `App` component into a set of smaller, more specialised ContextProviders. This is a pure refactor with no changes to public hook signatures, and all tests passing without changes.

The main motivation here was readability. Whilst creating another Ink-based application, I needed to create a different way of handling `focus`. It was difficult to discover whether I could re-use and/or customise the included `use-focus` and `use-focus-manager` because so many different concerns were included in the `App` component. In the new structure, it's now evident that the `eventEmitter` is the only cross-cutting dependency in the `FocusContext`.

The new ContextProviders are: `AppContextProvider`, `AnimationContextProvider`, `CursorContextProvider`, `FocusContextProvider`, `StdinContextProvider` ,`StdoutContextProvider`, `StderrContextProvider`.

Additionally, some of the raw `stdin` and `stdout` functionality is factored out into two helper classes:`StdinHelper` and `StdoutHelper`. These help to separate out the pure console/tty concerns from the React framework concerns.

### DIFF
--- a/src/components/AnimationContext.ts
+++ b/src/components/AnimationContext.ts
@@ -1,6 +1,6 @@
 import {createContext} from 'react';
 
-type AnimationContextValue = {
+export type AnimationContextValue = {
 	readonly renderThrottleMs: number;
 	readonly subscribe: (
 		callback: (currentTime: number) => void,
--- a/src/components/App.tsx
+++ b/src/components/App.tsx
@@ -9,7 +9,6 @@ import React, {
 	useEffect,
 } from 'react';
 import cliCursor from 'cli-cursor';
-import {type CursorPosition} from '../log-update.js';
 import {createInputParser} from '../input-parser.js';
 import AppContext from './AppContext.js';
 import StdinContext from './StdinContext.js';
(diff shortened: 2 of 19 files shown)


j1mr10rd4n:
Seeking collaboration/feedback on this draft! @sindresorhus @costajohnt @clicktodev et al.

costajohnt:
Hey @j1mr10rd4n , thanks for pulling me in. Pulling imperative TTY work out of React hooks is a reasonable instinct, and the top-level JSX is easier to scan.

That said, I don't think the architecture change earns its way in. The separation doesn't fully land: `AppContextProvider` still owns both `StdinHelper` and `StdoutHelper`, `handleSetBracketedPasteMode` is a stdout concern plumbed through `AppContext`, and the Cursor/Stdout/Stderr providers are single-`useMemo` relocations. `AppContext` went from 2 fields to 8. ~335 net new lines with no public API change is a hard sell.

The rewrite also introduced a subtle regression: `FocusContextProvider.tsx:126` early-returns when `focusablesCountRef.current === 0`, which now skips escape handling too. The original only gated escape on `isFocusEnabled`, not on focusable count. Hard to trigger in practice (removing the last focusable clears `activeFocusId` itself), but evidence that wholesale rewrites of a 700-line file carry real cost even when tests pass.

You mentioned the motivation was handling focus differently in a downstream app. A PR that actually exposes a pluggable focus API gives maintainers something concrete to weigh against the diff. Alternatively, a narrow PR that just extracts `StdinHelper`/`StdoutHelper` (the one piece with real cohesion benefit) without the provider explosion would be easier to land.

clicktodev:
thanks for the pr but i agree that the scope is too big. this should be broken into small focussed PRs.

j1mr10rd4n:
Great feedback, thanks for taking the time to review. I'll split out some smaller pieces.

PR #941, opened 2026-04-27
Fix linting errors, warnings

j1mr10rd4n:
This PR fixes the master branch CI build which is currently failing with linting errors in two test files:
* test/components.tsx
* test/kitty-keyboard.tsx

The transgression was `@typescript-eslint/no-unnecessary-type-assertion` with message "This assertion is unnecessary since the receiver accepts the original type of the expression." 

Fixes are also included for `react-hooks/exhaustive-deps` prettier warnings introduced  in commit cfaebbb.

### DIFF
--- a/src/hooks/use-input.ts
+++ b/src/hooks/use-input.ts
@@ -265,7 +265,7 @@ const useInput = (inputHandler: Handler, options: Options = {}) => {
 		return () => {
 			internal_eventEmitter.removeListener('input', handleData);
 		};
-	}, [options.isActive, internal_eventEmitter]);
+	}, [options.isActive, internal_eventEmitter, handleData]);
 };
 
 export default useInput;
--- a/src/hooks/use-paste.ts
+++ b/src/hooks/use-paste.ts
@@ -77,7 +77,7 @@ const usePaste = (
 		return () => {
 			internal_eventEmitter.removeListener('paste', handlePaste);
 		};
-	}, [options.isActive, internal_eventEmitter]);
+	}, [options.isActive, internal_eventEmitter, handlePaste]);
 };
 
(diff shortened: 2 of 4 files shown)


clicktodev:
should we make the exhuastive deps rule an error instead of warning?

huww98:
This PR introduced a bug in `src/hooks/use-input.ts` (and `use-paste.ts`).

The change at line 268 added `handleData` back to the `useEffect` dependency array:

```diff
-	}, [options.isActive, internal_eventEmitter]);
+	}, [options.isActive, internal_eventEmitter, handleData]);
```

However, `handleData` is created with `useEffectEvent` (line 176). Per the [React docs](https://react.dev/reference/react/useEffectEvent#caveats):

> Effect Event functions do not have a stable identity. Their identity intentionally changes on every render.

This means including them in the dependency array causes the effect to re-run on **every render**, tearing down and re-subscribing the event listener each time — completely negating the purpose of using `useEffectEvent` in the first place (which was introduced in cfaebbb specifically to avoid per-render re-subscription).

The exhaustive-deps lint rule likely flagged this as a missing dependency, but `useEffectEvent` return values should be excluded from deps (the `eslint-plugin-react-hooks` linter is supposed to understand this, but may not have been configured correctly).

This was correct before this PR and should be reverted for these specific lines.

clicktodev:
Its odd that this wasnt caught by the linter because the react docs say that the linter should warn about including Effect events in useEffect dep array

```ts
// 🔴 Wrong: including Effect Event in dependencies
useEffect(() => {
  onSomething();
}, [onSomething]); // ESLint will warn about this

```

Could this be a bug in the linter?

PR #945, opened 2026-05-11
Fix `useBoxMetrics` not accepting ref objects with an initial null value

gaac510:
This pull request hopes to change the current behaviours below to the expected behaviours.

Current Behaviours
--

With `complierOptions.strictNullChecks: true` set in `tsconfig.json`, compilation with the below code in the project always fails due to `useBoxMetrics` only accepting `RefObject<DOMElement>` and not also `RefObject<DOMElement | null>`.

```ts
const ref = useRef<DOMElement>(null);
const {...} = useBoxMetrics(ref);
```

Further, VSCode highlights a type error regardless of the compiler options sets:
    Image

Expected Behaviours
--

1. When `complierOptions.strictNullChecks: true` is set in `tsconfig.json`, the code above shouldn't cause compilation to fail.
2. `useBoxMetrics` should be able to accept `RefObject<DOMElement | null>` since it's common for a ref object (especially one that will be passed to a DOM node's ref attribute) to be created with an initial `null` value.
3.  VSCode shouldn't report a type error.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -2080,7 +2080,7 @@ Use `hasMeasured` to detect when the currently tracked element has been measured
 
 #### ref
 
-Type: `React.RefObject<DOMElement>`
+Type: `React.RefObject<DOMElement | null>`
 
 A ref to the `<Box>` element to track.
 
--- a/src/hooks/use-box-metrics.ts
+++ b/src/hooks/use-box-metrics.ts
@@ -44,7 +44,8 @@ const emptyMetrics: BoxMetrics = {
 	top: 0,
 };
 
-const findRootNode = (node?: DOMElement): DOMElement | undefined => {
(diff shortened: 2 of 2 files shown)


sindresorhus:
The docs still lists the `ref` type as `React.RefObject<DOMElement>`. Since this PR changes the accepted public type to `React.RefObject<DOMElement | null>`, and the example right below uses `useRef(null)`, the documented type should be updated too.

gaac510:
@sindresorhus Thanks. Updated the docs as well now.
BTW I can see my fork is now 1 commit behind but the merge request check still says "No conflicts with base branch", so... hopefully that mean I shouldn't need to sync my fork?

gaac510:
Hi, sorry, I haven't contributed to open source projects much, so I'm not familiar with the process. Is there anything else I need to do to complete this merge request?

Issue #904, opened 2026-03-17
rootNode.staticNode dangling reference causes OOM after <Static> unmount


TwoAbove:
Hey! Thanks for the awesome lib.

While writing a cli preview for a couple of apps, I hit a weird error where the bun process executing ink explodes in memory usage to 150+GB. I was stumped and thought it was an issue on my side. In a way it was, but while fixing it I think I stumbled upon a use-after-free error.

Ink: 6.8.0
OS: macOS 26.3
Env: Bun 1.3.10

Here's my theory about what's going on:

When a `<Static>` component unmounts, `cleanupYogaNode` frees via `freeRecursive()`, but `rootNode.staticNode` is never cleared. The renderer checks `node.staticNode?.yogaNode` on every render, and the JS object still exists (held by the ref), so the check passes. It then calls `getComputedWidth()`/`getComputedHeight()` on the freed WASM node.

For a while the freed memory still has valid dimensions. After churning enough Yoga nodes (from normal React stuff), the WASM allocator reuses the memory. The memory now has garbage values, and `Output.get()` tries to allocate a `width * garbage_height` grid -> instant OOM (I saw 10+ GB/s allocation).

Reproduces 50% of the time (it either does or it doesnt). The pre-churn phase in the POC simulates the churn before the `<Static>` unmount. Then you just hold down the down arrow until it freezes. The freeze usually happens right near the end of the list.

The fix that I thought of was to clear `rootNode.staticNode` in `removeChild/removeChildFromContainer` when the removed node has `internal_static`

```ts
removeChild(node, removeNode) {
    removeChildNode(node, removeNode);
    cleanupYogaNode(removeNode.yogaNode);
    if (removeNode.internal_static && currentRootNode) {
        currentRootNode.staticNode = undefined;
    }
}
```

You can roll the dice with the script below. It may take a few runs to get it to fail:

```tsx
import { Box, render, Static, Text, useInput, useStdout } from "ink";
import { createElement, useEffect, useMemo, useRef, useState } from "react";

// This component rapidly creates/destroys React elements to churn Yoga nodes,
// simulating what happens in a real app with rapid UI updates.
function NodeChurner({ onDone }: { onDone: () => void }) {
  const [tick, setTick] = useState(0);
  const maxTicks = 200;


  useEffect(() => {
    if (tick >= maxTicks) {
      onDone();
      return;
    }
    const id = setTimeout(() => setTick((t) => t + 1), 5);
    return () => clearTimeout(id);
[...]

PR #948, opened 2026-05-12
Fix: <Static> remount via key change drops new items

### LINKED ISSUE #947: <Static> remount via key change drops new items (post-7.0 regression)
## Summary

In ink 7.x, remounting `<Static>` via a `key` prop change causes the new instance's items to silently drop from stdout. The dynamic frame keeps rendering, so the symptom is "items vanished" rather than a crash. Apps that drive a Static replay by bumping a key — common pattern in TUI history panes (e.g. Gemini CLI, qwen-code's \`historyRemountKey\`) — see the entire static region go blank after \`/clear\`, model switches, Ctrl+O / compact-toggle, auth refresh, or any other forced replay trigger.

## Minimal reproduction

\`\`\`tsx
import React, {useEffect, useState} from 'react';
import {render, Box, Text, Static} from 'ink';

function App() {
  const [items, setItems] = useState([
    {id: 'h', text: '=== OLD HEADER ==='},
    {id: 'a', text: 'old item A'},
    {id: 'b', text: 'old item B'},
  ]);
  const [key, setKey] = useState(0);

  useEffect(() => {
    setTimeout(() => {
      // Remount the Static with a different items array.
      setItems([{id: 'h2', text: '=== NEW HEADER (after remount) ==='}]);
      setKey(k => k + 1);
    }, 300);
    setTimeout(() => {
      setItems(curr => [...curr, {id: 'n1', text: 'new item N1'}]);
    }, 600);
    setTimeout(() => process.exit(0), 1000);
  }, []);

  return (
    <Static key={key} items={items}>
      {item => <Box key={item.id}><Text>{item.text}</Text></Box>}
    </Static>
  );
}

render(<App />);
\`\`\`

**Expected stdout** (chronological):
\`\`\`
=== OLD HEADER ===
old item A
old item B
=== NEW HEADER (after remount) ===
new item N1
\`\`\`

**Actual stdout**:
\`\`\`
=== OLD HEADER ===
old item A
old item B
\`\`\`
The post-remount items are never emitted.

## Root cause

\`reconciler.removeChildFromContainer\` (and \`removeChild\`) unconditionally clear \`rootNode.staticNode\` whenever the removed node has \`internal_static === true\`:

https://github.com/vadimdemedes/ink/blob/1510851/src/reconciler.ts#L294-L301

React processes **insertions before deletions** inside a single commit. When a \`<Static>\` is remounted via key change, the commit looks like:

1. \`createInstance\` fires for the NEW internal_static node → registers \`rootNode.staticNode = newNode\`, marks \`isStaticDirty = true\`.
2. \`appendChildToContainer(rootNode, newStaticNode)\` attaches the new node.
3. \`removeChildFromContainer(rootNode, oldStaticNode)\` fires for the OLD node → sees \`removeNode.internal_static === true\` and **zeroes out the just-registered pointer**.

After the commit, \`renderer.ts:48\` evaluates \`if (node.staticNode?.yogaNode)\` → false → \`staticOutput\` is empty → \`hasStaticOutput\` is false → the new \`<Static>\`'s items never reach \`fullStaticOutput\` or stdout.

## Relation to #904 / #905

PR #905 (\"clear staticNode reference on Static unmount\", fixes #904) added these clears to plug the OOM after a Static unmount. The fix correctly handles the **pure unmount** path, but does not consider the **insertion-before-deletion order** of a key-driven remount in the same commit, where the new node has already taken ownership of \`rootNode.staticNode\` by the time the old node's removal callback fires.

## Suggested fix

Add an identity guard so the clear only runs when the removed node is still the one the root points at:

\`\`\`diff
 removeChildFromContainer(node, removeNode) {
   removeChildNode(node, removeNode);
   cleanupYogaNode(removeNode.yogaNode);

-  if (removeNode.internal_static && currentRootNode) {
+  if (
+    removeNode.internal_static &&
+    currentRootNode &&
+    currentRootNode.staticNode === removeNode
+  ) {
     currentRootNode.staticNode = undefined;
   }
 },
\`\`\`

The same guard needs to be added to \`removeChild\` for nested-remount cases.

I have a PR with the fix + a regression test ready: vadimdemedes/ink#TBD (will link as soon as I open it).

## Real-world impact

Both Gemini CLI and qwen-code use a \`<Static key={historyRemountKey}>\` pattern to force a full replay of the chat history after \`/clear\`, model switch, Ctrl+O, etc. After upgrading either to ink 7, the entire history pane stops rendering immediately after the first such trigger — confirmed end-to-end:
- qwen-code regression PR thread: https://github.com/QwenLM/qwen-code/pull/4083 (currently reverts ink 7 → 6 as a stop-gap)

## Environment

- ink \`7.0.2\` (master \`1510851\`)
- Node 22.x, React 19.2.4
- macOS / Ubuntu / Windows — reproduction is platform-independent (verified via the 30-line repro above, no terminal-emulator quirks involved)

chiga0:
Fixes #947.

## Summary

\`reconciler.removeChildFromContainer\` / \`removeChild\` unconditionally cleared \`rootNode.staticNode\` whenever the removed node was marked \`internal_static\`. React processes insertions before deletions inside a single commit, so when a \`<Static>\` is remounted via a \`key\` prop change:

1. \`createInstance\` fires for the **new** internal_static node and registers \`rootNode.staticNode = newNode\` (also marking \`isStaticDirty = true\`).
2. \`appendChildToContainer\` attaches the new node.
3. \`removeChildFromContainer\` fires for the **old** internal_static node, sees \`removeNode.internal_static === true\`, and zeroes out \`currentRootNode.staticNode\` — clobbering the pointer the new node just registered.

After that, \`renderer.ts\` sees \`node.staticNode === undefined\`, emits no static output, and the new \`<Static>\`'s items are silently dropped from stdout.

## Fix

Clear \`staticNode\` only when the node being removed is the one the root currently points to. Identical guard added to both \`removeChildFromContainer\` (root-level remount) and \`removeChild\` (nested remount).

\`\`\`diff
- if (removeNode.internal_static && currentRootNode) {
+ if (
+   removeNode.internal_static &&
+   currentRootNode &&
+   currentRootNode.staticNode === removeNode
+ ) {
    currentRootNode.staticNode = undefined;
  }
\`\`\`

The existing #904 / #905 path (pure unmount → \`staticNode\` should be cleared) is unaffected: in that case \`currentRootNode.staticNode\` still equals \`removeNode\` at removal time, so the guard passes through.

## Test

Adds \`remounting <Static> via key change emits the new items\` to \`test/components.tsx\`. Asserts that after a key-driven remount of \`<Static>\` with a disjoint items array, the new items reach stdout.

Verified:
- New test **fails** on \`master\` (\`1510851\`) — confirms the regression
- New test **passes** with this patch
- Existing Static tests all pass — including the #904 \`stops accumulating after Static unmounts\` regression (the pure-unmount path is preserved)
- Full \`ava\` suite shows no new failures vs. \`master\` baseline (two pre-existing flaky failures on \`cursor\` and \`wrap-ansi doesn't trim leading whitespace\` are unrelated and reproduce on \`master\` without this patch)

## Real-world impact

Both Gemini CLI and qwen-code use a \`<Static key={historyRemountKey}>\` pattern to force a full replay of the chat history after \`/clear\`, model switch, Ctrl+O, etc. After upgrading to ink 7, the entire history pane stops rendering immediately after the first such trigger. End-to-end repro on qwen-code: https://github.com/QwenLM/qwen-code/pull/4083 (currently has to revert ink 7 → 6 as a stop-gap).

### DIFF
--- a/src/reconciler.ts
+++ b/src/reconciler.ts
@@ -295,7 +295,18 @@ export default createReconciler<
 		removeChildNode(node, removeNode);
 		cleanupYogaNode(removeNode.yogaNode);
 
-		if (removeNode.internal_static && currentRootNode) {
+		// Only clear staticNode if the node being removed is the one the root
+		// currently points at. React processes insertions before deletions in
+		// a single commit, so when <Static> is remounted via a `key` change the
+		// new staticNode is registered by createInstance BEFORE this removal
+		// fires for the old node. Clearing unconditionally would clobber the
+		// fresh pointer and the renderer would emit no static output for the
+		// new Static instance.
+		if (
+			removeNode.internal_static &&
+			currentRootNode &&
+			currentRootNode.staticNode === removeNode
+		) {
 			currentRootNode.staticNode = undefined;
 		}
 	},
@@ -351,7 +362,13 @@ export default createReconciler<
 		removeChildNode(node, removeNode);
 		cleanupYogaNode(removeNode.yogaNode);
 
-		if (removeNode.internal_static && currentRootNode) {
+		// See removeChildFromContainer above for the rationale on the
+		// `currentRootNode.staticNode === removeNode` guard.
+		if (
+			removeNode.internal_static &&
+			currentRootNode &&
+			currentRootNode.staticNode === removeNode
+		) {
 			currentRootNode.staticNode = undefined;
 		}
 	},
--- a/src/reconciler.ts
+++ b/src/reconciler.ts
@@ -304,8 +304,7 @@ export default createReconciler<
 		// new Static instance.
 		if (
 			removeNode.internal_static &&
-			currentRootNode &&
-			currentRootNode.staticNode === removeNode
+			currentRootNode?.staticNode === removeNode
 		) {
 			currentRootNode.staticNode = undefined;
 		}
@@ -366,8 +365,7 @@ export default createReconciler<
 		// `currentRootNode.staticNode === removeNode` guard.
 		if (
 			removeNode.internal_static &&
-			currentRootNode &&
-			currentRootNode.staticNode === removeNode
+			currentRootNode?.staticNode === removeNode
 		) {
 			currentRootNode.staticNode = undefined;
 		}
--- a/test/components.tsx
+++ b/test/components.tsx
@@ -555,6 +555,62 @@ test('static output stops accumulating after Static unmounts (#904)', t => {
 	t.true(outputAfterChurn.includes('Dynamic'));
 });
 
+test('remounting <Static> via key change emits the new items', t => {
+	// Regression test for the post-#904 follow-up.
+	//
+	// React processes insertions before deletions within a commit. When a
+	// `<Static>` is remounted via a `key` prop change in the same commit:
+	//
+	//   1. createInstance fires for the new internal_static node, which sets
+	//      `rootNode.staticNode = newNode` and marks `isStaticDirty = true`.
+	//   2. appendChildToContainer attaches the new node.
+	//   3. removeChildFromContainer fires for the OLD internal_static node.
+	//
+	// Before this fix, step 3 unconditionally set
+	// `currentRootNode.staticNode = undefined`, clobbering the just-registered
+	// new staticNode pointer. The renderer then saw `staticNode === undefined`,
+	// emitted no static output, and the new Static's items were silently
+	// dropped from stdout. Real-world impact: TUIs that drive a Static
+	// "history replay" via key bumping (gemini-cli / qwen-code's
+	// `historyRemountKey` pattern) saw the history go blank after /clear,
+	// model switches, or any other `refreshStatic()`-equivalent trigger.
+	const stdout = createStdout();
+
+	function App({session}: {readonly session: number}) {
+		const items = session === 1 ? ['old-A', 'old-B'] : ['new-C', 'new-D'];
+		return (
+			<Box>
+				<Static key={session} items={items}>
+					{item => <Text key={item}>{item}</Text>}
+				</Static>
+				<Text>dynamic</Text>
+			</Box>
+		);
+	}
+
+	const {rerender} = render(<App session={1} />, {stdout, debug: true});
+
+	const afterFirstMount = (stdout.write as any).lastCall.args[0] as string;
+	t.true(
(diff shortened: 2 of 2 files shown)


src/reconciler.ts
```diff
@@ -295,7 +295,18 @@ export default createReconciler<
...
+		// Only clear staticNode if the node being removed is the one the root
+		// currently points at. React processes insertions before deletions in
+		// a single commit, so when <Static> is remounted via a `key` change the
+		// new staticNode is registered by createInstance BEFORE this removal
+		// fires for the old node. Clearing unconditionally would clobber the
+		// fresh pointer and the renderer would emit no static output for the
+		// new Static instance.
+		if (
+			removeNode.internal_static &&
+			currentRootNode &&
```
chiga0: This guard currently fails the repo lint (`npm run lint`) with `@typescript-eslint/prefer-optional-chain`. Please collapse it to `removeNode.internal_static && currentRootNode?.staticNode === removeNode` before merging. generated by GPT-5 model

src/reconciler.ts
```diff
@@ -351,7 +362,13 @@ export default createReconciler<
 		removeChildNode(node, removeNode);
 		cleanupYogaNode(removeNode.yogaNode);
 
-		if (removeNode.internal_static && currentRootNode) {
+		// See removeChildFromContainer above for the rationale on the
+		// `currentRootNode.staticNode === removeNode` guard.
+		if (
+			removeNode.internal_static &&
+			currentRootNode &&
```
chiga0: Same lint failure here: XO prefers optional chaining for this guard, so `npm run lint` currently exits non-zero until this is written as `removeNode.internal_static && currentRootNode?.staticNode === removeNode`. generated by GPT-5 model

src/reconciler.ts
chiga0: Fixed in 12f47b2 — collapsed both guards to `currentRootNode?.staticNode === removeNode`. `npm run lint` is now clean (only pre-existing warnings on master remain) and the new `remounting <Static> via key change emits the new items` test still passes.

src/reconciler.ts
chiga0: Same fix applied here in 12f47b2.

PR #950, opened 2026-05-12
Fix: drop stale <Static> output from fullStaticOutput on identity change

### LINKED ISSUE #949: fullStaticOutput retains items from unmounted/replaced <Static>, causing stale content to reappear on clearTerminal rewrites
## Summary

\`Ink#fullStaticOutput\` is the running concatenation of every chunk ever emitted by any \`<Static>\` instance during the lifetime of an Ink renderer. It is initialised to \`''\` in the constructor and only ever grows — there is no path that shrinks or resets it.

When ink later decides to rewrite the whole screen via \`shouldClearTerminalForFrame\` (output overflow, leaving fullscreen, unmount of a fullscreen frame), it writes:

\`\`\`
clearTerminal + this.fullStaticOutput + output
\`\`\`

That means: **any item ever emitted by a \`<Static>\` that has since been unmounted (or replaced via a key change) gets re-materialised above the current viewport on the next overflow-triggered rewrite**, even though the items no longer exist in the React tree.

## Minimal reproduction

\`\`\`tsx
import {render, Box, Static, Text} from 'ink';

function App() {
  const [showHistory, setShowHistory] = useState(true);
  const [counter, setCounter] = useState(0);

  useEffect(() => {
    // Unmount the history Static after a moment
    setTimeout(() => setShowHistory(false), 300);
    // Then keep growing dynamic output until ink decides to clearTerminal
    const id = setInterval(() => setCounter(c => c + 1), 25);
    return () => clearInterval(id);
  }, []);

  return (
    <Box flexDirection=\"column\">
      {showHistory && (
        <Static items={['old-A', 'old-B', 'old-C']}>
          {item => <Text key={item}>{item}</Text>}
        </Static>
      )}
      {Array.from({length: counter}, (_, i) => (
        <Text key={i}>dynamic line {i}</Text>
      ))}
    </Box>
  );
}

render(<App />);
\`\`\`

Expected: once \`<Static>\` is unmounted, \`old-A / old-B / old-C\` are gone from the screen. They were already written to stdout, and the terminal's own scrollback may still preserve them, but ink should not actively re-emit them.

Actual: when the dynamic area grows past the viewport and \`shouldClearTerminalForFrame=true\` fires, ink writes \`clearTerminal + fullStaticOutput + output\`. \`fullStaticOutput\` still contains \"old-A\nold-B\nold-C\n\", so those items reappear above the live region. From the user's point of view, content that was explicitly removed from the React tree comes back.

## Why this matters in practice

Long-lived TUIs that drive a \`<Static>\` replay by bumping a remount \`key\` after \`/clear\`, model-switch, Ctrl+O / compact-mode toggle, auth refresh, etc., see this constantly:

- Gemini CLI / qwen-code's \`historyRemountKey\` pattern: after \`/clear\`, the user expects history to be gone. The terminal viewport is wiped, but the moment a long AI response triggers \`shouldClearTerminalForFrame\`, ink re-emits the entire pre-\`/clear\` history above the current response. End-to-end reproduction thread: https://github.com/QwenLM/qwen-code/pull/4083
- The companion \"key-driven remount drops new items\" bug is being addressed separately in #947 / #948. This issue is **independent** of that one — it surfaces even on a pure unmount (no remount), as the repro above demonstrates.

## Root cause

\`ink.tsx\`:

\`\`\`ts
// constructor
this.fullStaticOutput = '';

// onRender (interactive frame path)
if (hasStaticOutput) {
  this.fullStaticOutput += staticOutput;
}

// renderInteractiveFrame (shouldClearTerminalForFrame=true path)
this.options.stdout.write(
  ansiEscapes.clearTerminal + this.fullStaticOutput + output,
);
\`\`\`

There is no code path that ever shrinks \`fullStaticOutput\`. The reconciler clears \`rootNode.staticNode\` on Static unmount (added by #905), which correctly stops the renderer from walking a freed Yoga tree, but the **already-accumulated string in \`fullStaticOutput\` is left in place** — and that's the part \`shouldClearTerminalForFrame\` replays.

## Suggested fix

When the live \`<Static>\` instance changes identity (first mount, last unmount, or remount via key), drop the accumulated output. The new instance (if any) re-emits its items into a fresh accumulator on the immediate render that follows.

I have a small PR with this fix + a regression test: vadimdemedes/ink#TBD (linking momentarily). The change is scoped to the identity-change cases — the classic Tap-style \"one Static mounted for the lifetime of the app\" use case is unchanged because \`staticNode\` identity never changes there.

## Environment

- ink \`7.0.2\` (master \`76d221c\`)
- Node 22.x, React 19.2.4
- macOS / Linux / Windows — platform-independent

## Related

- #904 / #905 — fixed the OOM / freed-Yoga path on unmount. This issue is the follow-up state-leak in the writer-side accumulator.
- #947 / #948 — fixed key-driven remount dropping new Static items. Companion bug; with both fixes applied, remount produces \"only new items, no resurrected old items\".

chiga0:
Fixes #949.

## Summary

\`Ink#fullStaticOutput\` is the running concatenation of everything every \`<Static>\` instance has ever emitted during the lifetime of an Ink renderer. It is initialised to \`''\` in the constructor and only ever grows — no code path shrinks or resets it. When \`shouldClearTerminalForFrame\` later fires (overflow, leaving fullscreen, fullscreen unmount), Ink writes \`clearTerminal + fullStaticOutput + output\`, which means any item emitted by a \`<Static>\` that has since been unmounted/replaced **re-materialises** above the current viewport even though it no longer exists in the React tree.

Real-world symptom: \"cleared history reappears after a long response scrolls the viewport\" in TUIs that drive a Static replay via a remount key (Gemini CLI, qwen-code, …). End-to-end repro thread: https://github.com/QwenLM/qwen-code/pull/4083.

## Fix

Track \`rootNode.previousStaticNode\` across commits. When \`resetAfterCommit\` detects that \`rootNode.staticNode\` has changed identity (first mount, last unmount, or remount via key), fire a new \`onStaticChange\` hook on the root node **before** \`onImmediateRender\`. Ink installs an \`onStaticChange\` handler that resets \`this.fullStaticOutput = ''\`. The new instance (if any) re-emits its items into a fresh accumulator on the immediate render that follows.

\`\`\`diff
 // resetAfterCommit, src/reconciler.ts
+  if (rootNode.staticNode !== rootNode.previousStaticNode) {
+    rootNode.previousStaticNode = rootNode.staticNode;
+    if (typeof rootNode.onStaticChange === 'function') {
+      rootNode.onStaticChange();
+    }
+  }
\`\`\`

\`\`\`diff
 // constructor wiring, src/ink.tsx
   this.rootNode.onImmediateRender = this.onRender;
+  this.rootNode.onStaticChange = this.handleStaticChange;
 ...
+  handleStaticChange = (): void => {
+    this.fullStaticOutput = '';
+  };
\`\`\`

## Scope (no behaviour change for the classic case)

- The **Tap-style \"one Static mounted for the lifetime of the app, items keep getting appended\"** use case is untouched: \`staticNode\` identity never changes → \`onStaticChange\` never fires → \`fullStaticOutput\` keeps accumulating exactly as today.
- The **#904 / #905 unmount path** is strengthened: in addition to clearing \`rootNode.staticNode\` (so the renderer stops walking a freed Yoga tree), Ink now also drops the now-orphan output from \`fullStaticOutput\`, so a later overflow-driven \`clearTerminal\` rewrite does not resurrect the unmounted Static's items.
- The original \`static output stops accumulating after Static unmounts (#904)\` regression test still passes — its length-stability assertion (\`outputAfterChurn.length === outputAfterUnmount.length\`) holds whether \`fullStaticOutput\` is left intact at unmount or reset at unmount; this PR doesn't change that invariant.

## Test

Adds \`fullStaticOutput is reset when <Static> unmounts so stale items are not replayed\` to \`test/components.tsx\`. Mounts a \`<Static items={['HISTORY-A', 'HISTORY-B']} />\` in debug mode (every \`stdout.write\` is \`fullStaticOutput + dynamicOutput\`), unmounts it, asserts that the subsequent write does NOT include \`HISTORY-A\` or \`HISTORY-B\`. Without the fix the assertion fails; with the fix it passes.

The **companion remount-via-key case** is intentionally NOT tested here because it additionally depends on the \`staticNode\`-pointer-clobbered-on-remount bug fixed by #948. Each fix stands alone; **this PR is independent of #948** and can be merged in either order. The two fixes compose to give the full \"remount produces only new items, no resurrected old items\" behaviour.

Verified:
- New test **fails** on \`master\` (\`76d221c\`) — confirms the bug
- New test **passes** with this patch
- Existing Static tests all pass (\`static output\`, \`skip previous output when rendering new static output\`, \`static output stops accumulating after Static unmounts (#904)\`, \`render only new items in static output on final render\`, non-interactive mode, concurrent mode)
- Other test failures in the full suite (\`cursor moves on space input\`, \`measure-element\`-related, \`wrap-ansi doesn't trim leading whitespace\`) reproduce on \`master\` without this patch and are unrelated.

## Why a hook on the root node and not inline in the reconciler

The reconciler owns the React → DOM mapping; \`fullStaticOutput\` is a private Ink-instance writer-side accumulator. A hook keeps the layering intact: the reconciler reports an identity change, Ink decides what to do with it. The same pattern is already used for \`onRender\`, \`onImmediateRender\`, and \`onComputeLayout\`.

### DIFF
--- a/test/components.tsx
+++ b/test/components.tsx
@@ -555,6 +555,60 @@ test('static output stops accumulating after Static unmounts (#904)', t => {
 	t.true(outputAfterChurn.includes('Dynamic'));
 });
 
+test('fullStaticOutput is reset when <Static> unmounts so stale items are not replayed', t => {
+	// In debug mode each stdout.write is `fullStaticOutput + dynamicOutput`.
+	// Before the fix, the items emitted by a Static that has since been
+	// unmounted stayed in `fullStaticOutput` forever and reappeared on every
+	// subsequent write (and on any `shouldClearTerminalForFrame` rewrite at
+	// runtime). After the fix, dropping the <Static> resets the
+	// accumulator so its items leave both the React tree and the output
+	// stream.
+	const stdout = createStdout();
+
+	function App({show, dynamicLabel}: {
+		readonly show: boolean;
+		readonly dynamicLabel: string;
+	}) {
+		return (
+			<Box>
+				{show ? (
+					<Static items={['HISTORY-A', 'HISTORY-B']}>
+						{item => <Text key={item}>{item}</Text>}
+					</Static>
+				) : null}
+				<Text>{dynamicLabel}</Text>
+			</Box>
+		);
+	}
+
+	const {rerender} = render(<App show dynamicLabel="d1" />, {
+		stdout,
+		debug: true,
+	});
+
+	const afterMount = (stdout.write as any).lastCall.args[0] as string;
+	t.true(
+		afterMount.includes('HISTORY-A') && afterMount.includes('HISTORY-B'),
+		'Static items must be emitted on first mount',
+	);
+
+	rerender(<App show={false} dynamicLabel="d2" />);
+
+	const afterUnmount = (stdout.write as any).lastCall.args[0] as string;
+	t.false(
+		afterUnmount.includes('HISTORY-A'),
+		'fullStaticOutput must NOT replay HISTORY-A after Static unmount',
+	);
+	t.false(
+		afterUnmount.includes('HISTORY-B'),
+		'fullStaticOutput must NOT replay HISTORY-B after Static unmount',
+	);
+	t.true(
+		afterUnmount.includes('d2'),
+		'new dynamic output must still render',
+	);
+});
+
 test('render only new items in static output on final render', t => {
 	const stdout = createStdout();
 
--- a/test/components.tsx
+++ b/test/components.tsx
@@ -565,7 +565,10 @@ test('fullStaticOutput is reset when <Static> unmounts so stale items are not re
 	// stream.
 	const stdout = createStdout();
 
-	function App({show, dynamicLabel}: {
+	function App({
+		show,
+		dynamicLabel,
+	}: {
 		readonly show: boolean;
 		readonly dynamicLabel: string;
 	}) {
@@ -603,10 +606,7 @@ test('fullStaticOutput is reset when <Static> unmounts so stale items are not re
 		afterUnmount.includes('HISTORY-B'),
 		'fullStaticOutput must NOT replay HISTORY-B after Static unmount',
 	);
-	t.true(
-		afterUnmount.includes('d2'),
-		'new dynamic output must still render',
-	);
+	t.true(afterUnmount.includes('d2'), 'new dynamic output must still render');
 });
 
 test('render only new items in static output on final render', t => {
--- a/src/dom.ts
+++ b/src/dom.ts
@@ -66,9 +66,17 @@ export type DOMElement = {
 	// Internal properties
 	isStaticDirty?: boolean;
 	staticNode?: DOMElement;
+	// Identity of `staticNode` at the end of the previous commit. Used by the
+	// reconciler to detect when the live <Static> instance has been replaced
+	// (key-driven remount), removed (no longer in the React tree) or added
+	// (first <Static> mount). On any identity change, ink resets its
+	// accumulated `fullStaticOutput` so future clearTerminal rewrites don't
+	// replay items that are no longer owned by the current React tree.
+	previousStaticNode?: DOMElement;
 	onComputeLayout?: () => void;
 	onRender?: () => void;
 	onImmediateRender?: () => void;
+	onStaticChange?: () => void;
 	internal_layoutListeners?: Set<LayoutListener>;
 } & InkNode;
 
--- a/src/ink.tsx
+++ b/src/ink.tsx
@@ -360,6 +360,7 @@ export default class Ink {
 		}
 
 		this.rootNode.onImmediateRender = this.onRender;
+		this.rootNode.onStaticChange = this.handleStaticChange;
 		this.log = logUpdate.create(options.stdout, {
 			incremental: options.incrementalRendering,
 		});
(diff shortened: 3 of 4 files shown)


test/components.tsx
```diff
@@ -555,6 +555,60 @@ test('static output stops accumulating after Static unmounts (#904)', t => {
...
+	// In debug mode each stdout.write is `fullStaticOutput + dynamicOutput`.
+	// Before the fix, the items emitted by a Static that has since been
+	// unmounted stayed in `fullStaticOutput` forever and reappeared on every
+	// subsequent write (and on any `shouldClearTerminalForFrame` rewrite at
+	// runtime). After the fix, dropping the <Static> resets the
+	// accumulator so its items leave both the React tree and the output
+	// stream.
+	const stdout = createStdout();
+
+	function App({show, dynamicLabel}: {
```
chiga0: This new test is not Prettier-formatted, which makes the repository lint step fail before AVA runs. I reproduced it with `npm run lint` and `npx prettier --check src/dom.ts src/ink.tsx src/reconciler.ts test/components.tsx`; Prettier reports this function signature here and the assertion formatting around line 606. Running Prettier on `test/components.tsx` should clear the blocking CI failure.

Generated By GPT-5 model

chiga0 (commented):
Review note submitted from local audit.

Generated By GPT-5 model

test/components.tsx
chiga0: Fixed in 7850f7e — ran `prettier --write test/components.tsx`, which reflowed the `App` props type signature and collapsed the final `t.true(...)` assertion onto one line. `npx prettier --check src/dom.ts src/ink.tsx src/reconciler.ts test/components.tsx` is now clean.

sindresorhus:
Can you fix the merge conflict?

chiga0:
> Can you fix the merge conflict?

It's all clean now. Please check again. 

By the way, cloud you please release package v7.0.3 after merged the pull request, thank you very much about that. Our package want to upgrade ink from v6.3 to v7, but find the issue. Thank you again.

PR #954, opened 2026-05-15
ci: run tests on Node 26

gameroman:


sindresorhus:
Thanks, but PRs like this are not very useful. It takes more time to check and reformat the pull request title than just doing it myself.

PR #955, opened 2026-05-15
Add onClick event support

msfeldstein:
## Summary

- Add terminal mouse tracking and SGR mouse parsing to support `onClick` on Ink `Box` and `Text` elements.
- Dispatch click events to the deepest matching node, then bubble through parents with `stopPropagation()` support.
- Add focused tests plus an `examples/on-click` playground for manual testing.

## Flow

```text
stdin mouse escape
  -> parse SGR mouse event
  -> find deepest rendered node containing x/y
  -> call target onClick
  -> bubble to parent onClick handlers unless stopped
```

## Design notes

- Enables xterm mouse modes 1000 + 1006 (SGR) only when the rendered tree contains an `onClick` handler.
- Parses left-button press events only; other mouse escapes are swallowed so they do not leak into `useInput`.
- Hit testing walks children in reverse DOM order so overlapping absolute boxes resolve to the topmost target.

## Test plan

- [x] `npm run lint` (passes with existing warnings)
- [x] `npm run typecheck`
- [x] `FORCE_COLOR=true npx ava test/components.tsx --match='*onClick*' --match='*mouse tracking*'`
- [x] Manual: `npm run example -- examples/on-click/index.ts`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/examples/on-click/index.ts
+++ b/examples/on-click/index.ts
@@ -1,0 +1,1 @@
+import './on-click.js';
--- a/examples/on-click/on-click.tsx
+++ b/examples/on-click/on-click.tsx
@@ -110,31 +110,36 @@ function OnClickDemo() {
 						position="absolute"
 						left={0}
 						top={0}
-						borderStyle="round"
-						borderColor="red"
-						paddingX={2}
+						width={18}
+						height={3}
+						backgroundColor="red"
+						paddingLeft={2}
+						paddingTop={1}
 						onClick={event => {
 							setBottomOverlayClicks(count => count + 1);
 							describeClick('Bottom overlay', event);
 						}}
 					>
-						<Text color="red">Bottom overlay</Text>
+						<Text color="white">Bottom overlay</Text>
 					</Box>
 
 					<Box
 						position="absolute"
 						left={4}
 						top={0}
+						width={18}
+						height={3}
 						backgroundColor="green"
-						borderStyle="round"
-						borderColor="green"
-						paddingX={2}
+						paddingLeft={2}
+						paddingTop={1}
 						onClick={event => {
 							setTopOverlayClicks(count => count + 1);
 							describeClick('Top overlay', event);
 						}}
 					>
-						<Text color="black">Top overlay</Text>
+						<Text bold color="black">
+							Top overlay
+						</Text>
 					</Box>
 				</Box>
 			</Box>
--- a/examples/on-click/on-click.tsx
+++ b/examples/on-click/on-click.tsx
@@ -1,0 +1,114 @@
+import React from 'react';
+import {Box, Text, render, useApp, useInput} from '../../src/index.js';
+
+function OnClickDemo() {
+	const {exit} = useApp();
+	const [primaryClicks, setPrimaryClicks] = React.useState(0);
+	const [secondaryClicks, setSecondaryClicks] = React.useState(0);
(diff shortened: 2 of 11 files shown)


sindresorhus:
Thanks for contributing, but I don't think this should land as-is.

The main issue is that the hit testing assumes Ink layout coordinates are the same as terminal screen coordinates. In the normal interactive path, Ink does not know or track the frame's absolute terminal origin. It writes the frame at the current cursor position and updates it relative to the previous frame, while SGR mouse coordinates are absolute screen coordinates. So clicks will be offset or just hit the wrong element depending on where the app was rendered. This only seems safe when Ink owns the whole viewport, like alternate screen or a fullscreen clear.

There is also a concrete parser bug: SGR wheel-up is button code 64, and this code turns that into button 0 with `% 4`, so scrolling over a clickable element can accidently fire `onClick`. I verified this manually: `\u001B[<64;1;1M` parses as the same left-click event as `\u001B[<0;1;1M`.

More broadly, enabling mouse tracking just because any element has `onClick` is too wide of a contract. Mouse mode changes terminal behavior globally, especially around scroll behavior, but this implementation only supports a narrow subset and discards the rest.

I think the smallest sane version would be to either not add this yet, or make it explicitly limited to alternate-screen/fullscreen apps where Ink owns the viewport. If that route is taken, the wheel parsing bug needs to be fixed first, and the supported scope should be documented clearly.

I ran the test suite locally and it passes, but the tests don't cover the coordinate-origin problem because the fake terminal effectively starts the app at row 0.

msfeldstein:
Good call on both counts.  I pushed to only enable it for ASB and fixed the scroll-causes-click issue, and verified it happened on the example before but not after the fix. 

At a higher level though, is mouse events something that ink does want to add?  I'm happy to take any feedback on more rigorous design principles of the API to make sure its done right, this is my first time really looking at the inner workings

msfeldstein:
and if mouse events are a non goal i can always just run a private fork, just let me know

sindresorhus:
I re-checked this manually and I still don't think it can land as-is.

The alternate-screen restriction fixes the original "random terminal origin" problem, and the wheel event case looks fixed. However, hit testing is still wrong when the app uses `<Static>`. Static output is prepended outside the dynamic layout that the mouse hit tester walks, so the visible terminal rows no longer match the Yoga coordinates. I reproduced this with one static line above a clickable dynamic line: clicking the visible dynamic row did nothing, while clicking the static row incorrectly fired the dynamic click handler.

There's also a terminal-state cleanup issue when stdout is a TTY but stdin cannot use raw mode. The code enables mouse tracking before raw mode setup succeeds. I reproduced a case where the enable sequence was written, but no matching disable sequence was written. That can leave the user's terminal in mouse tracking mode after the app errors.

The focused click tests pass, but they don't cover these cases. At minimum, this needs to either make these cases impossible before enabling mouse tracking, or document a much narrower supported contract and enforce it in code. I think the simple version is: require usable raw-mode stdin before writing any mouse-mode escape, and don't support click handling together with static output unless the coordinate offset is explicitely accounted for.

msfeldstein:
Makes sense, i ended up vendoring for now, but if this is of interest to upstream im happy to keep sheparding it through.  Closing for now but let me know otherwise

msfeldstein:
Really appreciate the deep review.  I'm still learning how a lot of the ink system works an this is helpful

PR #965, opened 2026-05-28
fix: handle incomplete stack frames in error overview

### LINKED ISSUE #963: ErrorOverview: duplicate keys on empty stack lines + incomplete parseLine guard for native frames
## Environment

- ink version: 5.2.1 (also present in 6.x — confirmed in bun cache)
- stack-utils version: 2.0.6
- React version: 18.3.x

## Problem

`ErrorOverview` has two bugs that can cause a secondary crash when rendering a primary render error, potentially preventing the error overlay from displaying at all.

### Bug 1: Duplicate `key={line}` on empty/trailing newlines

`error.stack.split('\n')` commonly yields empty strings from trailing newlines. All empty strings map to `key=""`, triggering a React duplicate-key warning and potential misreconciliation during concurrent updates:

```js
// ErrorOverview.js:57-76 (stack trace section)
.map(line => {
    // ...
    return React.createElement(Box, { key: line }, ...);  // key="" when line === ""
    // ...
    return React.createElement(Box, { key: line }, ...);  // key="" again → duplicate
})
```

Observed warning:
```
Warning: Encountered two children with the same key, ``.
    at ErrorOverview
```

### Bug 2: Incomplete guard for truthy-but-incomplete `parseLine` returns

`stack-utils@2.0.6`'s `parseLine` has a **three-valued return contract** that `ErrorOverview` treats as two-valued:

| `parseLine` returns | ErrorOverview treats as | Actual behavior |
|---|---|---|
| `null` | Unparseable → renders raw text | ✓ Correct |
| `{function, file, line, column, ...}` | Full frame → renders structured | ✓ Correct |
| Truthy incomplete object (e.g. `{native: true}`) | *Same as full frame* | ✗ Renders with `undefined` props |

`stack-utils` returns `{native: true}` for frames like `at native` — truthy, but missing `function`, `file`, `line`, and `column`. The current guard only catches `null`/`undefined`:

```js
const parsedLine = stackUtils.parseLine(line);
if (!parsedLine) {           // ← misses {native: true} and other partial shapes
    return ...raw line...;
}
// Falls through here with parsedLine.function === undefined, parsedLine.file === undefined, etc.
return React.createElement(Box, { key: line },
    React.createElement(Text, {}, parsedLine.function),    // undefined
    React.createElement(Text, {}, cleanupPath(parsedLine.file), ":", parsedLine.line, ...));
```

When both bugs combine (duplicate-key misreconciliation + `undefined` fiber type), the result is a crash in React's reconciler: `t.type is undefined` in compiled/minified builds.

## Proposed Fix

```js
.map((line, index) => {
    const parsedLine = stackUtils.parseLine(line);
    // Guard against null AND truthy-but-incomplete objects ({native: true}, etc.)
    if (!parsedLine || (!parsedLine.file && !parsedLine.function)) {
        return (React.createElement(Box, { key: index },   // ← index, not line
            React.createElement(Text, { dimColor: true }, "- "),
            React.createElement(Text, { dimColor: true, bold: true }, line)));
    }
    return (React.createElement(Box, { key: index },       // ← index, not line
        React.createElement(Text, { dimColor: true }, "- "),
        React.createElement(Text, { dimColor: true, bold: true }, parsedLine.function),
        React.createElement(Text, { dimColor: true, color: "gray" },
            ' ', "(", cleanupPath(parsedLine.file) ?? '', ":", parsedLine.line, ":", parsedLine.column, ")")));
});
```

Changes:
- `key={line}` → `key={index}` on both branches — prevents duplicate keys from empty/identical lines
- Guard extended to `!parsedLine || (!parsedLine.file && !parsedLine.function)` — catches `{native: true}` and any other partial shapes that lack the properties the structured branch renders

## Workaround

We worked around this in our project by wrapping `<App>` in a custom `ErrorBoundary` class component that intercepts errors before Ink's `App.getDerivedStateFromError` runs, rendering a safe fallback that avoids `parseLine` entirely.

cyphercodes:
## Summary
- Render `stack-utils` incomplete stack frames (for example `at native`) as raw stack lines instead of structured frames with `undefined` fields.
- Give repeated stack lines unique, stable keys so trailing/empty stack lines do not trigger React duplicate-key warnings.
- Add focused `ErrorOverview` regression coverage for native frames and repeated blank stack lines.

Closes #963

## Test Plan
- `FORCE_COLOR=true npx ava test/error-overview.tsx --serial`
- `FORCE_COLOR=true npx ava test/errors.tsx test/error-overview.tsx --serial`
- `npm run typecheck`
- `npm run lint` (passes with existing warnings in unrelated files)
- `npm run build`
- `git diff --check`

Note: I also ran `npm test`; it reached the broad AVA suite but failed in unrelated existing tests (`alternate-screen-example`, `cursor`, and `measure-element`) outside the touched ErrorOverview surface.

### DIFF
--- a/src/components/ErrorOverview.tsx
+++ b/src/components/ErrorOverview.tsx
@@ -27,6 +27,7 @@ export default function ErrorOverview({error}: Props) {
 	const filePath = cleanupPath(origin?.file);
 	let excerpt: CodeExcerpt[] | undefined;
 	let lineWidth = 0;
+	const stackLineCounts = new Map<string, number>();
 
 	if (filePath && origin?.line && fs.existsSync(filePath)) {
 		const sourceCode = fs.readFileSync(filePath, 'utf8');
@@ -96,11 +97,15 @@ export default function ErrorOverview({error}: Props) {
 						.slice(1)
 						.map(line => {
 							const parsedLine = stackUtils.parseLine(line);
+							const lineCount = stackLineCounts.get(line) ?? 0;
+							stackLineCounts.set(line, lineCount + 1);
+							const key = `${line}-${lineCount}`;
 
-							// If the line from the stack cannot be parsed, we print out the unparsed line.
-							if (!parsedLine) {
+							// If the line from the stack cannot be parsed, or parsed into an incomplete
+							// frame (for example, "at native"), we print out the unparsed line.
+							if (!parsedLine || (!parsedLine.file && !parsedLine.function)) {
 								return (
-									<Box key={line}>
+									<Box key={key}>
 										<Text dimColor>- </Text>
 										<Text dimColor bold>
 											{line}
@@ -111,7 +116,7 @@ export default function ErrorOverview({error}: Props) {
 							}
 
 							return (
-								<Box key={line}>
+								<Box key={key}>
 									<Text dimColor>- </Text>
 									<Text dimColor bold>
 										{parsedLine.function}
--- a/src/components/ErrorOverview.tsx
+++ b/src/components/ErrorOverview.tsx
@@ -102,8 +102,9 @@ export default function ErrorOverview({error}: Props) {
 							const key = `${line}-${lineCount}`;
 
 							// If the line from the stack cannot be parsed, or parsed into an incomplete
-							// frame (for example, "at native"), we print out the unparsed line.
-							if (!parsedLine || (!parsedLine.file && !parsedLine.function)) {
+							// frame without source location data (for example, "at native"), we print
+							// out the unparsed line.
+							if (!parsedLine?.file || !parsedLine.line || !parsedLine.column) {
 								return (
 									<Box key={key}>
 										<Text dimColor>- </Text>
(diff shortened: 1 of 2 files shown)


sindresorhus:
This still misses named native frames. I manually checked `parseLine('    at foo (native)')`, and it returns `{native: true, function: 'foo'}`. Since `function` is present, the new guard takes the structured frame path even though there is no file, line, or column.

Rendering that stack line currently produces `- foo (::)`, and the aria label includes `line undefined column undefined`.

I think the guard should treat frames without source location data as incomplete too, not just frames where both `file` and `function` are missing. Otherwise this fix covers `at native`, but not `at foo (native)`.

cyphercodes:
Updated this PR to handle named native frames like `at foo (native)` as incomplete stack frames too, so they render as the original raw stack line instead of `foo (::)` / undefined source-location metadata.

Local verification:
- `npm run typecheck` ✅
- `npx xo src/components/ErrorOverview.tsx test/error-overview.tsx` ✅
- `FORCE_COLOR=true npx ava test/error-overview.tsx` ✅
- `git diff --check` ✅

PR #968, opened 2026-06-02
feat: extend measureElement() to return position coordinates

chiga0:
## Summary

Extend `measureElement()` to return `{x, y, width, height}` — adding the element's position within the live layout region to the existing dimension output.

### Motivation

`measureElement()` previously returned only `{width, height}`, and `useBoxMetrics()` returns position relative to the parent. Neither provides coordinates for comparing element positions across the layout tree — needed for mouse interaction:

- **Hit-testing**: determining which element a click or scroll event targets
- **Scrollbar drag**: tracking mouse position against a scrollbar column
- **Hover/tooltip**: positioning UI relative to the cursor

### Coordinate contract

`x` and `y` are **layout-tree coordinates** — the element's position within the live layout region, computed by walking up the yoga parent chain and accumulating offsets.

- In **alternate-screen mode**, layout coordinates equal viewport coordinates directly.
- In **inline mode**, the caller subtracts the live region's viewport anchor from the mouse event's row to convert between coordinate spaces.
- `<Static>` uses `position: absolute` internally, so it does not affect sibling layout positions.

```ts
const metrics = measureElement(ref.current);
// metrics.x — column within the live layout region
// metrics.y — row within the live layout region
// metrics.width, metrics.height — element dimensions
```

### API

```ts
measureElement(node: DOMElement): { x: number; y: number; width: number; height: number }
```

Backward-compatible — existing callers that destructure `{width, height}` continue to work unchanged.

### Tests

11 tests covering:
- Root-level element → `(0, 0)`
- Nested element with `paddingLeft` → correct x offset
- Deeply nested element → offsets accumulate through ancestors
- Margin offset → `marginLeft` / `marginTop` reflected
- Sibling offset → second child gets correct y
- `<Static>` does not affect layout coordinates (position: absolute)
- No yoga node → returns `{x: 0, y: 0, width: 0, height: 0}`
- Existing dimension tests (measure after state update, throttled render, etc.)

### Prior art

`@jrichman/ink` (used by gemini-cli) exports a similar `getBoundingBox()` API. This PR brings equivalent functionality to stock ink by extending the existing `measureElement()` API, as suggested by @sindresorhus.


sindresorhus:
I think this should be folded into `measureElement()` instead of adding a separate `getBoundingBox()`.

`measureElement()` is already the imperative layout measurement API. Expanding it to return `{x, y, width, height}` gives callers one concept: measure this element. `getBoundingBox()` adds another top-level API that partly overlaps with `measureElement()`, and users then have to choose between dimensions-only and box-with-position even though both are the same layout snapshot.

The important bit is the coordinate contract. If `x`/`y` are meant for mouse hit testing, they need to be terminal/viewport coordinates for the rendered frame, not just Yoga offsets accumulated through parents. I manually checked the `Static` case: render one `Static` line above a live `Box` and the live `Box` is visually on the next terminal row, but this implementation reports `0,0`. So as written it is not actually absolute screen position for a supported rendering mode.

I’d suggest changing `measureElement()` to return:

```ts
type ElementMetrics = {
	x: number;
	y: number;
	width: number;
	height: number;
};
```

Then `useBoxMetrics()` can be the reactive version of the same shape. That gives Ink one measurement model instead of three related ones: `measureElement()` width/height, `useBoxMetrics()` left/top, and `getBoundingBox()` x/y. Easier to explain adn harder to misuse.

chiga0:
Thanks for the feedback — agreed on all points.

I've folded the logic into `measureElement()` so it now returns `{x, y, width, height}` instead of adding a separate API. The `getBoundingBox()` export is removed.

Regarding the `<Static>` coordinate issue you raised: you're right that the current implementation only accumulates Yoga layout offsets and doesn't account for static output lines above the live region. I'll address that in a follow-up once I dig into how the renderer tracks the static line count — it'll need a different mechanism than just walking yoga nodes.

The updated code is in the latest push.

chiga0:
> I think this should be folded into `measureElement()` instead of adding a separate `getBoundingBox()`.
> 
> `measureElement()` is already the imperative layout measurement API. Expanding it to return `{x, y, width, height}` gives callers one concept: measure this element. `getBoundingBox()` adds another top-level API that partly overlaps with `measureElement()`, and users then have to choose between dimensions-only and box-with-position even though both are the same layout snapshot.
> 
> The important bit is the coordinate contract. If `x`/`y` are meant for mouse hit testing, they need to be terminal/viewport coordinates for the rendered frame, not just Yoga offsets accumulated through parents. I manually checked the `Static` case: render one `Static` line above a live `Box` and the live `Box` is visually on the next terminal row, but this implementation reports `0,0`. So as written it is not actually absolute screen position for a supported rendering mode.
> 
> I’d suggest changing `measureElement()` to return:
> 
> ```ts
> type ElementMetrics = {
> 	x: number;
> 	y: number;
> 	width: number;
> 	height: number;
> };
> ```
> 
> Then `useBoxMetrics()` can be the reactive version of the same shape. That gives Ink one measurement model instead of three related ones: `measureElement()` width/height, `useBoxMetrics()` left/top, and `getBoundingBox()` x/y. Easier to explain adn harder to misuse.

@sindresorhus could you please check the PR again?

chiga0:
@sindresorhus @technicalpickles  hi, cloud you please take your time to review this pr. I had udpate the implemenation following the @sindresorhus 's suggestion. If ok, can you please merge the pr and publish a new version? I need this method to implemente the scroll bar in virtual box. thank you all again.

sindresorhus:
There’s still one contract issue to resolve before merging. The current implementation returns layout-tree coordinates, not terminal/viewport coordinates. I manually checked the `<Static>` case again: with one `<Static>` line above a live `<Box>`, the live box is rendered on the next terminal row, but `measureElement()` still reports `0,0`. So this still is not usable for mouse hit-testing in a supported rendering mode, which is the main use case described here.

I’d either make `x`/`y` actual rendered-frame coordintes, including static output above the live region, or narrow the contract and docs to say these are only layout-tree coordinates and not suitable for terminal mouse hit-testing. Given the PR motivation is scrollbars/mouse input, I think this should be fixed here rather than left for a follow-up.

Small docs thing too: the public docs still describe `measureElement()` as returning only `{width, height}`. They should be updated to the new shape if this lands.

chiga0:
@sindresorhus Thanks for the follow-up — both issues are addressed in the latest push:

**1. `<Static>` coordinate fix**

`measureElement()` now returns rendered-frame coordinates. The `y` value includes the number of lines flushed by `<Static>` above the live region:

- The `Ink` class tracks `internal_staticOutputHeight` on the root DOM node, updated every time `fullStaticOutput` changes (and reset to 0 on `<Static>` identity change).
- `measureElement()` walks up to the root node and adds that offset to `y`.

A test covers the exact case you described: two `<Static>` items above a live `<Box>` → `y` reports `2`, not `0`.

**2. Docs updated**

The readme now documents the `{x, y, width, height}` return shape and mentions the `<Static>` offset behavior.

sindresorhus:
The `y` contract still isn't right.

The static-line offset only yields correct viewport coordinates while the whole output fits on screen. `internal_staticOutputHeight` accumulates every `<Static>` line ever flushed and is never bounded to the viewport, so:

```
y = cumulativeStaticLines + layoutOffset
```

Mouse SGR coordinates are reported relative to the visible viewport (row 1 = top visible row), not total scrollback. Once cumulative output exceeds the terminal height (the normal regime for `<Static>`, since it exists for streaming logs), those lines scroll off the top and `y` stops mapping to any on-screen row. Example: 30-row terminal, 100 static log lines, 5-row live region. On screen the live element is around viewport row 25, but `measureElement()` returns `y = 100 + offset`, and it drifts further without bound. So the case the API is meant for is exactly the one it gets wrong.

This isnt a tuning bug: Ink redraws the live region at the cursor's current position and never tracks an absolute viewport row, so it can't recover the on-screen row by counting static lines.

Other TUI libs split two ways. Full-screen ones (Textual, Ratatui, blessed, prompt_toolkit) own the alternate screen, so mouse coords are buffer coords and scrolling is internal state. Inline ones (Ink's case) anchor to the live frame's actual cursor row via a Cursor Position Report (`ESC[6n`): Textual's inline mode asks the terminal where the cursor is and subtracts it from the mouse coordinate; Ratatui's inline viewport does the same. None derive the anchor from “lines printed so far,” because that breaks under scroll.

Two ways forward:

1. Robust: report `y` relative to the live region's actual cursor row (`ESC[6n`) and drop `internal_staticOutputHeight`. Correct even with scrolled logs.
2. Simple: define `y` as the row within the live layout tree (what the parent walk produces before adding static height) and document it as viewport-accurate only when output hasn't scrolled.

I'm not sure which is correct here.

chiga0:
@sindresorhus You're right — the cumulative static line count breaks once output scrolls past the viewport, which is exactly the normal regime for `<Static>`. Fixed in the latest push.

**What changed:**

Dropped `internal_staticOutputHeight` entirely. `x`/`y` are now defined as **layout-tree coordinates** — the element's position within the live layout region, computed purely from the yoga parent walk.

- In **alternate-screen mode** (the typical setup for mouse-interactive apps), layout coordinates equal viewport coordinates directly.
- In **inline mode**, the caller subtracts the live region's viewport anchor from the mouse event's row.
- `<Static>` uses `position: absolute` internally, so it never affects sibling layout positions — no offset needed.

This gives a stable, well-defined coordinate contract that's correct in all cases. Viewport-absolute coordinates via `ESC[6n` (Cursor Position Report) would be a separate concern for a future API.

Docs and tests updated — there's a test explicitly verifying that `<Static>` items don't shift the live element's `y`.

chiga0:
Hi @sindresorhus @vadimdemedes — friendly follow-up on this PR.

The latest push addresses the coordinate contract concern: `x`/`y` are now pure layout-tree coordinates (no cumulative static offset), with docs and tests updated accordingly. Would love your thoughts when you have a moment.

Thanks for your time!

PR #979, opened 2026-07-27
Fix dangling staticNode reference on indirect subtree removal

chiga0:
Follow-up to #904 / #905.

## Problem

PR #905 fixed the case where `<Static>` is **directly** removed — the `removeNode.internal_static` identity check clears `staticNode`. But when an **ancestor** of `<Static>` is removed (e.g. a parent `<Box>` is conditionally unmounted), `removeChild` receives the ancestor, whose `internal_static` is `false`. The check skips, `freeRecursive()` frees the entire subtree including the static node's Yoga WASM memory, and `staticNode` survives as a dangling reference.

The next render then calls `node.staticNode.yogaNode.getComputedWidth()` in `renderer.ts`. The JS wrapper object is still truthy after `free()` (yoga-layout does not invalidate it), so the `?.yogaNode` optional chain passes — and the WASM call traps with `RuntimeError: memory access out of bounds`.

This was reported in production by QwenLM/qwen-code#6820 (Qwen Code CLI), where `transcriptFreeze` toggles between `<App/>` (which contains `<Static>`) and `<TranscriptView/>`, unmounting the entire `<App/>` subtree.

## Fix

Two changes in `reconciler.ts`:

1. **`clearStaticNodeIfContained()`** — called before `removeChildNode` (while the parent chain is still intact), walks up from `staticNode` to check whether `removeNode` is an ancestor. If so, clears `staticNode` so the renderer skips the static output path. This subsumes the previous direct-removal identity check.

2. **`removeNode.yogaNode = undefined`** — after `freeRecursive()`, nulls out the freed WASM pointer on the removed DOM node. This is defense-in-depth: any stale JS reference to the removed node will short-circuit on `?.yogaNode` instead of trapping into invalid WASM memory.

## Test

Added a regression test in `test/components.tsx` that:
1. Renders `<Static>` nested inside a `<Box>` wrapper
2. Unmounts the wrapper (not `<Static>` directly)
3. Confirms the renderer does not crash and produces correct output on subsequent rerenders

All existing Static-related tests continue to pass (direct unmount, key-driven remount via both `removeChild` and `removeChildFromContainer` paths, `fullStaticOutput` reset).

### DIFF
--- a/src/dom.ts
+++ b/src/dom.ts
@@ -184,6 +184,32 @@ export const removeChildNode = (
 	}
 };
 
+const nullifyYogaNodes = (node: DOMNode): void => {
+	node.yogaNode = undefined;
+
+	if (node.nodeName !== '#text') {
+		for (const childNode of node.childNodes) {
+			nullifyYogaNodes(childNode);
+		}
+	}
+};
+
+/**
+Free the Yoga (WASM) memory of a removed subtree, then drop the `yogaNode`
+reference on every DOM node within it.
+
+`freeRecursive()` releases the WASM memory but leaves each JS wrapper object
+truthy, so every `?.yogaNode` guard in the codebase would still pass and then
+trap on a use-after-free when the wrapper is dereferenced (see
+QwenLM/qwen-code#6820). Nulling the references makes those guards effective and
+turns any lingering access into a safe no-op.
+*/
+export const freeYogaSubtree = (removeNode: DOMNode): void => {
+	removeNode.yogaNode?.unsetMeasureFunc();
+	removeNode.yogaNode?.freeRecursive();
+	nullifyYogaNodes(removeNode);
+};
+
 export const setAttribute = (
 	node: DOMElement,
 	key: string,
--- a/src/reconciler.ts
+++ b/src/reconciler.ts
@@ -83,6 +83,44 @@ const cleanupYogaNode = (node?: YogaNode): void => {
 	node?.freeRecursive();
 };
 
+/**
+ * Clear `staticNode` (and its change-detection counterpart `previousStaticNode`)
+ * when the node it points at is being removed as part of a larger subtree.
+ *
+ * The existing identity check (`staticNode === removeNode`) only catches direct
+ * removal of the `<Static>` element. When an *ancestor* of `<Static>` is
+ * removed, `freeRecursive()` frees the static node's Yoga WASM memory but the
+ * stale `staticNode` reference survives, and the next render calls
+ * `getComputedWidth()` on freed memory → `RuntimeError: memory access out of
+ * bounds` (see QwenLM/qwen-code#6820).
+ *
+ * Must be called BEFORE `removeChildNode` breaks the parent chain.
+ */
+const clearStaticNodeIfContained = (
+	removeNode: DOMElement | TextNode,
+): void => {
+	const staticNode = currentRootNode?.staticNode;
+
+	if (!staticNode) {
+		return;
+	}
+
+	// Walk up from staticNode to see if removeNode is an ancestor.
+	let current: DOMElement | undefined = staticNode;
+
+	while (current) {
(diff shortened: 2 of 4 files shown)


sindresorhus:
There are two issues here:

1. **The cached static node is tied to a global root.**

The helper checks whichever root most recently created `<Static>`. With two Ink instances using different stdout streams, removing `<Static>` from the first root after rendering one in the second leaves the first root's pointer stale. I reproduced stale output (`first-history\nfirst-2`), and further Yoga memory churn could still trigger the reported use-after-free.

Please derive the owning root from the host parent passed to `removeChild`, use the container directly for `removeChildFromContainer`, and add a two-root regression test.

2. **CI is currently failing.**

The new assertion is rejected by Prettier, so `npm test` exits before running the tests. Please format it and rerun CI.

One smaller concern: clearing `yogaNode` only on the removed top-level node does not invalidate descendants freed by `freeRecursive()`. I would either centralize that cleanup or remove this broader defense-in-depth mutation.

The test should also cover ancestor removal directly from the root, plus screen-reader and concurent rendering.

chiga0:
@sindresorhus Thanks for the thorough review — all points addressed in the latest push.

**1. Owning root derived from arguments, not a global**

`clearStaticNodeIfContained` now takes the root explicitly. It's derived from the removal hook's host parent (`findRootNode(node)` walks up to `ink-root`) rather than the module-global `currentRootNode`. `removeChildFromContainer` passes the container directly. So two instances on separate stdout streams no longer clobber each other's pointer.

Added a two-instance regression test that reproduces your `first-history\nfirst-2` stale output and confirms it no longer replays.

**2. CI / Prettier**

Fixed the formatting that was aborting `npm test` before the suite ran.

**3. The `yogaNode` cleanup — centralized instead of partial**

You're right that nulling only the top node left descendants dangling. I removed that partial mutation and centralized it: `freeYogaSubtree()` in `dom.ts` frees the subtree's WASM memory and then nulls `yogaNode` on **every** DOM node in the removed subtree.

This targets the actual root cause — `freeRecursive()` leaves each JS wrapper truthy, defeating the `?.yogaNode` guards throughout the codebase. Nulling the whole subtree makes those guards effective and closes the entire use-after-free class, not just the reported instance.

Clearing `staticNode` is still needed alongside it — otherwise the stale pointer suppresses the `onStaticChange` reset and replays stale static output (and the screen-reader path reads `staticNode` without a `?.yogaNode` guard at all).

**4. Additional test coverage**

Added tests for: ancestor removal directly from the root (`removeChildFromContainer` path), the two-instance case above, screen-reader mode, and concurrent mode.

sindresorhus:
I found one issue:

The dirty flag is still set on the last root that created a `<Static>` node, instead of on the root being updated. With two Ink instances, render a `<Static>` in both, trigger a normal update in the first instance, then append an item to its `<Static>` during the same throttled window. The new item gets removed before the first root gets its immediate render, so it never reaches stdout.

This is reproducible with the default `maxFps`. The first instance keeps rendering its old static item and the new dynamic frame, but drops the newly added static item. Please derive the owning root from the updated node and add a regression test for updating `<Static>` in one instance after another instance has mounted `<Static>`. The existing two-instance test only covers ancestor removal, so it doesn’t excercise this update path.

chiga0:
Thanks — fixed in 8f6e998.

**Dirty flag now derived from the updated node**

The `currentRootNode` module-level global is removed entirely. All three call sites (`commitUpdate`, `removeChild`, `removeChildFromContainer`) now derive the owning root by walking `parentNode` up to the `ink-root` node, so `isStaticDirty` always lands on the root that owns the updated `<Static>` node.

**Regression test added**

New test covers the exact repro you described: render `<Static>` in two instances, then append to the first instance's `<Static>` after the second instance mounted one. Before the fix the appended item was dropped; the test now asserts it reaches stdout, and that the second instance's updates remain independent.

Issue #765, opened 2025-09-11
Support scrolling primitives


jacob314:
# Support overflow="scroll"

Currently the foundational ink `<Box>` components support overflow="hidden" and overflow="visible" but not overflow="scroll".
This makes it difficult for CLI Chat applications such as Gemini CLI that need to constrain large amounts of text to fit within the visible area of the terminal while  needing to show easy to understand visual indicators of how much content scrolled off the screen. This is crucial to avoid flicker and to ensure the UI is by default scrolled to the most relevant information. Gemini CLI partially solves this with the workarounds of [MaxSizedBox](https://github.com/google-gemini/gemini-cli/blob/78744cfbca8872af1fdfcdccd5d0e70a008ce684/packages/cli/src/ui/components/shared/MaxSizedBox.tsx#L99) but this implementation is much more fragile and limited than a scrolling solution directly integrated into Ink.
Fortunately the Yoga layout model that Ink uses has support for [YGOverflowScroll](https://github.com/facebook/yoga/blob/dc2581f229cb05c7d2af8dee37b2ee0b59fd5326/yoga/enums/Overflow.h#L21) suggesting there are idiomatic and efficient ways this can be implemented.

# Proposed Ink Scrolling Feature Implementation Plan

This document outlines the plan and implementation details for adding enhanced scrolling capabilities to Ink's `<Box>` component.

## 1. Goal

The primary goal was to introduce a robust scrolling mechanism directly into the native Ink `<Box>` component, inspired by the overflow scroll model on the web.

-   **Scrollable Overflow:** A `<Box>` with `overflow: 'scroll'` should not expand beyond its flexbox-defined height. Instead, its content should be scrollable.
-   **Virtual Scrollbar:** A visual, non-interactive scrollbar should be rendered on the right-hand side of the box to indicate the scroll position and the amount of overflow.
-   **Initial Scroll Position:** The component should accept a prop to define whether the initial scroll position is at the top or the bottom of the content.
-   **Programmatic Control:** An imperative API should be available to programmatically get scroll information (like `scrollHeight`) and set the scroll position.
-   **Overflow Indicators** Optional lines at the top and bottom of the box to indicate that there is more content hidden above or below the visible area.
-   **Horizontal Scrolling:** Support for horizontal scrolling with a horizontal scrollbar.

### Non goals for the initial implementation
- Supporting lazy list rendering to increase efficiency for very large scrollable lists.
[...]

PR #988, opened 2026-08-09
Add scrolling primitives: `contentOffsetX`/`contentOffsetY` props and client/scroll size metrics

costajohnt:
This implements the scrolling primitives sketched in #765, following the direction there: core primitives only, so a `ScrollView` can be built in userland (example included in the readme).

- `contentOffsetX` / `contentOffsetY` props on `Box`. Children are shifted left/up by the given number of columns/rows at render time. Combined with `overflow="hidden"`, this is the scroll mechanism. Layout is untouched, the offset is applied in `renderNodeToOutput` when recursing into children, so Yoga sees nothing new.
- `measureElement()` now also returns `clientWidth`/`clientHeight` (size excluding borders) and `scrollWidth`/`scrollHeight` (content extent including overflow, computed from child layout extents and clamped to be at least the client size, matching DOM semantics). These extend the `x`/`y`/`width`/`height` shape from #968.
- `useBoxMetrics()` returns the same four new fields, so a userland ScrollView can be fully reactive to content changes via the existing layout listener.
- The readme ScrollView recipe (and the runnable `examples/scroll`) wrap the content in a `flexShrink={0}` container. Without it, Yoga squeezes the children into the fixed-height viewport and there is nothing to scroll; the recipe notes this explicitly since it is the first thing anyone building a ScrollView will hit.

One deliberate divergence from the snippet in #765: child extents are measured relative to the inside of the container's border (parent border offsets are subtracted), so `scrollWidth === clientWidth` when content exactly fits a bordered box. The issue snippet measured against the outer edge, which over-reports the scroll extent by the border width.

Deliberately not included, per the discussion in #765: scrollbars, overflow indicators, an imperative scroll API, and any built-in `ScrollView` component. Offsets are also not clamped to the content extent, clamping policy belongs to the userland component (the readme example clamps via `scrollHeight - clientHeight`).

Tests: rendering tests for vertical/horizontal offsets, offsets inside borders, nested offset containers, offset beyond content, and a zero-offset behavior lock; metric tests for client/scroll sizes with overflowing content, borders, and content that grows after mount. All fail on `master` without the change (except the behavior lock). Verified manually in a real terminal with the included example (both axes, clamping at all edges).

Also relevant to #222.

### DIFF
--- a/examples/scroll/index.ts
+++ b/examples/scroll/index.ts
@@ -1,0 +1,1 @@
+import './scroll.js';
--- a/examples/scroll/scroll.tsx
+++ b/examples/scroll/scroll.tsx
@@ -19,30 +19,32 @@ function ScrollView({
 	const contentRef = useRef(null);
 	const {clientWidth, clientHeight} = useBoxMetrics(ref);
 	const content = useBoxMetrics(contentRef);
-	const [scrollTop, setScrollTop] = useState(0);
-	const [scrollLeft, setScrollLeft] = useState(0);
+	const [requestedScrollTop, setRequestedScrollTop] = useState(0);
+	const [requestedScrollLeft, setRequestedScrollLeft] = useState(0);
 	const maxScrollTop = Math.max(0, content.height - clientHeight);
 	const maxScrollLeft = Math.max(0, content.width - clientWidth);
 
+	// Clamp on every render, not just in the key handlers: the maximum shrinks
+	// when the content gets shorter or the terminal gets wider, and an offset
+	// left over from before would scroll the viewport past its content.
+	const scrollTop = Math.min(requestedScrollTop, maxScrollTop);
+	const scrollLeft = Math.min(requestedScrollLeft, maxScrollLeft);
+
 	useInput((_input, key) => {
 		if (key.upArrow) {
-			setScrollTop(previousScrollTop => Math.max(0, previousScrollTop - 1));
+			setRequestedScrollTop(Math.max(0, scrollTop - 1));
 		}
 
 		if (key.downArrow) {
-			setScrollTop(previousScrollTop =>
-				Math.min(maxScrollTop, previousScrollTop + 1),
-			);
+			setRequestedScrollTop(Math.min(maxScrollTop, scrollTop + 1));
 		}
 
 		if (key.leftArrow) {
-			setScrollLeft(previousScrollLeft => Math.max(0, previousScrollLeft - 2));
+			setRequestedScrollLeft(Math.max(0, scrollLeft - 2));
 		}
 
 		if (key.rightArrow) {
-			setScrollLeft(previousScrollLeft =>
-				Math.min(maxScrollLeft, previousScrollLeft + 2),
-			);
+			setRequestedScrollLeft(Math.min(maxScrollLeft, scrollLeft + 2));
 		}
 	});
 
--- a/examples/scroll/scroll.tsx
+++ b/examples/scroll/scroll.tsx
@@ -16,9 +16,12 @@ function ScrollView({
 	readonly children: React.ReactNode;
 }) {
 	const ref = useRef(null);
-	const {clientHeight, scrollHeight} = useBoxMetrics(ref);
+	const {clientWidth, clientHeight, scrollWidth, scrollHeight} =
+		useBoxMetrics(ref);
 	const [scrollTop, setScrollTop] = useState(0);
+	const [scrollLeft, setScrollLeft] = useState(0);
 	const maxScrollTop = Math.max(0, scrollHeight - clientHeight);
+	const maxScrollLeft = Math.max(0, scrollWidth - clientWidth);
 
 	useInput((_input, key) => {
 		if (key.upArrow) {
@@ -30,6 +33,16 @@ function ScrollView({
 				Math.min(maxScrollTop, previousScrollTop + 1),
 			);
 		}
+
+		if (key.leftArrow) {
+			setScrollLeft(previousScrollLeft => Math.max(0, previousScrollLeft - 2));
+		}
+
+		if (key.rightArrow) {
+			setScrollLeft(previousScrollLeft =>
+				Math.min(maxScrollLeft, previousScrollLeft + 2),
+			);
+		}
 	});
 
 	return (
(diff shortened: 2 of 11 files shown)


sindresorhus:
Overall, I think the direction makes sense. Keeping scrolling in userland and adding only render offsets plus measurement primitives feels like the right boundary.

I manually tested this and found two things that should be fixed before merging:

1. Nested clipping is broken when an offset is applied. If the scrolled content contains another overflow-hidden box, the inner clip replaces the outer clip instead of intersecting it. Content can then escape the viewport and overwrite surrounding UI. In my test, a line above the scroll view was replaced by the first clipped line. This is an existing renderer bug, but this PR makes it reachable through the normal documented scrolling path and currently claims that nested offsets compose. I think clips should be intersected, or nested clipped content needs to be explicitly unsupported and the composition test removed.

2. Offsets accept any number, but fractional values are passed directly into terminal cell coordinates. A vertical offset of 0.5 dropped one of two rendered lines, and negative fractions can erase the content entirely. These should be finite integer column/row values, normalized once at the render boundary, with tests for both axes.

I am less convinced that scrollWidth and scrollHeight should try to emulate the DOM. The current calculation only looks at direct child boxes and their positive right/bottom edges, so it does not actually have DOM semantics. Recursively handling descendant overflow, reverse layouts, nested clips, negative origins, and padding would add quite a lot of complexity.

The simpler option is to keep clientWidth/clientHeight, require the nonshrinking content wrapper that the example already uses, and measure that wrapper with a seperate useBoxMetrics ref. The maximum offset is then content.width - viewport.clientWidth (and the same for height). That makes the supported boundary explicit and avoids maintaining a partial browser scrolling model. I would prefer removing scrollWidth/scrollHeight and using that approach. If those fields stay, their docs should at least describe the narrower direct-child, normal-flow contract instead of claiming total DOM-like content size.

One small nonblocking cleanup: useBoxMetrics currently spreads the complete measureElement result, so x and y are present at runtime even though they are not part of its return type. Explicitly selecting the intended fields would keep the runtime shape and change detection consistent.

sindresorhus:
I found three remaining issues:

1. Horizontal scrolling loses a column when the offset lands in the middle of a double-width character. With `你A` in a two-column viewport, offsets 1 and 2 both render `A`. At offset 1 it should render ` A`, since the first visible cell is the clipped half of `你` and `A` still belongs in the second cell. This is reproducable with the current code and means column offsets are not actually preserved for CJK text.

2. The example only clamps the offset inside the arrow handlers. If the content becomes shorter, or the terminal becomes wider after scrolling horizontally, the maximum changes but the existing offset stays out of range. The viewport can remain blank until another key press resets it. The current offset should be clamped whenever the measured maximum shrinks.

3. The documented `content.height - clientHeight` calculation is only correct when the viewport has no padding. For example, a height-4 viewport with one row of top padding and height-4 content reports a maximum of 0, but the fourth row is clipped and needs an offset of 1. The simplest solution may be to explicitly say that padding belongs on the content wrapper, not the viewport, instead of expanding the measurement API to model padding.

costajohnt:
All three are fixed, pushed in 7b334d8 and 170f05c. I should have written this up when I pushed rather than leaving the commits to speak for themselves.

**1. Wide characters at the clip edge.** Confirmed, and your `你A` case reproduced exactly as described. `sliceAnsi` works in columns but drops a wide character entirely when the slice edge falls between its halves, so the visible half stopped occupying its cell and everything after it shifted left by one column. Clipping now pads back the columns the slice lost, on both edges: `你A` at offset 1 renders ` A`, and a sweep across every offset in the line keeps the total width constant. Tests cover the column sweep and the overlap case where both edges cut a wide character.

**2. Stale offset when the maximum shrinks.** The example now clamps on every render rather than only inside the arrow handlers, so a shorter content or a wider terminal pulls the current offset back into range immediately instead of leaving the viewport blank until the next key press. The state holds the requested offset and the clamped value is what gets rendered and what the handlers step from. The readme recipe carries the same change.

**3. Padding on the viewport.** Took your suggestion and documented the constraint instead of growing the measurement API to model padding. The readme and the `useBoxMetrics` docs now both say the `content.height - clientHeight` bound assumes an unpadded viewport, and that padding belongs on the content wrapper, with the reason spelled out: `clientWidth`/`clientHeight` exclude borders but not padding, so a padded viewport reports more room than its content actually gets and the bound comes out one row short per padded edge.

CI is green on Node 22 and 24. No API surface changed for any of these.

sindresorhus:
There is one blocking clipping bug left:

Nested horizontal clips with no overlap can still write a cell outside both clips when the content contains a wide character. I reproduced this with an outer clip covering columns 0 through 5 and an inner clip covering 6 through 10. Their intersection is empty, so nothing should be drawn, but an underlying XXXXXXXXXX becomes XXXXXX XXX. The wide-character padding path receives a slice where from is greater than to, manufactures a space, moves the write position to the inner left edge, and erases the existing cell. This is the same class of corruption the clip intersection change is meant to prevent. Empty intersections should be rejected before processing a write, using x1 >= x2 and y1 >= y2, with a regression test that combines disjoint clips, underlying content, and a wide character. The current disjoint-clip test only uses ASCII content, so it does not catch this.

Two smaller points:

The padding synthesized when a clip splits a wide character loses the character's inline background style. For example, a red-background 你 followed by an unstyled A, viewed at horizontal offset 1, produces an unstyled blank followed by A. This is reproducable with colors forced on. I would not add another string-level special case for this. Either clipping eventually needs to operate on styled terminal cells, or styling of the hidden half-cell should be explicitly treated as best-effort. I do not think this one needs to block the PR if we choose the latter boundary.

Measured x/y and left/top remain Yoga coordinates even though content offsets move descendants on screen. Keeping layout coordinates unchanged is consistent with the design, but the mouse-coordinate documentation is now incomplete: adjusting only for the live region position is not enough inside a scrolled container. It should say that these metrics exclude ancestor content offsets and that callers must account for them when hit-testing scrolled content.

Longer term, the cleaner solution is to clip against the styled cell representation rather than patching slice-ansi output with lostLeading/lostTrailing spaces. That would naturally preserve cell occupancy, styles, and empty intersections. I would not expand this PR into that refactor, though. For this PR, I would fix the empty-intersection bug, clarify the coordinate contract, and keep the split-cell styling behavior best-effort unless there is a simple cell-level solution.

costajohnt:
Fixed in 568b53a.

**Empty intersections.** `Output#get` now rejects a clip whose bounds are empty on either axis (`x1 >= x2` or `y1 >= y2`) before any slicing, so the wide-character padding path never sees `from` past `to`. Regression test combines an outer clip over 0-5, an inner clip over 6-11, underlying `XXXXXXXXXX` content and a wide character at the split; it went `XXXXXX XXX` before and is untouched now.

**Split-cell styling.** Kept as best-effort, documented in the readme next to the offset props: the blank standing in for the hidden half does not carry the character's inline styling.

**Coordinates.** `x`/`y` and `left`/`top` docs (readme and `useBoxMetrics` types) now state they are layout coordinates that exclude ancestor content offsets, so hit-testing inside a scrolled container has to subtract those too.

Agree the cell-level clipping refactor is the right long-term shape; left it out of this PR as you suggested. CI green on 22 and 24.

sindresorhus:
One remaining bug: scrolling `'A\nABCDEFGHIJ'` by six columns in a bordered, two-column viewport erases the right border on the short row. The padding logic turns an empty slice into five spaces.

Returning an empty string when from >= to fixes this locally.

```js
test('horizontal scrolling preserves borders beside fully clipped rows', t => {
		<Box
			width={4}
			height={4}
			borderStyle="single"
			overflow="hidden"
			contentOffsetX={6}
		>
			<Box width={10} flexShrink={0}>
				<Text>{'A\nABCDEFGHIJ'}</Text>
			</Box>
		</Box>,
	);

	t.is(output, ['┌──┐', '│  │', '│GH│', '└──┘'].join('\n'));
});
```

costajohnt:
Fixed in ce75f60. `sliceLineToColumns` returns an empty string when `from >= to`, and your test is in as written.

Issue #969, opened 2026-06-10
Ink 7 rendering broken on all Windows terminals


zeonseoi:
Since upgrading from Ink 6 to 7, the ANSI escape sequences generated by Ink 7's render pipeline are incompatible with every Windows terminal — Git Bash, PowerShell, cmd, Windows Terminal. All produce garbled, overlapped text.

Reverting to Ink 6 resolves the issue completely.

PR #994, opened 2026-08-17
Fix: clear viewport with home + erase-down instead of `clearTerminal` (preserve scrollback)

### LINKED ISSUE #935: Render loop calls clearTerminal (CSI 3 J) — wipes terminal scrollback when output exceeds viewport

### LINKED ISSUE #990: When output exceeds terminal height, every re-render erases the user's entire scrollback (`ESC[3J`) — and the only alternative is duplicate-stamping
## Summary

> **Terminology:** the *frame* is the block of lines Ink renders and live-updates — the app's current UI (in Claude Code: the streaming response + input box + status bar). On every state change Ink erases and redraws it in place. Content that has scrolled off above it is ordinary committed terminal history.

When an Ink app's frame is taller than the terminal viewport, **every single re-render wipes the terminal's entire scrollback buffer** — including the user's own shell history from before the app started. Measured on `ink@7.1.1`:

| Verified behavior (ink@7.1.1, tmux, macOS, node 23) | Result |
|---|---|
| Ticking 40-row app in a 20-row terminal, 4 seconds | **12 full-scrollback wipes** (`ESC[2J ESC[3J ESC[H` per render) |
| 102 lines of shell history in scrollback, then one window resize while the app runs | **0 lines remain** — everything before the app is gone |
| Phone/web terminals (address bar fires a resize on nearly every scroll gesture; the on-screen keyboard fires two per message) | scrollback is effectively **permanently empty** while a tall Ink app runs |

**The whole failure in one loop** — Ink's actual captured byte output replayed through xterm.js:

Scrollback wipe demo: app running, history intact, one resize erases it

**Before** — 80×24, user scrolled up; 100 lines of shell history intact above the running app:

Before resize: 100 lines of shell history intact above the running app

**After one resize (80→64 cols)** — Ink emitted `ESC[3J`; scrolling up shows nothing:

After one resize: scrollback empty

## Mechanism

Ink erases previous frames with cursor-relative motion ([`src/log-update.ts`](https://github.com/vadimdemedes/ink/blob/master/src/log-update.ts)):

```ts
// standard renderer
stream.write(returnPrefix + ansiEscapes.eraseLines(previousLineCount) + str + cursorSuffix);
// incremental renderer
buffer.push(ansiEscapes.cursorUp(previousLines.length - 1));
```

Cursor motion is **viewport-relative**: once part of the frame has scrolled off the top, `cursorUp` saturates at the viewport edge and the scrolled-off rows are unreachable. Ink handles this in [`src/ink.tsx`](https://github.com/vadimdemedes/ink/blob/master/src/ink.tsx) by clearing the terminal whenever the frame overflows:

```ts
const isFullscreen = isTty && outputHeight >= viewportRows;
const shouldClearTerminal = shouldClearTerminalForFrame({ ... });
// when true:
this.options.stdout.write(ansiEscapes.clearTerminal + ...)
```

`ansiEscapes.clearTerminal` is `ESC[2J ESC[3J ESC[H` — and **`ESC[3J` erases the scrollback buffer**, not just the screen. Byte capture of a re-rendering overflowing app confirms one `3J` per render (12 in 4s). Resize goes through the same path ([`resized()`](https://github.com/vadimdemedes/ink/blob/master/src/ink.tsx) → `calculateLayout()` → `onRender()`), plus `log.clear()` on width shrink — so **every SIGWINCH wipes scrollback too**.

```
    THE OVERFLOW DILEMMA — cursor motion can't reach scrolled-off rows,
    so a frame taller than the viewport forces one of two failure modes:

  ┌╌╌ scrollback ╌╌╌╌╌╌┐            ┌╌╌ scrollback ╌╌╌╌╌╌┐
  │ (user's shell       │           │ frame rows 1-10     │ ◄ stale copy 1
  │  history: ERASED    │           │ frame rows 1-10     │ ◄ stale copy 2
  │  by ESC[3J on every │           │ frame rows 1-10     │ ◄ stale copy 3 …one per re-render
  │  re-render)         │           ├──── viewport ───────┤
  ├──── viewport ───────┤           │ frame rows 11-30    │
  │ frame rows 11-30    │           │                     │
  └─────────────────────┘           └─────────────────────┘
   MODE A: ink@7.1.1 today           MODE B: fork without ESC[3J
   "destroy the scrollback"          "duplicate-stamp the scrollback"
```

```mermaid
sequenceDiagram
    participant V as Viewport (phone browser)
    participant A as Ink app (frame > rows)
    participant S as Terminal scrollback

    Note over V: user scrolls → address bar hides
    V->>A: SIGWINCH
    A->>S: ESC[3J — scrollback erased (wipe #1)
    Note over V: user scrolls → address bar shows
    V->>A: SIGWINCH
    A->>S: ESC[3J — wipe #2
    Note over V: keyboard opens, closes
    V->>A: SIGWINCH ×2
    A->>S: wipes #3, #4
    Note over A: app streams output, re-rendering ~3×/s
    A->>S: continuous wipes
    Note over S: scrollback is effectively always empty
```

## Mode B at scale: the duplicate-stamping variant (Claude Code case study)

Dropping the `ESC[3J]` (as downstream forks do — Claude Code's bundled renderer repaints with `ESC[H` + a full pre-wrapped rewrite and no scrollback clear, verified by PTY capture) preserves scrollback but trades destruction for **duplication**: every repaint's overflow scrolls another copy of the frame top into scrollback, and nothing can ever erase it.

Real Claude Code session: the same response stamped 5x after three resizes

Reproduced in tmux with byte capture: a two-paragraph response went from 2 occurrences (prompt echo + response) to **5** after three resizes — one stamp per resize — while shell history stayed intact; the captured resize bytes contain **zero** `ESC[3J`/`ESC[2J`, only `ESC[H` + full rewrites.

Measured in real Claude Code sessions on a phone-sized web terminal:

- 4 window-width changes over a short conversation → every transcript paragraph appeared **3–4×** in scrollback;
- a few minutes of scrolling on a phone (one resize per gesture) with an overflowing frame → the welcome screen stamped **~70×** ≈ **2,100 junk lines**;
- because each stamp is wrapped to the width at that moment, copies are not byte-identical, so terminals can't trivially deduplicate them — and interrupted repaints were observed committing spliced partial rows (`ClauClaudeeCode`, `…planet's landg`).

So both available strategies lose: **wipe the user's history, or bury it in stale copies.**

## Reproduction (verified end-to-end before filing)

```js
// repro.js  (plain node, no JSX/build step — package.json needs "type": "module")
import React from 'react';
import {render, Text, Box} from 'ink';

const h = React.createElement;
const rows = Array.from({length: 40}, (_, i) =>
	h(Text, {key: i}, `row-${String(i + 1).padStart(2, '0')} the quick brown fox jumps over the lazy dog`)
);
render(h(Box, {flexDirection: 'column'}, ...rows));
setInterval(() => {}, 1 << 30); // keep the process alive so resizes reach a live app
```

```bash
npm init -y && npm pkg set type=module && npm i ink react

tmux new-session -d -s inkdemo -x 80 -y 24
tmux send-keys -t inkdemo 'seq -f "shell-history-line-%.0f" 1 100' Enter && sleep 1
tmux capture-pane -t inkdemo -p -S -3000 | grep -c shell-history-line   # → 102 (history present)

tmux send-keys -t inkdemo 'node repro.js' Enter && sleep 2
tmux resize-window -t inkdemo -x 64 -y 24 && sleep 1
tmux capture-pane -t inkdemo -p -S -3000 | grep -c shell-history-line   # → 0 (history erased)
```

Byte-level confirmation (via `tmux pipe-pane`): the resize emits exactly one `ESC[2J ESC[3J ESC[H` followed by the full frame; a state-updating variant of the same app emits one per render (12 wipes in 4 s at ~3 renders/s).

## Can this be mitigated downstream? (we tried — here's how far it goes)

We hit Mode B at scale through [reminal](https://github.com/harshalgajjar/Reminal), a terminal-sharing tool we maintain — it puts any terminal session (Claude Code included) in a phone browser, which is exactly the small-viewport, resize-happy environment where this bug bites hardest. Everything below is implemented and battle-tested there; the cleanup lives in [`internal/client/dedup.go`](https://github.com/harshalgajjar/Reminal/blob/main/internal/client/dedup.go) and [`internal/client/agent.go`](https://github.com/harshalgajjar/Reminal/blob/main/internal/client/agent.go) for anyone who wants to study or borrow it. A terminal cannot change the app, so the only options are heuristics on the emitted bytes. Sharing what proved necessary mainly as a data point for how much machinery the workaround takes — any terminal or multiplexer author attempting the same will need equivalents of all three:

1. **Resize-jitter coalescing** — debounce viewport-size reports (shrinks fast; small rows-only grows only after 2 s of stability) so the app sees ~1 SIGWINCH per real transition instead of one per scroll gesture. This alone reduced stamping ~30×.
2. **Captured-frame resize segments** — at each SIGWINCH, snapshot the words of the screen being repainted and where its overflow will land in scrollback. A row is later dropped only if it word-matches the frame captured at *that* resize **and** the same content provably occurs again later. Stale copies always re-occur; genuine output never matches — lossless by construction.
3. **Paragraph-level and truncated-prefix dedup** for the residue, plus periodic re-sync of live viewers.

This gets real sessions to roughly every content line exactly once — but it required a server-side VT emulator, word-level fuzzy matching, and careful correctness arguments to approximate what the app could guarantee in one line of intent: *don't touch what has scrolled away.*

One more finding from running this in production, relevant to anyone considering downstream cleanup: **the mitigation's memory must be unbounded in principle.** A stamped copy stays in scrollback for as long as the scrollback lives, so the cleaner must remember *every* resize's fingerprint for that entire lifetime. We initially kept the last 32 resize fingerprints; real phone sessions (a keyboard show/hide per message, reconnects whenever the phone sleeps) exhausted that within minutes, and any stamp older than the window became permanently uncleanable — the copies had drifted into differently-wrapped, splice-corrupted variants no content matcher can safely pair. We now keep 1024 and it is still only a budget, not a bound. The app, by contrast, needs no memory at all — it just has to not re-emit what already scrolled off.

## Suggested directions

1. **Clamp the frame to the viewport.** When `outputHeight > rows`, render only the bottom `rows` lines and never rewind/erase/rewrite above them. Content that has scrolled off is, by definition, already committed history.
2. **Flush overflow through the `<Static>` pipeline exactly once — and emit it as soft-wrapped logical lines** (no baked-in line breaks; let the terminal wrap). The soft-wrap half matters: committed rows with hard breaks are frozen at their emission width forever, which is the root of the long-standing "transcript doesn't reflow on terminal resize" complaint class downstream (see the discussion in anthropics/claude-code#51828 — Codex-style transcript emission demonstrates the reflowing behavior). Pre-wrapping is what makes re-emission seem necessary in the first place; drop it and there is nothing to re-emit *and* scrollback reflows natively.
3. **At minimum, use `ESC[2J` + home instead of `clearTerminal`'s `ESC[3J`** so the clear path stops destroying the user's scrollback (also raised in #621). This converts Mode A into Mode B, which is less destructive but still leaks copies — 1–2 are the real fix.

Happy to provide the raw byte captures, the xterm.js replay page used for the screenshots, or to test a branch.

Related: #621 (scrollback cleared when UI taller than terminal), #907 (resize artifacts when wrapping changes row count), #450, #359 (flicker when taller than screen), #222, #153; downstream: anthropics/claude-code#51828.

---
**Edit (2026-08-12):** folded the scrollback-reflow connection (suggested direction 2) in from the discussion on anthropics/claude-code#51828 — pre-wrapping turns out to be the shared root of both the duplication here and the frozen-scrollback complaints there.

---
*Environment: ink@7.1.1, node v23.10.0, tmux 3.x, macOS 14; behavior confirmed by raw PTY byte capture (`tmux pipe-pane`) and replay through xterm.js 5.5.*

costajohnt:
## Summary

Frames taller than the viewport fall back to a full clear on every re-render, and `ansiEscapes.clearTerminal` emits `CSI 3J`, which erases the terminal's scrollback. A ticking overflowing app, or a single resize, wipes the user's shell history (#935, #990).

This replaces the sequence with `ansiEscapes.cursorTo(0, 0) + ansiEscapes.eraseDown` (`CSI 1;1H CSI J`), as suggested in the #936 review. `eraseScreen` (`CSI 2J`) is not a safe replacement on its own: VS Code and Windows Terminal push the viewport into scrollback before clearing, so repeated redraws churn bounded history. Home + erase-down clears only what is visible.

Builds on #936 by moeritze and the terminal-matrix measurements razchiriac posted on #935, which confirmed home + erase-down preserves history with no model-dependent churn.

## Changes

- `src/ink.tsx`: full-clear path writes `homeAndEraseDown` (`cursorTo(0, 0) + eraseDown`) instead of `clearTerminal`. Exported so tests assert against the real sequence; deliberately not named `clearViewport`, which ansi-escapes already exports as `CSI 2J + CSI H`. Applies to the `#969` Windows path too: that path needs a full viewport clear per fullscreen frame, and home + erase-down provides one without `3J`.
- New fixture `issue-935-overflow-rerender`: prints scrollback lines, then renders a `rows + 1` frame eight times, with frame 7 carrying a longer `STALE-ROW` label so a missing or partial clear leaves residue behind.
- New test `#935: overflowing rerenders must not erase terminal scrollback`: asserts a full clear happens per rerender, `CSI 3J` / `CSI 2J` / `clearTerminal` are absent from the raw PTY output, the pre-existing scrollback lines survive when replayed through the existing `reconstructTerminalLines` reducer, the final viewport shows the last frame, and `STALE-ROW` is absent from the whole reconstructed buffer. Fails on master; also fails against `cursorTo(0, 0)` alone (no erase) and against `eraseScreen + home`.
- Existing byte-level assertions (`clearTerminalCount`, `split(clearTerminal)`, cursor sync-path check) updated to the new sequence and renamed `fullClearCount`.

## Verification

- `npm test`: 1063 passed, same 4 pre-existing known failures as master.
- Live tmux repro from #990 (80x24, 100 lines of shell history, 40-row app, one resize to 64 cols): master keeps 0 history lines after the resize, this branch keeps all 102.
- Ticking variant (40-row app re-rendering 3x/s for 4 s, then a resize), raw PTY bytes via `tmux pipe-pane`: ink@7.1.1 emits 18x `CSI 2J CSI 3J` and keeps 0 history lines; this branch emits 17x `CSI 1;1H CSI J`, no 2J/3J, keeps all 102.
- Same captures replayed through `@xterm/headless` 6.1.0-beta.301 with `scrollOnEraseInDisplay` off and on: published keeps 0/100 history lines in both; this branch keeps 100/100 with an identical 414-line buffer in both models. Swapping the sequence for `2J + home` in the same capture keeps history too but grows the buffer to 798 lines with `scrollOnEraseInDisplay` on, matching the #936 review's concern.
- Eyeballed side by side in Terminal.app on macOS 14: with this branch, scrolling up and resizing while the app ticks leaves the shell history intact; with ink@7.1.1 it is gone after the first redraw.

As noted in the #936 review, an oversized dynamic frame still consumes scrollback simply by being written, so this is best-effort preservation. Concretely: each full clear re-emits `fullStaticOutput` plus the frame from home, and the rows that scroll off the top now stay in history instead of being erased by `3J`, so a long-running overflowing render leaves duplicate copies of its top rows (including `<Static>` lines) in scrollback. That churn existed before too, `3J` just hid it by deleting everything. Apps with taller-than-viewport UI should still use the alternate screen or `<Static>`. This just stops Ink from actively destroying history.

Fixes #935
Fixes #990
Closes #936

### DIFF
--- a/src/ink.tsx
+++ b/src/ink.tsx
@@ -123,6 +123,12 @@ const stripKittyQueryResponsesAndTrailingPartial = (
 // pre-7.0 behavior of fully clearing between fullscreen frames there.
 const isWindowsConsole = process.platform === 'win32';
 
+// Full-clear fallback for frames taller than the viewport. `clearTerminal`
+// also emits CSI 3J, which erases the terminal's scrollback, and CSI 2J, which
+// VS Code and Windows Terminal handle by pushing the viewport into scrollback
+// first (#935). Home + erase-down clears only what is visible.
+const clearViewport = ansiEscapes.cursorTo(0, 0) + ansiEscapes.eraseDown;
+
 const shouldClearTerminalForFrame = ({
 	isTty,
 	viewportRows,
@@ -1125,7 +1131,7 @@ export default class Ink {
 			}
 
 			this.options.stdout.write(
-				ansiEscapes.clearTerminal + this.fullStaticOutput + outputToRender,
+				clearViewport + this.fullStaticOutput + outputToRender,
 			);
 			this.lastOutput = output;
 			this.lastOutputToRender = outputToRender;
--- a/src/ink.tsx
+++ b/src/ink.tsx
@@ -123,7 +123,7 @@ const stripKittyQueryResponsesAndTrailingPartial = (
 // pre-7.0 behavior of fully clearing between fullscreen frames there.
 const isWindowsConsole = process.platform === 'win32';
 
-// Full-clear fallback for frames taller than the viewport. It must clear only
+// Full-clear path (see `shouldClearTerminalForFrame`). It must clear only
 // what is visible: `ansiEscapes.clearTerminal` emits CSI 3J, which erases the
 // terminal's scrollback, and both it and `ansiEscapes.clearViewport` emit
 // CSI 2J, which VS Code and Windows Terminal handle by pushing the viewport
--- a/src/ink.tsx
+++ b/src/ink.tsx
@@ -123,11 +123,14 @@ const stripKittyQueryResponsesAndTrailingPartial = (
 // pre-7.0 behavior of fully clearing between fullscreen frames there.
 const isWindowsConsole = process.platform === 'win32';
 
-// Full-clear fallback for frames taller than the viewport. `clearTerminal`
-// also emits CSI 3J, which erases the terminal's scrollback, and CSI 2J, which
-// VS Code and Windows Terminal handle by pushing the viewport into scrollback
-// first (#935). Home + erase-down clears only what is visible.
-const clearViewport = ansiEscapes.cursorTo(0, 0) + ansiEscapes.eraseDown;
+// Full-clear fallback for frames taller than the viewport. It must clear only
+// what is visible: `ansiEscapes.clearTerminal` emits CSI 3J, which erases the
+// terminal's scrollback, and both it and `ansiEscapes.clearViewport` emit
+// CSI 2J, which VS Code and Windows Terminal handle by pushing the viewport
+// into scrollback first (#935). Exported so tests assert against the real
+// sequence rather than a copy.
+export const homeAndEraseDown =
+	ansiEscapes.cursorTo(0, 0) + ansiEscapes.eraseDown;
 
 const shouldClearTerminalForFrame = ({
 	isTty,
@@ -1131,7 +1134,7 @@ export default class Ink {
 			}
 
 			this.options.stdout.write(
-				clearViewport + this.fullStaticOutput + outputToRender,
+				homeAndEraseDown + this.fullStaticOutput + outputToRender,
 			);
 			this.lastOutput = output;
 			this.lastOutputToRender = outputToRender;
--- a/test/cursor.tsx
+++ b/test/cursor.tsx
@@ -11,10 +11,10 @@ import {
 	useStdout,
 	useStderr,
 } from '../src/index.js';
+import {homeAndEraseDown} from '../src/ink.js';
 import {createStdin, emitReadable} from './helpers/create-stdin.js';
 import createStdout from './helpers/create-stdout.js';
 
-const clearViewport = ansiEscapes.cursorTo(0, 0) + ansiEscapes.eraseDown;
 const showCursorEscape = '\u001B[?25h';
 const hideCursorEscape = '\u001B[?25l';
 
@@ -784,7 +784,7 @@ for (const {name, incremental} of inkRenderingModes) {
 			await waitUntilRenderFlush();
 
 			const synced = getWriteCalls(stdout).slice(writesBeforeRerender).join('');
-			t.true(synced.includes(clearViewport), 'took the sync path');
+			t.true(synced.includes(homeAndEraseDown), 'took the sync path');
 			// 6 lines with no trailing newline: the cursor is left on row 5, so y=2
 			// is cursorUp(3), not the cursorUp(4) a visible-line-count basis gives.
 			t.true(
--- a/test/cursor.tsx
+++ b/test/cursor.tsx
@@ -14,6 +14,7 @@ import {
 import {createStdin, emitReadable} from './helpers/create-stdin.js';
 import createStdout from './helpers/create-stdout.js';
 
+const clearViewport = ansiEscapes.cursorTo(0, 0) + ansiEscapes.eraseDown;
 const showCursorEscape = '\u001B[?25h';
 const hideCursorEscape = '\u001B[?25l';
 
@@ -783,7 +784,7 @@ for (const {name, incremental} of inkRenderingModes) {
 			await waitUntilRenderFlush();
 
 			const synced = getWriteCalls(stdout).slice(writesBeforeRerender).join('');
-			t.true(synced.includes(ansiEscapes.clearTerminal), 'took the sync path');
+			t.true(synced.includes(clearViewport), 'took the sync path');
 			// 6 lines with no trailing newline: the cursor is left on row 5, so y=2
 			// is cursorUp(3), not the cursorUp(4) a visible-line-count basis gives.
 			t.true(
--- a/test/fixtures/issue-450-fixture-helpers.tsx
+++ b/test/fixtures/issue-450-fixture-helpers.tsx
@@ -8,19 +8,24 @@ type RerenderFixtureOptions = {
 	readonly includeStaticLine?: boolean;
 	readonly rowsFallback?: number;
 	readonly heightForFrame: (rows: number, frameCount: number) => number;
+	readonly labelForFrame?: (frameCount: number) => string;
 };
+
+const defaultLabelForFrame = (frameCount: number) => `frame ${frameCount}`;
 
 function Issue450RerenderFixtureComponent({
 	completionMarker,
 	frameLimit,
 	includeStaticLine,
 	heightForFrame,
+	labelForFrame,
 	rows,
 }: {
 	readonly completionMarker?: string;
 	readonly frameLimit: number;
 	readonly includeStaticLine: boolean;
 	readonly heightForFrame: (rows: number, frameCount: number) => number;
+	readonly labelForFrame: (frameCount: number) => string;
 	readonly rows: number;
 }) {
 	const {exit} = useApp();
@@ -61,7 +66,7 @@ function Issue450RerenderFixtureComponent({
 			<Box height={targetHeight} flexDirection="column">
 				<Text>#450 top</Text>
 				<Box flexGrow={1}>
-					<Text>{`frame ${frameCount}`}</Text>
+					<Text>{labelForFrame(frameCount)}</Text>
 				</Box>
 				<Text>#450 bottom</Text>
 			</Box>
@@ -75,6 +80,7 @@ export const runIssue450RerenderFixture = ({
 	includeStaticLine = false,
 	rowsFallback = 6,
 	heightForFrame,
+	labelForFrame = defaultLabelForFrame,
 }: RerenderFixtureOptions): void => {
 	const rows = Number(process.argv[2]) || rowsFallback;
 	process.stdout.rows = rows;
@@ -85,6 +91,7 @@ export const runIssue450RerenderFixture = ({
 			frameLimit={frameLimit}
 			includeStaticLine={includeStaticLine}
 			heightForFrame={heightForFrame}
+			labelForFrame={labelForFrame}
 			rows={rows}
 		/>,
 	);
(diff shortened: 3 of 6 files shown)


harshalgajjar:
Amazing! <3

PR #995, opened 2026-08-20
Fix: clear viewport with home + erase-down instead of clearTerminal (preserve scrollback)

### LINKED ISSUE #990: When output exceeds terminal height, every re-render erases the user's entire scrollback (`ESC[3J`) — and the only alternative is duplicate-stamping

GujjetiMokshith:
This PR fixes #990 by replacing `ansiEscapes.clearTerminal` with `ansiEscapes.cursorTo(0, 0) + ansiEscapes.eraseDown`.

### Problem
When the output exceeds the terminal height, Ink falls back to erasing the terminal. The use of `clearTerminal` emits `ESC[2J ESC[3J ESC[H`, which clears the entire scrollback buffer, effectively erasing all of the user's shell history above the current application.

### Solution
This PR changes the behavior to simply move the cursor to the top-left using `ansiEscapes.cursorTo(0, 0)` (which translates to `ESC[1;1H`) and then erases downwards with `ansiEscapes.eraseDown` (`ESC[J`). This still completely clears the active viewport but prevents wiping out the preserved scrollback history.


sindresorhus:
https://github.com/vadimdemedes/ink/pull/994

PR #996, opened 2026-09-05
ci: verify build output in test workflow

sabtain05:
## Summary

- Run the TypeScript build in the existing Node.js CI matrix
- Keep build validation ahead of the existing test suite

## Why

The workflow already tests on Node.js 22 and 24, but it does not explicitly verify that the package builds successfully. This adds that check without changing the existing test strategy.


sindresorhus:
This is moot.

Issue #450, opened 2021-06-17
Flickering when rendering element with height precisely equal to `process.stdout.rows`


valyrie97:
Im using the ["new" windows terminal](https://www.microsoft.com/store/productId/9N0DX20HK701). My test code is the following:

```tsx
import React, {useState} from 'react';
import {render, Box} from 'ink';


const App = () => {
	const [count, setCount] = useState(0);

	setTimeout(() => {
		setCount(count + 1);
	}, 100);

	return <Box height={process.stdout.rows} borderStyle="round" width={10}></Box>;
};

const app = <App />;
render(app);
```

when i reduce the height by one, the problem goes away entirely. I assume this has something to do with an accidental scroll, as my scrollbar instantly becomes mere pixels at full height, but is never changed when reduced. Im not sure why this would be happening, however, as im never drawing more than the height, so it should not overflow and scroll.

marcus13345:
![popopopopopop](https://user-images.githubusercontent.com/6365746/122353888-798a3000-cf1e-11eb-9ef2-a727ea3e1880.gif)

vadimdemedes:
Don't think I'd be able to help here, since I don't have a Windows machine. Sometimes I see flickering on macOS as well due to high re-rendering rate (in your case, 100ms, which is pretty fast) and when there are a lot of lines to erase.

marcus13345:
maybe then this is out of scope for the use case of ink. My original intention was to be able to show things like animations happening per frame, however if your perspective is 100ms being too fast, maybe this is simply not a good use case for ink?

I know the technical limitations of a terminal are restrictive, and that asking for that many new characters is a lot, especially at such a rate. However, there are potential solutions, like overwriting characters dynamically instead of clearing and completely re-rendering. though my familiarity with react's internal rendering engine is minimal.

NSExceptional:
@vadimdemedes I see flickering too, in the same project on both windows and mac. It's a big box but it's mostly empty squares. Surely there's some way around the jittering?

dustinlacewell:
Just remarking that I still experience this today.

nigulasikk:
In gemini-cli, on Windows PCs and in small-screen scenarios, this rendering issue is extremely severe.

for  your reference: https://github.com/google-gemini/gemini-cli/issues/7016

HomyeeKing:
@vaunus  
[...]

Issue #397, opened 2020-11-11
Static + exit results in broken/mismatched output


milesj:
I'm working on a command that does some validation and outputs it nicely to the screen via `Static`, since it only needs to render once. If there are any errors, I want to hard fail with a non-zero exit code, so I'm using `exit`. Here's some stripped down example code:

```js
const React = require('react');
const { Box, Text, Static, useApp } = require('ink');

const items = [{ key: 'foo' }, { key: 'bar' }, { key: 'baz' }];

function ThrowError() {
  const { exit } = useApp();

  React.useEffect(() => {
    exit(new Error('Exit'));
  }, [exit]);

  return (
    <Box>
      <Static items={items}>{({ key }) => <Text key={key}>{key}</Text>}</Static>
      <Text>Content</Text>
    </Box>
  );
}
```

I would expect the above to output the static items, then cause an exit. However, this is what I receive.

```
foo
bar
baz
foo
bar
baz
Content
```

2 sets of static output, no `Exit` thrown error, AND it doesn't set an exit code! It also seems to be rendering `Content`, which it should not be since it's not static? HOWEVER, if I comment out the `Static` component and re-run, I now receive this (using my custom error renderer).

```
Content

 ERROR 

 Exit

 STACK TRACE 

    at /boost/packages/cli/examples/commands/ErrorCompCommand.js:13:10
    at commitHookEffectList (/boost/node_modules/react-reconciler/cjs/react-reconciler.development.js:12025:26)
    at commitPassiveHookEffects (/boost/node_modules/react-reconciler/cjs/react-reconciler.development.js:12059:11)
    at Object.invokeGuardedCallbackImpl (/boost/node_modules/react-reconciler/cjs/react-reconciler.development.js:11563:10)
    at invokeGuardedCallback (/boost/node_modules/react-reconciler/cjs/react-reconciler.development.js:11740:31)
    at flushPassiveEffectsImpl (/boost/node_modules/react-reconciler/cjs/react-reconciler.development.js:15390:7)
    at unstable_runWithPriority (/boost/node_modules/scheduler/cjs/scheduler.development.js:697:12)
    at runWithPriority (/boost/node_modules/react-reconciler/cjs/react-reconciler.development.js:1881:10)
    at flushPassiveEffects (/boost/node_modules/react-reconciler/cjs/react-reconciler.development.js:15359:12)
    at /boost/node_modules/react-reconciler/cjs/react-reconciler.development.js:15238:11
```

[...]

PR #1005, opened 2026-09-19
Stop replaying `<Static>` output on full-clear frames

costajohnt:
Follow-up to #994. Since the full-clear path preserves scrollback, it still rewrites the accumulated `<Static>` output (`fullStaticOutput`) after every clear. Each overflowing rerender therefore stamps another copy of the whole static history into scrollback and rewrites all of it to the terminal. This is the "duplicate stamping" described in #990, and it is not in a release yet (7.1.1 predates #994).

Reproduced by running a fixture in a pty and replaying the bytes through xterm.js headless (8 rows, a `<Static>` with 3 items, a 12-line live region, 3 rerenders): master leaves 5 copies of the static lines in the terminal buffer; with this change, one.

## Change

The full clear now erases only the previous live frame and never replays static output:

- When the previous frame filled the viewport (`lastOutputHeight >= viewportRows`), `homeAndEraseDown` already erases exactly that frame, and it keeps the absolute erase that Windows consoles need (#969).
- When it did not (a small frame growing past the viewport), erase it relative to the cursor via `log.clear()` instead, so `<Static>` and console rows that are still in the viewport survive and scroll into history as the new frame is written.
- `fullStaticOutput` is only accumulated in debug mode now. A boolean keeps the unmount guard from #397.
- New `<Static>` output from the current frame is still written once, ahead of the frame, in both erase branches. The full clear previously dropped it.

## Behavior change

When the live region shrinks from fullscreen back to a small frame, the frame is written at the top of an empty viewport; the static log is one page up in scrollback. Before, the replay made the viewport look continuous (static tail + frame) at the cost of duplicating the log. If the continuous look is preferred, replaying only the last `rows - frameHeight` static lines would keep it without scrolling anything into history, at the cost of one bounded duplicate per shrink. Happy to switch to that.

## Tests

`full clears do not replay <Static> output into scrollback` and `growing into an overflowing frame keeps <Static> rows that are still in the viewport`, plus `<Static> items added while the frame overflows are written once`, for standard and incremental rendering, using `reconstructTerminalLines`. All fail on master. The existing #450, #935 and #969 tests still pass.


sindresorhus:
Could we keep replaying the accumulated `<Static>` output in alternate-screen mode?

This works on the primary screen because previously rendered static lines remain in scrollback. The alternate screen has no scrollback, though. If the live frame overflows and later shrinks, `<Static>` has no new output to emit, so the full clear writes only the live frame and the existing static content disappears permanantly. The previous implementation restored it from the accumulated output.

I think the simplest fix is to keep accumulating and replaying the full static output only when the alternate screen is active, while retaining the new write-once behavior for the primary screen. Otherwise we would need to explicitly document `<Static>` with overflowing alternate-screen output as unsupported, which would be a regression from the current behavior.

costajohnt:
Yes, that was a regression on the alternate screen. Fixed by keeping the accumulation and replay only when `alternateScreen` is on: `renderFrame` appends to `fullStaticOutput` in that mode and the full-clear path writes `homeAndEraseDown + fullStaticOutput + frame` as before. The primary screen keeps the write-once behavior.

Added two tests for standard and incremental rendering: the frame grows past a 5-row viewport and shrinks back with three static items, asserting the reconstructed screen is `S1, S2, S3, live`; and static items added while the frame overflows, asserting every full clear writes them once in order. Both fail on the previous commit.

One pre-existing gap, not from this PR: resuming from suspend re-enters the alternate screen and redraws only the frame, so static rows are lost until the next full clear. Happy to do that as a follow-up if you want it.

sindresorhus:
> One pre-existing gap, not from this PR: resuming from suspend re-enters the alternate screen and redraws only the frame, so static rows are lost until the next full clear. Happy to do that as a follow-up if you want it.

👍

